Issue #2, opened 2019-08-08
Feature Request: parallel pulling of remote dataset


Renthal:
It would be cool if one could specify the number of workers used to pull a remote dataset s.t. it could can be parallelized (and thus faster)

Issue #5, opened 2019-08-12
Dataset.__init__ and LocalDataset.__init__


Renthal:
Unresolved reference on `Client` on the init type declaration on both classes

Issue #4, opened 2019-08-12
download_all_images_from_annotations()


Renthal:
`download_all_images_from_annotations()` at line 182 in `dataset.py` 

- unresolved reference attributed on `annotations_path.glob()` since `annotations_path` is declared to be of type `str` it can't have an attribute glob

Similar to

`download_image_from_json_annotation()` at line 215in `dataset.py` 

- unresolved reference attributed on `annotations_path.stem` since `annotations_path` is declared to be of type `str` it can't have an attribute stem

PR #6, opened 2019-08-12
Fix Issues #2, #4 and #5

Renthal:
- Fixed the issues on the imports. 
- Modified the download function to enable parallel exploration of the generator and provided a minimal example on how to perform parallel downloading of a dataset.
- Split of `dataset.py` into sub files  (`remote_dataset.py` and `local_dataset.py`) and introduced two support scripts `upload_manager.py`
- Added documentation on the downloading functions

#4 #5  #2


darwin/client.py
```diff
@@ -104,18 +103,18 @@ def _get_headers(self, refresh=False):
...
                 "Content-Type": "application/json",
-                "Authorization": f"Bearer {self._refresh_token}",
+                "Authorization": f"Bearer {self.refresh_token}",
             }
         else:
-            return {"Content-Type": "application/json", "Authorization": f"Bearer {self._token}"}
+            return {"Content-Type": "application/json", "Authorization": f"Bearer {self.token}"}
 
-    def get(self, endpoint: str, retry: bool = True, raw: bool = False):
+    def get(self, endpoint: str, retry: bool = False, raw: bool = False):
```
andreaazzini: I think the `retry` value should be `True` by default, since that's probably the most common and desirable use case of it

darwin/client.py
```diff
@@ -124,15 +123,15 @@ def get(self, endpoint: str, retry: bool = True, raw: bool = False):
             return response
         return response.json()
 
-    def put(self, endpoint: str, payload: Dict, retry: bool = True):
+    def put(self, endpoint: str, payload: Dict, retry: bool = False):
```
andreaazzini: Same here for `retry` (basically, in all places where it was `True` I liked `True`)

darwin/dataset/remote_dataset.py
```diff
@@ -0,0 +1,91 @@
+from __future__ import annotations
+
+import io
+import zipfile
+from pathlib import Path
+from typing import TYPE_CHECKING, List, Optional
+
+import darwin
+from darwin.dataset.download_manager import download_all_images_from_annotations
+from darwin.dataset.upload_manager import _split_on_file_type, upload_file_to_s3
```
andreaazzini: Not sure how useful it is to split `download_manager` and `upload_manager` from here, since this is the only place we'll ever need those functions.

Renthal: I think reason would be code clarify and scalability, plus it makes the code easier to maintain (...adn the pythonic way of "explicit rather than implicit"); but I agree it is possible to merge these two files.

PR #7, opened 2019-09-03
dataset.pull(): add filter by image_status

almazan:
This PR adds a new parameter in the `dataset.pull()` method called `image_status` that can be used to filter the images that are gonna be downloaded by their current status in Darwin (eg. "done" or "in review").

For example, `dataset.pull(image_status="done")` will download only those images that have been annotated and reviewed.

### DIFF
--- a/darwin/dataset.py
+++ b/darwin/dataset.py
@@ -98,8 +98,8 @@ class Dataset:
             )
             yield
 
-    def pull(self, image_status: str = ""):
-        """Downloads a rermote project (images and annotations) in the projects directory. """
+    def pull(self, image_status: str = None):
+        """Downloads a remote project (images and annotations) in the projects directory. """
         query = f"/datasets/{self.dataset_id}/export?format=json"
         if image_status:
             query += f"&image_status={image_status}"
--- a/darwin/dataset.py
+++ b/darwin/dataset.py
@@ -98,7 +98,7 @@ class Dataset:
             )
             yield
 
-    def pull(self, image_status: str = None):
+    def pull(self, image_status: Optional[str]):
         """Downloads a remote project (images and annotations) in the projects directory. """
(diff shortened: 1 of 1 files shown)


darwin/dataset.py
```diff
@@ -98,11 +98,13 @@ def _add_files_to_dataset(self, filenames: List, fps: int):
             )
             yield
 
-    def pull(self):
+    def pull(self, image_status: str = ""):
         """Downloads a rermote project (images and annotations) in the projects directory. """
```
Renthal: There is a typo in "remote"

darwin/dataset.py
```diff
@@ -98,11 +98,13 @@ def _add_files_to_dataset(self, filenames: List, fps: int):
...
-        response = self._client.get(
-            f"/datasets/{self.dataset_id}/export?format=json", raw=True
-        )
+        query = f"/datasets/{self.dataset_id}/export?format=json"
+        if image_status:
```
Renthal: I know this perfectly works fine, however, I like to avoid using empty string as False evaluators because it makes it somewhat less intuitive. For example, here we can use` image_status=None` in the signature and go ahead with a` if image_status is not None:` , what do you think?

darwin/dataset.py
almazan: Fixed

darwin/dataset.py
almazan: I totally agree. Changed it to `None`.

darwin/dataset.py
```diff
@@ -101,7 +101,7 @@ def _add_files_to_dataset(self, filenames: List, fps: int):
     def pull(self, image_status: str = None):
```
simedw: the type would then be `image_status: Optional[str]`

almazan: Didn't know about `Optional`. Fixed.

PR #8, opened 2019-09-04
Allow Client to be created from a token

andreaazzini:


darwin/client.py
```diff
@@ -41,6 +41,18 @@ def default(cls):
         config_path = Path.home() / ".darwin" / "config.yaml"
         return Client.from_config(config_path)
 
+    @classmethod
+    def from_token(cls, token: str, projects_dir: Optional[str] = None):
+        if not projects_dir:
+            projects_dir = Path.home() / ".darwin" / "projects"
+        return cls(
+            token=token,
+            refresh_token=token,
```
simedw: Is this token an access token or refresh token?
Probably make sense to make the other token `None` then.

PR #9, opened 2019-09-16
Fix UnicodeDecodeError caused by README

andreaazzini:


darwin/cli_functions.py
```diff
@@ -294,11 +294,3 @@ def find_files(root: Path, recursive: bool, exclude: List[str]) -> List[Path]:
             ):
                 files += [file]
     return files
-
-
-# Low-level helper functions
```
andreaazzini: I removed this function, since it was never called anywhere

darwin/dataset.py
```diff
@@ -157,8 +158,8 @@ def upload_file_to_s3(client: Client, file: requests.Response) -> str:
     end_point = response["postEndpoint"]
 
     s3_response = upload_to_s3(signature, end_point, file_path)
-    if not str(s3_response.status_code).startswith("2"):
-        process_response(s3_response)
+    # if not str(s3_response.status_code).startswith("2"):
+    #     process_response(s3_response)
```
andreaazzini: Since we currently don't have this `process_response` function implemented, I commented in these lines

PR #10, opened 2019-09-19
Add integration with Pytorch datasets

almazan:
This PR extends the Darwin client and adds integration with Pytorch:

- New function `fetch_darwin_dataset` in `darwin/pytorch/utils.py` that, given a dataset identifier and a client, pulls locally the dataset from Darwin, and creates the train, validation and test splits.
- New class `Dataset` in `darwin/pytorch/dataset.py` that loads Darwin-formatted datasets and implements a `__getitem__` function that can be used to load data from standard Pytorch's data loaders.
- Pytorch's transform `ConvertPolysToMask` in `darwin/pytorch/transforms.py` to convert polygons represented as sequences of coordinates into `Pytorch.tensor` masks.

Example:

```
from darwin.pytorch.dataset import get_dataset
import darwin.pytorch.utils as utils
import darwin.pytorch.transforms as T

from torch.utils.data import DataLoader

# Initialize client (optional)
cli = Client(...)

# Initialize transforms (optional)
trfs = T.RandomHorizontalFlip()

# Get dataset
dataset = get_dataset('fiber-production-inspection', image_set="train", 
                      transforms=trfs,  client=cli, poly_to_mask=True, 
                      val_perc=0.1, test_perc=0.05, split_seed=42,
                      force_fetching=True)

image, targets = dataset[0]

# Create Pytorch dataloader
data_loader = DataLoader(dataset, num_workers=8, collate_fn=utils.collate_fn)
```


darwin/pytorch/utils.py
```diff
@@ -0,0 +1,232 @@
+import os
+import errno
+import torch
+import numpy as np
+import matplotlib.pyplot as plt
+import json
+from pycocotools import mask as coco_mask
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, Dict, List, Optional
```
simedw: you don't seem to be using `TYPE_CHECKING` anywhere.

almazan:
It has two external dependencies that might be overcome. `pycocotools` is used to transform the polygon into mask, but maybe we could integrate those functions into the darwin-cli. And `tqdm` is used to show the progress when pulling a dataset, but this can be removed.

simedw:
tqdm is already a dependency on the cli part.

darwin/pytorch/utils.py
```diff
@@ -0,0 +1,232 @@
...
+
+def fetch_darwin_dataset(
+    db_name,
+    client: Optional = None,
+    val_perc: Optional[float] = 0.1,
+    test_perc: Optional[float] = 0,
+    image_status: Optional[str] = "done",
+    force_resplit: Optional[bool] = False,
+    split_seed: Optional[int] = None,
+    **kwargs
```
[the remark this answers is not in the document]
almazan: I was using it before the re-formatting but don't use it anymore. I removed it.

darwin/pytorch/utils.py
almazan: True. Same for `Any`, `Dict`, `List`. Removed those four.

darwin/pytorch/utils.py
```diff
@@ -0,0 +1,232 @@
...
+):
+    if client is None:
+        client = Client.default()
+
+    # Get data
+    local_datasets = {dataset.slug: dataset for dataset in client.list_local_datasets()}
+    if db_name in local_datasets:
```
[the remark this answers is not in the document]
almazan: Yeah, makes sense. I'm working on this, because I think it may require removing the local copy first.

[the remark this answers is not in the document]
almazan: Just added support for the `force_fetching` flag. It checks firsts if there's a local copy of the dataset and removes it. This was necessary in order to deal with cases where some images have been removed from the remote copy.

https://github.com/v7labs/darwin-cli/blob/df571ad7e96e3d3377194b6bf0e074f6a4c75e9a/darwin/pytorch/utils.py#L190-L213

almazan:
I've added a new parameter `split_seed` that can be used to fix the seed of the random split: https://github.com/v7labs/darwin-cli/blob/df571ad7e96e3d3377194b6bf0e074f6a4c75e9a/darwin/pytorch/utils.py#L234-L238

almazan:
Updated the instructions in the first comment showing new flags and options.

simedw:
> Updated the instructions in the first comment showing new flags and options.

I think it would be great if we had two examples in the readme file. (one using all the default parameters and one using optional parameter (like client, transforms etc)

seenitall:
- Add example to README
- Above example needs update in the parameters in the below functions:
`# Initialize transforms (optional)
trfs = T.RandomHorizontalFlip(prob=0.5)

# Get dataset
dataset = get_dataset('fiber-production-inspection', image_set="train", 
                      transforms=trfs,  client=client, poly_to_mask=True, 
                      val_percentage=0.1, test_percentage=0.05, split_seed=42,
                      force_fetching=True)`

seenitall:
https://github.com/v7labs/darwin-py/blob/ec5ff149618580692df5f1b625fe7c417468f568/darwin/pytorch/dataset.py#L116

Get a Value Error when preparing the dataset:
```
ValueError: polygon's area should be <= bbox's area. Failed 7544.0 <= 13474.99999999998
```

Should we raise an error in these cases, or just a skip the annotation and raise a Warning?

almazan:
> https://github.com/v7labs/darwin-py/blob/ec5ff149618580692df5f1b625fe7c417468f568/darwin/pytorch/dataset.py#L116
> 
> Get a Value Error when preparing the dataset:
> 
> ```
> ValueError: polygon's area should be <= bbox's area. Failed 7544.0 <= 13474.99999999998
> ```
> 
> Should we raise an error in these cases, or just a skip the annotation and raise a Warning?

Actually, this should not happen. This was just to check that I'm computing the area correctly, and looks like I'm not. 

Both the area and the bounding box are computed from the polygon. Computing the bounding box is straightforward, and if the area of the polygon is computed correctly it can't be larger than the area of the bounding box.

I'll take a look into this.

seenitall:
It's calculated alright, the order of area and bbox area are just miss-placed in the comparison?

almazan:
> * Add example to README
> * Above example needs update in the parameters in the below functions:
> 
> ```
> # Initialize transforms (optional)
> trfs = T.RandomHorizontalFlip(prob=0.5)
> 
> # Get dataset
> dataset = get_dataset('fiber-production-inspection', image_set="train", 
>                       transforms=trfs,  client=client, poly_to_mask=True, 
>                       val_percentage=0.1, test_percentage=0.05, split_seed=42,
>                       force_fetching=True)
> ```

I've added a default parameter to the `RandomHorizontalFlip` transform with prob=0.5

seenitall:
https://github.com/v7labs/darwin-py/blob/ec5ff149618580692df5f1b625fe7c417468f568/darwin/pytorch/transforms.py#L29

"image" is a PIL Image, it should be "size" instead (ojo: .size has the order inverted w.r.t. .shape):
```
height, width = image.size[-2:]
```
Nevertheless you don't use height. width in the function, so you may want to avoid this line.

darwin/pytorch/utils.py
```diff
@@ -0,0 +1,288 @@
...
+    if client is None:
+        client = Client.default()
+
+    # Get data
+    dataset = None
+    local_datasets = {dataset.slug: dataset for dataset in client.list_local_datasets()}
+    if dataset_name in local_datasets:
+        if force_fetching:
+            # Remove the local copy of the dataset
+            dbpath = os.path.join(client.project_dir,  dataset_name)
```
andreaazzini: Minor comment: we tend to prefer `pathlib.Path` rather than `os.path` overall.

darwin/pytorch/utils.py
```diff
@@ -0,0 +1,288 @@
...
+    for mask, c in zip(masks, classes):
+        color = np.append(colors[c], 128)
+        outmask[mask > threshold] = color
+
+    outmask = Image.fromarray(outmask)
+    image.paste(outmask, (0, 0), outmask)
+    return image
+
+
+def get_colors(n, name='hsv'):
```
andreaazzini: Are we using this function anywhere?

darwin/pytorch/utils.py
```diff
@@ -0,0 +1,288 @@
...
+        if num_test > 0:
+            with open(split_path / 'test.txt', 'w') as f:
+                for i in test_idx:
+                    f.write(f"{annot_files[i].stem}\n")
+
+    return root, split_id
+
+
+# VISUALIZATION
+def visualize_mask_output(image, masks, classes, colors, threshold=0.5):
```
andreaazzini: It's probably premature to include visualization function now, maybe we should do it later in a dedicated module

almazan: I added because I know it's gonna be useful in the future, but I totally agree that it should be in its own dedicated module for visualisations. I'll remove it for the time being.

darwin/pytorch/dataset.py
```diff
@@ -0,0 +1,147 @@
...
+        transforms = T.Compose(trfs)
+
+    return Dataset(root, image_set=image_set, split_id=split_id, transforms=transforms)
+
+
+class Dataset(object):
+    def __init__(
+        self,
+        root: Path,
+        image_set: Optional[str] = "train",
```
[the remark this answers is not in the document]
almazan: Agree. I'll also add checks for the name of the partition and if it exists. But I'd keep it as optional in `get_dataset`.

darwin/pytorch/utils.py
almazan: It's another visualization-related function. I'll remove it as well

darwin/pytorch/utils.py
almazan: Fixed :)

simedw:
let's change then naming from pytorch to torch.

`from darwin.torch.dataset import get_dataset`

andreaazzini:
The code starts look good!
One more thing: we recently agreed on formatting all our Python code with:
- autoflake: removes unused imports and unused variables
- isort: sorts imports
- black: code formatter

As @simedw suggested [here](https://vseven.slack.com/archives/CJUN8TRJA/p1568740965050300), please run `autoflake --in-place --remove-all-unused-imports --remove-unused-variables -r . && python -m isort -y && black -l 100 .`

Another good practice (but I guess out of the scope here) is to run `mypy` for static type checking. It won't fix anything by itself, but gives a good idea about the issues with type annotations in the code.

simedw:
Starting to look good :) 

@almazan I see that you written your own Dataset base class (the `__mul__` part is nice but would probably confuse users), shall we base it on https://pytorch.org/docs/stable/torchvision/datasets.htm would make it a bit more standard?  

Also let's not use abbreviation of function/method variables.
`def extract_classes(files: List, anno_type="polygon"):`
is clearer as
`def extract_classes(files: List, annotation_type="polygon"):`

`def subsample(self, perc: float):`
is clearer as 
`def subsample(self, percentage: float):`

Another that I'm worried about is that `get_dataset(..)` does a lot of things and isn't very composable. It downloads the dataset, sub samples it for train/test/val and then picks the dataset implementation to initialize passing the transforms. I do however like that `mode="raw"` allows for user to create a custom Dataset. 

Could InstanceSegmenation take a dataset as input?
```python
from darwin.torch import Dataset, InstanceSegmentationDataset
dataset = Dataset("birds", image_set="train")
dataset = InstanceSegmentationDataset(dataset, transforms=[...])
```
or
```python
from darwin.torch import DataSource, InstanceSegmentationDataset
datasource = DataSource("birds", image_set="train")
dataset = InstanceSegmentationDataset(datasource, transforms=[...])
```
but then again we don't want to lose the ability to specify root manually.

```python
dataset1 = InstanceSegmentationDataset("/home/project",
   image_set="train",
   transforms=[...]
)
dataset2 = InstanceSegmentationDataset.from_raw(
   DataSource("birds", image_set="train"),
   transforms=[...]
)
```
^-- Not sure about it

Happy for comments/suggestions.

Renthal:
There are two files called `dataset.py`. One in `darwin` and the other in `darwin.torch`. In the PR **Fix Issues #2, #4 and #5** the `darwin.dataset.py` is split into local and remote dataset among other things. Maybe we should consider merging that PR first/as well (or reject it if no longer compatible with the current changes)

darwin/torch/dataset.py
```diff
@@ -0,0 +1,259 @@
...
+    polygon_area,
+)
+
+
+def get_dataset(
+    dataset_name: str,
+    image_set: Optional[str] = "train",
+    mode: Optional[str] = "raw",
+    transforms: Optional = None,
+    poly_to_mask: Optional[bool] = False,
```
Renthal: `poly_to_mask` is unused, is this intended?

darwin/torch/dataset.py
```diff
@@ -26,19 +26,36 @@ def get_dataset(
...
-        transforms: List of PyTorch transforms
-        client: Darwin's client
-        val_percentage: percentage of images used in the validation set
-        test_percentage: percentage of images used in the validation set
-        force_fetching: discard local dataset and pull again from Darwin
-        force_resplit: discard previous split and create a new one
-        split_seed: fix seed for random split creation
-
-    Output:
+    Parameters
```
Renthal: The numpy docstring notation goes by:
```
Parameters
----------
[variable_name] : [variable_type]
       [description]
index : int 
       example of description

Returns
--------
[variable_name] : [variable_type]
[description]
```
The advantage of using it is that then automated tools (such as Pycharm) auto-detect issues between methods signatures and documentation (in this case extra parameters in doc string). I updated the docstring with some TODOs highlighting the missing parts and issues.

darwin/torch/dataset.py
```diff
@@ -0,0 +1,276 @@
...
+            root, image_set=image_set, split_id=split_id, transforms=transforms
+        )
+    else:
+        raise ValueError("Dataset type {mode} not supported.")
+
+    return dataset
+
+
+class Dataset(object):
+    def __init__(self, root: Path, image_set: str, split_id: Optional[str] = None, transforms=None):
```
Renthal: Genuine question, what is the different between: ` split_id: Optional[str] = None` and `transforms=None` in terms of type notation? I mean, `transforms` is also optional from the Python POV as there is a default argument so why would we go using the Optional[type] notation? Follow up question: shouldn't we be consistent and use the same notation for all parameters? I am sorry if its a stupid question I never used this type specifications before.

darwin/torch/dataset.py
```diff
@@ -0,0 +1,276 @@
...
+        anno = [obj for obj in anno if obj["name"] in self.classes]
+
+        res = dict(image_id=idx, annotations=anno)
+        return res
+
+    def __len__(self):
+        return len(self.images)
+
+    def subsample(self, perc: float):
+        num = int(len(self.annotations) * perc)
```
Renthal: this is not rounding but is flooring the float, is this intentional? If all split use floor there is a chance that some images will be missed

darwin/torch/dataset.py
Renthal: I fixed in a later commit by removing it

darwin/torch/dataset.py
Renthal: I also fixed that in a later commit

darwin/torch/utils.py
```diff
@@ -0,0 +1,269 @@
...
+                )
+        else:
+            dataset = local_datasets[dataset_name]
+
+    if dataset is None:  # Could not find it locally or force_fetching is True
+        remote_datasets = [dataset.slug for dataset in client.list_remote_datasets()]
+        if dataset_name in remote_datasets:
+            dataset = client.get_remote_dataset(slug=dataset_name)
+            progress, _count = dataset.pull(image_status=image_status)
+            with tqdm(total=_count, desc=f"Downloading '{dataset_name}' dataset") as pbar:
```
Renthal: Here I had a multi-threaded implementation, maybe could be useful?
```
import multiprocessing as mp
with mp.Pool(mp.cpu_count()) as pool:
   pool.map(run, progress())
```
but to work it requires a small portion of the other PR on issue #6

darwin/torch/utils.py
```diff
@@ -0,0 +1,269 @@
...
+                f.write(f"{c}\n")
+
+    # Create split
+    split_id = f"split_val{val_percentage}_test{test_percentage}"
+    if split_seed is not None:
+        np.random.seed(split_seed)
+        split_id += f"_seed{split_seed}"
+
+    split_path = lists_path / split_id
+    if not split_path.exists() or force_resplit:
```
Renthal: This splitting does not take into account of the classes distribution. If it is guaranteed that the classes exist (from either type of annotation) then we can use the solution proposed above in the other comment?

setup.py
```diff
@@ -16,10 +16,13 @@
         "argcomplete",
         "docutils",
         "humanize",
+        "pycocotools",
         "pyyaml>=5.1",
         "requests",
         "sh",
         "tqdm",
+        "torch",
+        "torchvision",
```
simedw: Actually I don't think we should add torch/torchvision here.
My reasoning is that you could use our cli without ever needing torch installed, so why force it.
I also want use to add support for tensorflow (darwin.tf) in the future, but we should force the user to install both tf and pytorch.

It's better if we add a check in darwin/torch/__init__.py that complains if torch is not available. 
Which is also a reason not to include torch in the darwin/__init__.py

darwin/torch/utils.py
```diff
@@ -0,0 +1,275 @@
...
+    local_datasets = {dataset.slug: dataset for dataset in client.list_local_datasets()}
+    if dataset_name in local_datasets:
+        if force_fetching:
+            # Remove the local copy of the dataset
+            dbpath = Path(client.project_dir, dataset_name)
+            try:
+                shutil.rmtree(dbpath)
+            except PermissionError:
+                print(
+                    "Could not remove dataset in {dbpath}. Permission denied. \
```
andreaazzini: Not an f-string

setup.py
Renthal: Agreed

darwin/torch/utils.py
Renthal: Fixed

PR #12, opened 2019-09-24
Pytorch simplify get db

seenitall:
- Removed get_dataset function and include the custom transformations in each of the CustomDatasets. Now the datasets can be imported and called directly (see README):
```
from darwin.torch.dataset import ClassificationDataset

ClassificationDataset('bird-species', image_set="val", val_percentage=0.25, transforms=trfs, client=client)
```
- Updated README


darwin/torch/dataset.py
```diff
@@ -14,72 +16,17 @@
 )
 
 
-def get_dataset(
```
andreaazzini: After giving it some thoughts, I don't think getting rid of this function is a good idea.

My major arguments are:
1. The logic to fetch the dataset from darwin has been moved to the dataset classes' constructors. I think that our Datasets here should just wrap around Pytorch ones to load images, and load the annotations according to the _mode_ we're at. Basically, they exist on a lower level than the user's API to fetch the data.
2. To add a further point on top of that: supposing that this code is going to be used by Python coders that are not 100% knowledgeable about the internals of Pytorch. Then, having a unique function to know about is much easier than knowing that, since you're doing instance segmentation, you need a specific dataset.

Please let me know if you think these points make sense, as this discussion is rather important to have sooner rather than later, so we can be all on the same page.

@simedw @seenitall

darwin/torch/dataset.py
```diff
@@ -182,8 +130,34 @@ def __str__(self):
         return format_string
 
 
+class BaseDataset(Dataset):
```
andreaazzini: If we don't fetch the dataset in classes' constructors anymore, then we wouldn't need this class

darwin/torch/dataset.py
```diff
@@ -182,8 +130,34 @@ def __str__(self):
...
+        self.classes = []
+
+
 class ClassificationDataset(Dataset):
-    def __init__(self, root: Path, image_set: str, split_id: Optional[str] = None, transforms=None):
+    def __init__(
+        self,
+        dataset_name: str,
+        image_set: str,
+        transforms: Optional = [],
```
andreaazzini: The type for the `transform` field should be `Optional[List[object]]` (if we're lucky, it could also be `Optional[List[Callable]]` but I haven't tried.

It's always good to try running `mypy` on a branch, to see type related warnings.

darwin/torch/dataset.py
```diff
@@ -197,7 +171,21 @@ def _load_anno_and_remap(self, idx: int):
...
+        client: Optional[Client] = None,
+        **kwargs,
+    ):
+
+        root, split_id = fetch_darwin_dataset(dataset_name, client, **kwargs)
+        transforms = self.TRANSFORMS + transforms
```
Renthal: this needs to be wrapped into a compose

darwin/torch/dataset.py
Renthal: I think the logic of downloading the dataset should not be in the dataset itself. The dataset itself should be a Pytorch-like dataset which only provides the logic of loading the data and preparing it for the dataloader to fetch it trough `__get_item__()`. 
The `get_dataset` function behaves very much like a factory method pattern and I think its a clean and scalable solution should there be another type of datasets in the future.

darwin/__init__.py
```diff
@@ -1,4 +0,0 @@
-import darwin.exceptions
-from darwin.client import Client
-from darwin.dataset import Dataset, LocalDataset
-from darwin.team import Team
```
simedw: I don't think we want to delete this.

Renthal:
This could no longer be merged due to conflicts. 
I manually integrated what done in this PR and added some features in the PR #21

PR #17, opened 2019-10-10
Download dataset

Renthal:
`RemoteDataset.pull()` now matches the functionality of the CLI.
The optional parameters `blocking` and `multi_threaded` allow to either get the iterator or use a single thread to download the data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,8 @@ Downloads a remote project, images and annotations, in the projects directory (s
 from darwin.client import Client
 
 client = Client.default()
-dataset = client.get_remote_dataset(slug="example-dataset").pull()
+dataset = client.get_remote_dataset(slug="example-dataset")
+dataset.pull()
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -85,15 +85,12 @@ for _ in progress():
 #### Pull a [remote] project
 
 Downloads a remote project, images and annotations, in the projects directory (specified in the authentication process [default: ~/.darwin/projects]).
+
 ```python
 from darwin.client import Client
 
 client = Client.default()
-dataset = client.get_remote_dataset(slug="example-dataset")
-progress, _count = dataset.pull()
-for f in progress():
-    f()
-    print("file synced")
+dataset = client.get_remote_dataset(slug="example-dataset").pull()
 ```
 
 ### Command line
--- a/README.md
+++ b/README.md
@@ -94,11 +94,9 @@ dataset = client.get_remote_dataset(slug="example-dataset")
 dataset.pull()
 ```
 
-
 ### Command line
 
 `darwin` is also accessible as a command line tool.
-
 
 #### Authentication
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -85,17 +85,15 @@ for _ in progress():
...
 ```python
 
 client = Client.default()
-dataset = client.get_remote_dataset(slug="example-dataset")
-progress, _count = dataset.pull()
-for f in progress():
-    f()
-    print("file synced")
+dataset = client.get_remote_dataset(slug="example-dataset").pull()
```
simedw: I guess it should be 
```
dataset = client.get_remote_dataset(slug="example-dataset")
dataset.pull()
```

setup.py
```diff
@@ -5,7 +5,7 @@
 
 setuptools.setup(
     name="darwin-py",
-    version="0.0.3",
+    version="0.0.4",
```
simedw: we don't need to increase until we have pushed it to pip.

README.md
```diff
@@ -85,17 +85,15 @@ for _ in progress():
...
 ```
 
+
```
andreaazzini: No need for this line if I'm not mistaken

darwin/dataset/remote_dataset.py
```diff
@@ -60,8 +61,28 @@ def _add_files_to_dataset(self, filenames: List, fps: int):
...
             yield
 
-    def pull(self):
-        """Downloads a rermote project (images and annotations) in the projects directory. """
+    @staticmethod
+    def _f(x):
+        """Support function for pool.map() in pull()"""
+        x()
+
+    def pull(self, blocking:Optional[bool] = True, multi_threaded : Optional[bool] = True):
```
andreaazzini: Here, a formatter would probably complain
```suggestion
    def pull(self, blocking: Optional[bool] = True, multi_threaded: Optional[bool] = True):
```

darwin/dataset/remote_dataset.py
```diff
@@ -73,9 +94,20 @@ def pull(self):
...
             )
+            # If blocking is selected, download the dataset on the file system
+            if blocking:
+                if multi_threaded:
+                    import multiprocessing as mp
+                    with mp.Pool(mp.cpu_count()) as pool:
+                        pool.map(RemoteDataset._f, progress())
+                else:
+                    for f in progress():
+                        f()
```
andreaazzini: Perhaps this would be cleaner? (Line 110 would also need to be removed if this suggestion is accepted)
```suggestion
                        f()
                return None, count
            return progress
```

setup.py
Renthal: I manually import this in Gale for the moment hence the upgrade. 
Shall i roll back to 0.0.3 ? (or 0.0.2 ?)

README.md
Renthal: Correct

Renthal:
Progress bar added for all situations, and CLI adapted accordingly

Issue #18, opened 2019-10-10
count from dataset.pull() misbehaves


Renthal:
The number is wrong, and depending if some files are already present on the file system it returns different values.

Renthal:
Fixed in #21

PR #20, opened 2019-10-11
Hot fix

Renthal:
Fix for #18 

The issue was a conflict on names while renaming the annotations to match the file name.
Solution is to temporarily pre-pend all annotations with `V7_` as to ensure that no image will ever have that exact filename. 

This PR needs to be merged to master after #17

### DIFF
--- a/darwin/dataset/remote_dataset.py
+++ b/darwin/dataset/remote_dataset.py
@@ -1,9 +1,10 @@
 from __future__ import annotations
 
 import io
+
 import zipfile
 from pathlib import Path
-from typing import TYPE_CHECKING, List, Optional
+from typing import List, Optional, TYPE_CHECKING
 
 import darwin
 from darwin.dataset.download_manager import download_all_images_from_annotations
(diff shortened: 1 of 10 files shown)


darwin/dataset/remote_dataset.py
```diff
@@ -100,7 +101,10 @@ def pull(self, blocking: Optional[bool] = True, multi_threaded: Optional[bool] =
                     print(f"Could not remove dataset in {annotations_dir}. Permission denied.")
             annotations_dir.mkdir(parents=True, exist_ok=False)
             z.extractall(annotations_dir)
-            annotation_format = "json"
+            # Rename all json files pre-pending 'V7_' in front of them.
+            # This is necessary to avoid overriding them in a later stage.
+            for annotation_path in annotations_dir.glob(f"*.{annotation_format}"):
+                annotation_path.rename(annotation_path.parent / f"V7_{annotation_path.name}")
```
andreaazzini: Whatever prepended string you choose, we should probably get rid of it after all the annotations have been renamed.

Renthal: It is the case :)

PR #21, opened 2019-10-14
Split dataset

Renthal:
* New unified API for pulling a dataset
* Added dataset stratified splitting support 
* Added Pytorch dataset


darwin/torch/dataset.py
```diff
@@ -172,79 +104,86 @@ def _map_annotation(self, index: int):
...
+        self.classes = [e.strip() for e in open(str(self.dataset.local_path / "lists/classes_masks.txt"))]
+        # Prepend the default transform to convert polygons to instance masks
+        if self.transforms is None:
+            self.transforms = []
+        self.transforms.insert(0, T.ConvertPolysToInstanceMasks())
+        self.transforms = T.Compose(transforms)
 
     def _map_annotation(self, index: int):
+        """See superclass for documentation"""
         with open(str(self.annotations_path[index])) as f:
```
simedw: try this instead: `with self.annotation_path[index].open() as f`

darwin/dataset/utils.py
```diff
@@ -0,0 +1,291 @@
...
+        "classes_tags.txt", annotation_files, lists_path, "tag", force_resplit
+    )
+
+    # Create split id, path and final split paths
+    assert val_percentage is not None
+    assert test_percentage is not None
+    assert split_seed is not None
+    assert 0 < val_percentage < 1.0
+    assert 0 <= test_percentage < 1.0
+    assert val_percentage + test_percentage < 1.0
```
simedw: Since `val_percentage`, `test_percentage` and `split_seed` come from the user we shoudn't assert them but instead raise a ValueException

darwin/dataset/utils.py
```diff
@@ -0,0 +1,291 @@
...
+
+    splits = {}
+    # Do the actual split
+    if not split_path.exists() or force_resplit:
+        os.makedirs(str(split_path), exist_ok=True)
+
+        # RANDOM SPLIT
+        train_path = Path(split_path / "random_split_train.txt")
+        val_path = Path(split_path / "random_split_val.txt")
+        test_path = Path(split_path / "random_split_test.txt")
```
simedw: if `split_path` is already of the type Path, why do you need to wrap it around in `Path(split_path / "....")` ?

darwin/dataset/utils.py
```diff
@@ -0,0 +1,291 @@
...
+        splits['random'] = {'train': train_path, 'val':val_path, 'test':test_path}
```
simedw:
```suggestion
        splits['random'] = {'train': train_path, 'val': val_path, 'test': test_path}
```

darwin/dataset/utils.py
```diff
@@ -0,0 +1,291 @@
...
+        # Compute split sizes
+        dataset_size = sum(1 for _ in annotation_files)
```
simedw:
```suggestion
        dataset_size = len(annotation_files)
```

darwin/dataset/utils.py
```diff
@@ -0,0 +1,291 @@
...
+        test_indices = indices[train_size + val_size :]
+        # Write files
+        _write_to_file(annotation_files, train_path, train_indices)
+        _write_to_file(annotation_files, val_path, val_indices)
+        _write_to_file(annotation_files, test_path, test_indices)
+
+        # STRATIFIED SPLIT ON POLYGONS
+        train_path = Path(split_path / "stratified_polygon_train.txt")
+        val_path = Path(split_path / "stratified_polygon_split_val.txt")
+        test_path = Path(split_path / "stratified_polygon_test.txt")
```
simedw: same here `train_path = split_path / "stratified_polygon_train.txt"`

darwin/dataset/utils.py
```diff
@@ -0,0 +1,291 @@
...
+        )
+        # Write files
+        _write_to_file(annotation_files, train_path, train_indices)
+        _write_to_file(annotation_files, val_path, val_indices)
+        _write_to_file(annotation_files, test_path, test_indices)
+
+        # STRATIFIED SPLIT ON TAGS
+        train_path = Path(split_path / "stratified_tags_train.txt")
+        val_path = Path(split_path / "stratified_tags_val.txt")
+        test_path = Path(split_path / "stratified_tags_test.txt")
```
simedw: and here

darwin/dataset/utils.py
```diff
@@ -0,0 +1,291 @@
...
+
+    # Extract list of classes and create respective files
+    lists_path = dataset.local_path / "lists"
+    lists_path.mkdir(parents=True, exist_ok=True)
+    idx_to_classes_polygon = make_class_list(
+        "classes_polygon.txt", annotation_files, lists_path, "polygon", force_resplit, add_background=True
+    )
+    idx_to_classes_tags = make_class_list(
+        "classes_tags.txt", annotation_files, lists_path, "tag", force_resplit
+    )
```
simedw: do we need this for bounding boxes too?

darwin/dataset/utils.py
```diff
@@ -0,0 +1,291 @@
...
+    split_seed : in
+        Fix seed for random split creation
+
+    Returns
+    -------
+    splits : dict
+        Keys are the different splits (random, tags, ...) and values are the relative file names
+    """
+    assert dataset is not None
+    annotation_path = Path(dataset.local_path / "annotations")
```
simedw:
```suggestion
    annotation_path = dataset.local_path / "annotations"
```

darwin/dataset/utils.py
```diff
@@ -0,0 +1,291 @@
...
+    Parameters
+    ----------
+    annotation_files : list
+        List of json files with the GT information of each image
+    file_path : Path
+        Path to the file where to save the list of indices
+    split_idx : Iterable
+        Indices of files for this split
+    """
+    with open(str(file_path), "w") as f:
```
simedw:
```suggestion
    with file_path.open() as f:
```

darwin/torch/dataset.py
```diff
@@ -1,158 +1,90 @@
...
+        if self.transforms is not None:
+            self.transforms = T.Compose(transforms)
 
+        # Populate internal lists of annotations and images paths
+        if not self.split.exists():
+            raise FileNotFoundError(f"Could not find partition {self.split}"
+                                    f" in {self.dataset.local_path}.")
         extensions = ["jpg", "jpeg", "png"]
-        stems = [e.strip() for e in open(str(file_partition))]
+        stems = (e.strip() for e in open(str(self.split)))
```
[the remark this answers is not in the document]
Renthal: Fixed, thanks

darwin/dataset/utils.py
Renthal: Possibly yes; I kept it for maximal flexibility, not knowing what type of datasets we may have in the future this is good to have.

darwin/dataset/utils.py
Renthal: Done

darwin/dataset/utils.py
Renthal: Done

darwin/dataset/utils.py
Renthal: Done

darwin/torch/dataset.py
Renthal: Done

darwin/torch/dataset.py
```diff
@@ -1,158 +1,90 @@
...
 
-    def subsample(self, perc: float):
-        num = int(len(self.annotations_path) * perc)
-        #TODO warning this does not protect from x-contamination! See PR
-        indices = random.sample(range(len(self.annotations_path)), num)
-        self.original_images_path = self.images_path
-        self.images_path = [self.images_path[i] for i in indices]
-        self.original_annotations_path = self.annotations_path
-        self.annotations_path = [self.annotations_path[i] for i in indices]
+    def extend(self, ds, extend_classes: bool = False):
```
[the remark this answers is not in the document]
Renthal: Done

darwin/dataset/utils.py
```diff
@@ -0,0 +1,291 @@
+import json
+import os
+from collections import defaultdict
+from pathlib import Path
+from typing import List, Optional, Iterable
+from sklearn.model_selection import train_test_split
```
andreaazzini: We are only depending on `sklearn` to use `train_test_split`. Is it worth it?

darwin/dataset/utils.py
```diff
@@ -0,0 +1,291 @@
...
+    add_background : bool
+        Add the '__background__' class to the list of classes
+
+    Returns
+    -------
+    idx_to_classes: dict
+    Dictionary where keys are image indices and values are all classes
+    contained in that image
+    """
+    fname = lists_path / file_name
```
andreaazzini: Very minor, but we should use the same naming convention throughout the codebase. Is it `filename`, `file_name` or `fname`? Either way, let's choose one and go for it!

P.S.: my vote would go to `filename` but you can totally ignore that 😊

darwin/dataset/utils.py
```diff
@@ -0,0 +1,291 @@
...
+    if not fname.exists() or force_resplit:
+        classes, idx_to_classes = extract_classes(annotation_files, annotation_type=annotation_type)
+        classes_names = list(classes.keys())
+        if add_background:
+            classes_names.insert(0, "__background__")
```
andreaazzini: 👍

darwin/dataset/utils.py
```diff
@@ -0,0 +1,291 @@
...
+        with open(str(fname), "w") as f:
```
andreaazzini: I don't think you need to cast a `Path` to a `str` when opening it

darwin/torch/dataset.py
```diff
@@ -1,158 +1,90 @@
...
         for stem in stems:
-            annotation_path = self.root / f"annotations/{stem}.json"
+            annotation_path = self.dataset.local_path / f"annotations/{stem}.json"
             for extension in extensions:
-                image_path = self.root / f"images/{stem}.{extension}"
+                image_path = self.dataset.local_path / f"images/{stem}.{extension}"
                 if image_path.is_file():
                     self.images_path.append(image_path)
                     self.annotations_path.append(annotation_path)
                     break
```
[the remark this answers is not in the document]
Renthal: Adopted. Additionally, extra error messages have been added.

darwin/dataset/utils.py
Renthal: You need to if you want to avoid a warning, because open takes a str as input.
However, you are right and I upgraded to `with fname.open() as f:`

darwin/dataset/utils.py
Renthal: Done (`filename` chosen)

darwin/dataset/utils.py
Renthal: Valid point. [Here](https://github.com/scikit-learn/scikit-learn/blob/1495f6924/sklearn/model_selection/_split.py#L1998) you can see the function. I think extracting that would be non-trivial (for me at least). Is there a way to do it nicely?

andreaazzini: If the dependency is not too big, I guess we can always change it later, what do you think @simedw

PR #24, opened 2019-10-24
Darwin dataset fix

seenitall:


darwin/client.py
```diff
@@ -455,7 +455,7 @@ def login(cls, email: str, password: str, projects_dir: Optional[Path] = None):
             refresh_token=data["refresh_token"],
             api_url=api_url,
             base_url=Client.default_base_url(),
-            projects_dir=projects_dir,
+            projects_dir=str(projects_dir),
```
andreaazzini: What would be broken if we still passed `project_dir` as `Path`?
Would it be possible to cast it to `str` where the issue occurs?

darwin/dataset/upload_manager.py
```diff
@@ -74,6 +74,7 @@ def add_files_to_dataset(
             yield
 
         for video_file in data["video_data"]:
+            print(client, video_file, videos)
```
andreaazzini: Let's get rid of `print`s

darwin_py.egg-info/PKG-INFO
```diff
@@ -0,0 +1,261 @@
+Metadata-Version: 2.1
```
andreaazzini: I think we should not version this file. If you agree, please remove it from the staged files with:
`git rm --cached darwin_py.egg-info/PKG-INFO`

darwin/dataset/remote_dataset.py
```diff
@@ -100,6 +100,9 @@ def push(
             )
             if not files_to_upload:
                 raise NotFound("No files to upload, check your path and exclusion filters")
+        elif extensions_to_exclude is not None:
```
andreaazzini: I think `extensions_to_exclude` is an optional argument for `find_files()`, so perhaps we can avoid this `elif` branch and simply do
```python
files_to_upload = find_files(files_list=files_to_upload, exclude=extensions_to_exclude)
```

darwin/dataset/upload_manager.py
```diff
@@ -36,10 +36,10 @@ def add_files_to_dataset(
...
     dataset_id : str
         ID of the dataset to add the files to
-    filenames : list[Path]
-        List of filenames to upload
-    fps : int
-        Number of file per seconds to upload
+    filenames : list[Path]sign_upload
+        List of filenames to uploadsign_upload
+    fps : intsign_upload
+        Number of file per seconds to uploadsign_upload
```
andreaazzini: There are a bunch of weird `sign_upload`s here

darwin/dataset/upload_manager.py
```diff
@@ -88,6 +88,7 @@ def chunk(items, size):
     current_names = set()
     left_over = []
     for item in items:
+        print("item", item)
```
andreaazzini: Please remove `print` here

darwin/utils.py
```diff
@@ -57,7 +57,8 @@ def prompt(msg: str, default: Optional[str] = None) -> str:
 
 
 def find_files(
-        root: Path,
+        root: Optional[Path] = None,
```
andreaazzini: When is this function called with `root=None`?

darwin/utils.py
```diff
@@ -80,22 +81,32 @@ def find_files(
...
+        for file in files_list:
+            file = Path(file)
             if (file.suffix in SUPPORTED_IMAGE_EXTENSIONS + SUPPORTED_VIDEO_EXTENSIONS
                     and file.suffix not in exclude):
                 files += [file]
-    return files
+        return files
+    if root is not None:
+        if not root.is_dir():
+            # print ("TODO: when are we supposed to enter here")
```
andreaazzini: It looks like this branch is designed so that `root` could be a file, rather than a directory. I assume that may happen in a recursive call.

darwin/utils.py
```diff
@@ -80,22 +81,32 @@ def find_files(
...
+            if (root.suffix in SUPPORTED_IMAGE_EXTENSIONS + SUPPORTED_VIDEO_EXTENSIONS
+                    and root.suffix not in exclude):
+                return [root]
+            else:
+                return []
+        files: List[Path] = []
+        for file in root.iterdir():
+            if file.is_dir():
+                if recursive:
+                    files += find_files(file, recursive, exclude)
```
andreaazzini: This recursive call looks wrong, since you added `files_list` as second argument of the `find_files` function.
```suggestion
                    files += find_files(file, recursive=recursive, exclude=exclude)
```
should fix it

darwin/utils.py
```diff
@@ -80,22 +81,32 @@ def find_files(
...
+            else:
+                if (file.suffix in SUPPORTED_IMAGE_EXTENSIONS + SUPPORTED_VIDEO_EXTENSIONS
+                        and file.suffix not in exclude):
+                    files += [file]
```
andreaazzini: I would probably rewrite the `if`s conditions a bit to enhance readability (at least imho)

```python
if file.is_dir() and recursive:
    files += find_files(file, recursive=recursive, exclude=exclude)
elif not file.is_dir() and file.suffix in SUPPORTED_IMAGE_EXTENSIONS + SUPPORTED_VIDEO_EXTENSIONS and file.suffix not in exclude:
    files += [file]
```

(the formatter would surely make a better job in formatting the code though 😁)

darwin/utils.py
seenitall: The cli sends a list of files instead of a root

darwin/utils.py
seenitall: I agree this function needs rewriting but I do not know all the cases we enter it for now. I'll test all the functionality for the other PR

darwin/utils.py
seenitall: Fixed

PR #27, opened 2019-11-01
added upload annotations

Renthal:
Experimental feature to upload annotations from the front end


darwin/dataset/upload_manager.py
```diff
@@ -167,3 +167,62 @@ def sign_upload(client, image_id, key, file_path):
...
+        Path to the folder which contains all the json files representing the annotations to add
+
+    Notes
+    -----
+        This function is experimental and the json files `image_mapping` and `class_mapping` can
+        actually only be retrieved from the backend at the moment.
+    """
+
+    # Read and prepare the image id mappings in a dict format {'class name': 'class id'}
+    with open(str(image_mapping)) as json_file:
```
andreaazzini: You can use `open()` with `Path`s without casting them to strings
```suggestion
    with open(image_mapping) as json_file:
```

darwin/dataset/upload_manager.py
```diff
@@ -167,3 +167,62 @@ def sign_upload(client, image_id, key, file_path):
...
+        image_mapping = {str(cm['original_filename']): cm['id'] for cm in json.load(json_file)}
```
andreaazzini: What happens if you remove the casting here as well?

darwin/dataset/upload_manager.py
```diff
@@ -167,3 +167,62 @@ def sign_upload(client, image_id, key, file_path):
...
+
+    # Read and prepare the class mappings in a dict format {'class name': 'class id'}
+    with open(str(class_mapping)) as json_file:
+        class_mapping = {str(cm['name']): cm['id'] for cm in json.load(json_file)}
```
andreaazzini: Same two comments above would apply here as well, if they make sense

darwin/dataset/upload_manager.py
```diff
@@ -167,3 +167,62 @@ def sign_upload(client, image_id, key, file_path):
...
+
+    # For each annotation found in the folder send out a request
+    for f in annotations_path.glob(f"*.json"):
```
andreaazzini: No need for the f-string
```suggestion
    for f in annotations_path.glob("*.json"):
```

darwin/dataset/upload_manager.py
```diff
@@ -167,3 +167,62 @@ def sign_upload(client, image_id, key, file_path):
...
+                del (annotation['name'])
+                # Compose the list of annotations as the payload wants
+                payload_annotations.append({
+                    'annotation_class_id':  class_id,
+                    'data': annotation
+                 })
+            payload = {"annotations": payload_annotations}
+            # Compose the endpoint
+            endpoint = f"dataset_images/{image_mapping[data['image']['filename']]}/annotations"
+            response = client.put(endpoint=endpoint, payload=payload, retry=True)
```
andreaazzini: Brilliant ✨

darwin/dataset/upload_manager.py
Renthal: Probably nothing but I thought better safe than sorry, in case of weird file names

PR #28, opened 2019-11-04
Upload dataset fix

Renthal:
Upload dataset now actually works (with multi thread support as well)

### DIFF
--- a/darwin/dataset/upload_manager.py
+++ b/darwin/dataset/upload_manager.py
@@ -1,12 +1,16 @@
 import functools
 import itertools
-from pathlib import Path
-from typing import Any, Dict, List, Optional
 import json
+from pathlib import Path
+from typing import Any, Dict, List, Optional, TYPE_CHECKING
+
 import requests
 
 from darwin.exceptions import UnsupportedFileType
 from darwin.utils import SUPPORTED_IMAGE_EXTENSIONS, SUPPORTED_VIDEO_EXTENSIONS
+
+if TYPE_CHECKING:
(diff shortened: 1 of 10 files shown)


darwin/dataset/upload_manager.py
```diff
@@ -43,72 +31,194 @@ def add_files_to_dataset(
...
     current_list = []
     current_names = set()
     left_over = []
-    for item in items:
-        name = Path(item).name
-        if name in current_names:
-            left_over.append(item)
+    for file in files:
+        if file.name in current_names:
+            #TODO I don't fully understand this logic. We enter here IFF there is the same
```
andreaazzini: I think it's worth removing the `TODO` from the code, and ask it to the rest of the team, so we make sure everybody is on the same page

simedw: This code is a bit weird but this is the general gist of it:

We send the filenames to the backend before we start uploading the data, if we sent the full file path then it's easy to know which dataset_image_id belongs to which file on disk (we get a list of {dataset_image_id and full_file_path back). 
But we don't want the full path to be sent to the server since it will be displayed in the UI and users won't like that .

So we need to just send the filename, but then if we have more than one file with that filename  (in the same upload: e.g. /cat/1.png and /dot/1.png (since both are 1.png)) we don't know which dataset_image_id goes to which image. 

We solved that by making sure they they aren't chunked together, e.g. /cat/1.png and /dog/1.png will go in two different chunks. 

There is probably a better solution to this :)

PR #29, opened 2019-11-05
Resume upload

Renthal:
It is now possible to resume a dataset.push() and the error messages should be collected in the repsonses json file

### DIFF
--- a/darwin/cli_functions.py
+++ b/darwin/cli_functions.py
@@ -203,7 +203,7 @@ def remove_remote_project(project_slug: str):
 def upload_data(
     project_slug: str,
     files: Optional[List[str]],
-    extensions_to_exclude: Optional[List[str]],
+    files_to_exclude: Optional[List[str]],
     fps: Optional[int],
 ):
     """Uploads the files provided as parameter to the remote dataset selected
@@ -214,15 +214,15 @@ def upload_data(
         Slug of the project to retrieve
     files : list[str]
         List of files to upload. Can be None.
-    extensions_to_exclude : list[str]
-        List of extension to exclude from the file scan (which is done only if files is None)
+    files_to_exclude : list[str]
+        List of files to exclude from the file scan (which is done only if files is None)
     fps : int
         Number of files per second to upload
     """
     client = _load_client()
     try:
         dataset = client.get_remote_dataset(slug=project_slug)
-        dataset.push(extensions_to_exclude=extensions_to_exclude, fps=fps, files_to_upload=files)
+        dataset.push(files_to_exclude=files_to_exclude, fps=fps, files_to_upload=files)
     except NotFound:
         _error(f"No dataset with name '{project_slug}'")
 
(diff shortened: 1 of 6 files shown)


darwin/dataset/remote_dataset.py
```diff
@@ -90,30 +97,55 @@ def push(
         count : int
             The files count
         """
-        if files_to_upload is None and not self.local_path.exists():
+        # Resolving where to look for images
+        if source_folder is None:
+            source_folder = self.local_path / "images"
+        # This is where the responses from the upload function will be saved/load for resume
+        responses_path = (source_folder.parent / "upload_responses.json")
```
andreaazzini: No need for parentheses

darwin/dataset/upload_manager.py
```diff
@@ -106,6 +108,20 @@ def _chunk_filenames(files: List[Path], size: int):
     No batch has any duplicates with regards to file name.
     This is needed due to a limitation in the upload api.
 
+    This code is a bit weird but this is the general gist of it:
```
andreaazzini: I would personally keep documentation a little more formal, while keeping more explanatory and informal language in comments, when necessary

darwin/dataset/upload_manager.py
```diff
@@ -215,66 +233,46 @@ def upload_file_to_s3(
...
-    file_path
-
-    Returns
-    -------
-    requests.Response
-    Response of the post request
-    """
-    test = {"file": open(file_path, "rb")}
-    return requests.post("http:" + end_point, data=signature, files=test)
+    return requests.post("http:" + end_point, data=signature, files={"file": file_path.open('rb')})
```
andreaazzini: Perhaps it would be good practice to actually determine which response we get, and do something in the case of 4** or 5** status codes

darwin/dataset/remote_dataset.py
```diff
@@ -61,11 +61,13 @@ def __init__(
 
     def push(
         self,
-        blocking: Optional[bool] = True,
-        multi_threaded: Optional[bool] = True,
-        extensions_to_exclude: Optional[List[str]] = None,
-        fps: Optional[int] = mp.cpu_count(),
+        blocking: bool = True,
+        multi_threaded: bool = True,
+        fps: int = 1,
```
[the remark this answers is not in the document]
Renthal: Yes, apparently fps is not files per second but rather frame per second and has thus nothing to do with the number of cpus

darwin/dataset/remote_dataset.py
Renthal: Yep, fixed

darwin/dataset/upload_manager.py
Renthal: Addressed

darwin/cli_functions.py
```diff
@@ -203,7 +203,7 @@ def remove_remote_project(project_slug: str):
 def upload_data(
     project_slug: str,
     files: Optional[List[str]],
-    extensions_to_exclude: Optional[List[str]],
```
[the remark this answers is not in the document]
Renthal: We are not. By having a list of files to exclude it's rather easy to feed all files belonging to a particular extension (e.g. `ls | grep -e "*.txt"` sort of command or `glob` if you are in python code)

darwin/dataset/upload_manager.py
Renthal: These gets logged later on in the pipeline

PR #33, opened 2019-11-06
upload annotations + multi-thread upload + resume

Renthal:


darwin/client.py
```diff
@@ -77,7 +77,12 @@ def get(self, endpoint: str, retry: bool = False, raw: bool = False):
             return self.get(endpoint=endpoint, retry=False)
 
         if response.status_code != 200:
-            print("TODO, fix me get", response.json(), response.status_code)
+            print(
+                f"Client get request response ({response.json()}) with unexpected status "
+                f"({response.status_code}). "
+                f"Client: ({self})"
+                f"Request: (endpoint={endpoint})"
+            )
```
andreaazzini: So my personal opinion is that these prints shouldn't be there by default, but there should be a flag that would be checked in order to figure out if printing debug information. Something like `--debug` or `--verbose`

Renthal: Previous version was printing a TODO. I would suggest that work on that on a specific PR which addresses failed post/put/get requests if you agree.

andreaazzini: It's ok if we deal with it in a separate PR, but let's not print stuff here then, when it could lead to data leaking

Renthal: Agreed. Commented out.

PR #48, opened 2019-12-10
API Keys

simedw:
- [x] authentication 
- [x] team
- [x] create
- [x] remove
- [x] remote
- [x] push
- [x] version
- [x] report
- [x] pull
- [x] release
- [x] local
- [x] refactor so that team_slug / dataset_slug get split inside the api instead of in cli_functions
- [ ] store version of dataset downloaded


darwin/cli.py
```diff
@@ -22,61 +35,77 @@ def main():
...
-        projects_dir = prompt("Project directory", "~/.darwin/projects")
-        f.authenticate(email, password, projects_dir)
-        print("Authentication succeeded. ")
+        api_key = getpass.getpass(prompt="API key: ", stream=None)
+        if api_key.strip() == "":
+            print(
+                "API Key needed, generate one for your team: https://darwin.v7labs.com/?settings=api-keys"
+            )
+            return
+        default_team = input("Make this team default? [y/N] ") in ["Y", "y"]
```
Renthal: As per discussion, I leave a note: make the call first and prompt the team name to the user to make it more intuitive

darwin/cli.py
```diff
@@ -22,61 +35,77 @@ def main():
...
-        f.remove_remote_project(project_name)
-
-    # Upload new data to a project (remotely)
-    elif args.command == "push":
-        f.upload_data(args.project_name, args.files, args.exclude, args.fps)
-
     # Version
     elif args.command == "version":
-        print("0.0.1")
+        print("0.3")
```
Renthal: Isn't is 0.0.3 ? (as per `setup.py`)

darwin/cli.py
```diff
@@ -22,61 +35,77 @@ def main():
...
+
+    elif args.command == "dataset":
+        # List existing projects (remotely)
```
Renthal:
```suggestion
```

darwin/cli.py
```diff
@@ -22,61 +35,78 @@ def main():
...
-    elif args.command == "push":
-        f.upload_data(args.project_name, args.files, args.exclude, args.fps)
-
     # Version
     elif args.command == "version":
-        print("0.0.1")
+        print("0.3")
+
+    elif args.command == "dataset":
+        if args.action == "remote":
```
Renthal: What about `list` as a command name instead or remote to match e.g. conda's commands or `ls` to match Docker?

darwin/cli.py
```diff
@@ -22,61 +35,78 @@ def main():
...
-        path = f.path(args.project_name)
-        print(path)
-
-    # Print the url of a remote project
-    elif args.command == "url":
-        url = f.url(args.project_name)
-        print(url)
-
     # Download dataset
     elif args.command == "pull":
```
Renthal: should we not have this inside the elif args.command == "dataset" below to match the API of other operations such as push, create remove and so on?

darwin/cli_functions.py
```diff
@@ -20,33 +20,34 @@
 from darwin.utils import find_files, persist_client_configuration, secure_continue_request
 
 
-def authenticate(email: str, password: str, projects_dir: str) -> Config:
+def authenticate(api_key: str, datasets_dir: str, default_team: bool) -> Config:
     """Authenticate user against the server and creates a configuration file for it
```
Renthal:
```suggestion
    """Authenticate the API key against the server and creates a configuration file for it
```

darwin/cli_functions.py
```diff
@@ -20,33 +20,34 @@
...
 
     Parameters
     ----------
-    email : str
-        Email to use for the client login
-    password : str
-        Password to use for the client login
-    projects_dir : str
+    team : str
+        Teams to use for the client login
```
Renthal: this parameter is not in the signature

Renthal: and `default_team : bool` is missing

darwin/cli_functions.py
```diff
@@ -20,33 +20,34 @@
...
-        client = Client.login(email=email, password=password, projects_dir=projects_dir)
+        client = Client.login(api_key=api_key, datasets_dir=datasets_dir)
         config_path = Path.home() / ".darwin" / "config.yaml"
         config_path.parent.mkdir(exist_ok=True)
-        return persist_client_configuration(client)
+        default_team = client.team if default_team else None
+        return persist_client_configuration(client, default_team=default_team)
 
     except InvalidLogin:
         _error("Invalid credentials")
```
Renthal: Should the error message hint to the API key or we leave a generic "credentials" ?

darwin/config.py
```diff
@@ -4,61 +4,107 @@
...
+
         if save:
             self._save()
 
     def _save(self):
         """Persist the configuration to the file system"""
         with io.open(self._path, "w", encoding="utf8") as f:
             yaml.dump(self._data, f, default_flow_style=False, allow_unicode=True)
+
+    def set_team(self, team: str, api_key: str):
```
Renthal: Proposed renaming: `set_team_api_key`

darwin/config.py
```diff
@@ -4,61 +4,107 @@
...
+
+    def set_default_team(self, team: str):
+        if self.get(f"teams/{team}") is None:
+            raise InvalidTeam()
+        self.put("global/default_team", team)
+
+    def set_global(
+        self, api_endpoint: str, base_url: str, directory: str, default_team: Optional[str] = None
+    ):
+        self.put("global/api_endpoint", api_endpoint)
```
Renthal: Here we could benefit for a docstring ti explain better api_endpoint and base_url format/content expectation maybe

darwin/config.py
```diff
@@ -4,61 +4,107 @@
...
+        self.put("global/base_url", base_url)
+        self.put("global/datasets_dir", directory)
+        if default_team:
+            self.put("global/default_team", default_team)
+
+    def get_team(self, team: str):
```
Renthal: team could be optional and if None we return `get_default_team`

darwin/cli.py
simedw: will update setup.py

darwin/config.py
```diff
@@ -4,61 +4,107 @@
...
+        )
+        return {"slug": team, "api_key": api_key, "default": default}
+
+    def get_default_team(self):
+        default_team = self.get("global/default_team")
+        if default_team:
+            return self.get_team(default_team)
+        teams = list(self.get("teams").keys())
+        if len(teams) > 1:
+            raise InvalidTeam()
```
Renthal: We could make the error more specific. In this case "invalid team" is a bit weird since the user asks for the default one. Maybe we could say something like "no default team set" or along these lines

darwin/cli.py
simedw: the reason we went for remote was that we want to clearly show the difference between datasets on the server vs datasets on the local machine. I don't think docker lets you list remote images.

But another name than remote might be good.

darwin/cli.py
simedw: only local and pull are left, but that's because I haven't implemented them, waiting for andrea to finish the new export code.

darwin/dataset/remote_dataset.py
```diff
@@ -19,10 +19,10 @@ class RemoteDataset:
     def __init__(
         self,
         *,
+        team: str,
         name: str,
         slug: Optional[str] = None,
```
Renthal: we could make slug a @property which calls a _slugify (xD) method on name. In this way we can ensure consistency between name and slug by design

darwin/config.py
simedw: it now takes one more argument so I think we can keep it as set_team

darwin/dataset/remote_dataset.py
```diff
@@ -42,19 +42,17 @@ def __init__(
             spaces are replaced by dashes, e.g., `bird-species`. This string is unique within a team
         dataset_id : int
```
Renthal: team : str is missing

darwin/dataset/remote_dataset.py
```diff
@@ -42,19 +42,17 @@ def __init__(
...
         image_count : int
             Dataset size (number of images)
         progress : float
             How much of the dataset has been annotated 0.0 to 1.0 (1.0 == 100%)
         client : Client
             Client to use for interaction with the server
         """
+        self.team = team
         self.name = name
         self.slug = slug or name
```
Renthal: this could be dangerous and create failing requests further down the pipeline. Could be fixed with the property comment made above

darwin/dataset/remote_dataset.py
simedw: probably is that the slug could been changed later on so the name and slug doesn't match any more.

darwin/dataset/remote_dataset.py
```diff
@@ -205,17 +197,23 @@ def pull(
 
     def remove_remote(self):
         """Archives (soft-deletion) the remote dataset"""
-        self.client.put(f"projects/{self.project_id}/archive", payload={})
+        self.client.put(f"datasets/{self.dataset_id}/archive", payload={})
+
+    def get_report(self, granularity="day"):
```
Renthal:
```suggestion
    def get_report(self, granularity: str = "day"):
```

Renthal: could benefit from a docstring as the purpose of this method is not trivial (at least to me)

darwin/utils.py
```diff
@@ -82,34 +79,35 @@ def find_files(
     list[Path]
     List of all files belonging to supported extensions. Can't return None.
     """
-    if files_list is None and root is None:
-        raise ValueError(f"Invalid combined value for root (None) and files_list (None)")
 
     # Init the return value
```
Renthal:
```suggestion
```

darwin/dataset/identifier.py
```diff
@@ -0,0 +1,39 @@
+# class DatasetIdentifier:
+#     def __init__(self, slug):
+#         self.team_slug, self.dataset_slug, self.version = parse(slug)
```
Renthal: should we remove this?

darwin/dataset/release.py
```diff
@@ -0,0 +1,66 @@
+import datetime
+import shutil
+
+import requests
+
+from darwin.dataset.identifier import DatasetIdentifier
+
+
+class Release:
```
Renthal: why not use dataclass here too then ?

darwin/cli.py
```diff
@@ -22,61 +36,73 @@ def main():
...
+                for action in dataset_parser._actions
+                if isinstance(action, argparse._SubParsersAction)
+            ]
+
+            print(dataset_parser.description)
+            print("\nCommands:")
+            for subparsers_action in subparsers_actions:
+                # get all subparsers and print help
+                for choice in sorted(subparsers_action._choices_actions, key=lambda x: x.dest):
+                    print("    {:<19} {}".format(choice.dest, choice.help))
```
andreaazzini: Wouldn't still make sense to move this piece of code somewhere else?

darwin/cli_functions.py
```diff
@@ -17,36 +18,49 @@
...
         config_path = Path.home() / ".darwin" / "config.yaml"
         config_path.parent.mkdir(exist_ok=True)
-        return persist_client_configuration(client)
+
+        if default_team is None:
+            default_team = input(f"Make {client.default_team} the default team? [y/N] ") in ["Y", "y"]
+        if datasets_dir is None:
+            datasets_dir = prompt("Datasets directory", "~/.darwin/datasets")
+
+        datasets_dir = Path(os.path.expanduser(datasets_dir))
```
andreaazzini: We are always trying to use `pathlib` functions in place of `os.path` ones
In this case: https://docs.python.org/3/library/pathlib.html#pathlib.Path.home
```suggestion
        datasets_dir = Path.home() / datasets_dir
```

simedw: I totally agree with preferring Path, but this is actually not the same thing in this case.
 
`expanduser` replaces `~` with the home path, if there is a `~` in the string, but if there isn't it won't do anything. 

While Path.home() always adds the home directory to the path. (Imagine `datasets_dir = "/datasets/"`)

But I just learned from the docs that Path has `expanduser`


```suggestion
        datasets_dir = Path(datasets_dir).expanduser()
```

andreaazzini: Ah-ha, I misunderstood what you used `expanduser` for, but still a useful suggestion

image

darwin/cli.py
simedw: done

PR #53, opened 2019-12-18
Export dataset from CLI

Renthal:
Enabling the creation of a release from CLI

New:
- [x] CLI command for creating a release 
- [x] Create release function matches the new logic of Andrea's filtering


darwin/dataset/remote_dataset.py
```diff
@@ -260,11 +260,13 @@ def get_report(self, granularity="day"):
             raw=True,
         ).text
 
-    def release(self, name: Optional[str] = None):
+    def release(self, annotation_class_ids: Optional[List] = None, name: Optional[str] = None):
```
simedw: I don't think name is optional anymore. @andreaazzini ?

Renthal: it is not below, but for the python function it is. In fact, if it is None then empty list is provided to the payload (I simulated a requested and inspected with the networking tab)

darwin/dataset/remote_dataset.py
```diff
@@ -273,9 +269,20 @@ def release(self, name: Optional[str] = None):
         release: Release
             The release created right now
         """
-        self.client.post(f"/datasets/{self.dataset_id}/exports", team=self.team)
+        if annotation_class_ids is None:
+            annotation_class_ids = []
+        payload = {"annotation_class_ids": annotation_class_ids, "name": name}
+        self.client.post(f"/datasets/{self.dataset_id}/exports", payload=payload, team=self.team)
```
simedw: should probably use `error_handlers=[name_taken, validation_error]`

darwin/dataset/remote_dataset.py
```diff
@@ -273,9 +269,20 @@ def release(self, name: Optional[str] = None):
...
+        time.sleep(2.0) # This is necessary avoid a race condition on pulling the latest release
```
simedw: Exporting could take > 10 min to run if the dataset is big enough, what we would get back is an in progress export record that isn't very useful. 
Sleeps are in general not a good way to solve this issues

darwin/options.py
```diff
@@ -83,41 +85,56 @@ def __init__(self):
...
+            "export", help="Export the a version of a dataset."
+        )
+        parser_export.add_argument("dataset", type=str, help="Remote dataset name to export.")
+        parser_export.add_argument(
+            "annotation_class",
+            type=str,
+            nargs='?',
+            help="List of class filters"
+        )
+        parser_export.add_argument("name", type=str, help="Name with with the version gets tagged.")
```
simedw: How about putting the name first, since the classes are optional

darwin/dataset/remote_dataset.py
```diff
@@ -253,18 +254,13 @@ def remove_remote(self):
...
 
-    def get_report(self, granularity="day"):
-        return self.client.get(
-            f"/reports/{self.dataset_id}/annotation?group_by=dataset,user&dataset_ids={self.dataset_id}&granularity={granularity}&format=csv&include=dataset.name,user.first_name,user.last_name,user.email",
-            team=self.team,
-            raw=True,
-        ).text
-
-    def release(self, name: Optional[str] = None):
+    def export(self, annotation_class_ids: Optional[List] = None, name: Optional[str] = None):
```
simedw: Name shouldn't be optional (otherwise the backend sends: `{'errors': {'name': ["can't be blank"]}}`
I would also change `Optional[List]` to `Optional[List[str]]`
and change the order of the arguments: `export(self, name: str, annotation_class_ids: Optional[List[str]])`

simedw: Also the name is more like a slug (it doesn't have spaces) @andreaazzini was that check only done on the frontend side? I seem to be able to create "test test" ?

darwin/dataset/remote_dataset.py
simedw: we should also add an error handler for `{'errors': {'name': ['is reserved']}}`

Renthal: I added the `error_handlers=[name_taken, validation_error]` but I am not entirely certain what you mean by the `{'errors': {'name': ['is reserved']}}` (respectively, I am not sure how to address it?)

darwin/dataset/remote_dataset.py
Renthal: I adopted the proposed changes. Regarding the name vs slug notation, I used name because that's what's written in the payload for the post request sniffed from the front end, but it can surely be changed.

darwin/options.py
Renthal: done

darwin/dataset/remote_dataset.py
Renthal: Nice catch, I was not entirely sure what was the reason of the race condition. Since we can't know in advance how long is it gonna take and we can't also use a blocking solution (at least yet?) I removed the return statement (which was anyway not used).

darwin/cli.py
```diff
@@ -69,6 +69,9 @@ def run(args, parser):
             f.remove_remote_dataset(args.dataset)
         elif args.action == "report":
             f.dataset_report(args.dataset, args.granularity or "day")
+        # Create a new release for that dataset
```
andreaazzini: This is probably droppable

README.md
```diff
@@ -111,10 +111,24 @@ $ darwin dataset push test /path/to/folder/with/images
 100%|████████████████████████| 2/2 [00:01<00:00,  1.27it/s] 
 ```
 
-Downloads a remote project, images and annotations, in the projects directory 
-(specified in the authentication process [default: `~/.darwin/projects`]).
+
+Before we can download a dataset we first need to generate a release
```
andreaazzini: We could probably rephrase this line (possibly drop `we`)

darwin/exceptions.py
```diff
@@ -38,4 +38,4 @@ class ValidationError(Exception):
 
 
 class Unauthorized(Exception):
-    pass
+    pass
```
andreaazzini: Isn't the formatter complaining about this?

PR #56, opened 2020-01-13
New feature: default split

almazan:
This PR introduces the concept of default splits for Darwin datasets, and adds a flag to make a given split the default one. The default split will be always located under `DB_ROOT/lists/split` in the local copy of the dataset.

### DIFF
--- a/darwin/dataset/utils.py
+++ b/darwin/dataset/utils.py
@@ -213,6 +213,7 @@ def split_dataset(
     test_percentage: Optional[float] = 0.2,
     force_resplit: Optional[bool] = False,
     split_seed: Optional[int] = 0,
+    make_default_split: Optional[bool] = True
 ):
     """
     Given a local a dataset (pulled from Darwin) creates lists of file names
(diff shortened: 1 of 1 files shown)


darwin/dataset/utils.py
```diff
@@ -333,6 +336,15 @@ def split_dataset(
...
 
+    # Create symlink for default split
+    cwd = os.getcwd()
+    os.chdir(lists_path)
+    if not os.path.exists('split') or make_default_split:
+        if os.path.exists('split'):
+            os.unlink('split')
+        os.symlink(split_id, 'split')
+    os.chdir(cwd)
+
```
simedw: What do you think about this instead?
```python
split = lists_path / 'split'
if make_default_split or not split.exists():
   split.unlink(missing_ok=True)
   split.symlink_to(lists_path  / split_id)
```

almazan: Updated. I added an if statement because `missing_ok` is only supported in python 3.8.

PR #59, opened 2020-01-16
Refactor util functions and extend RemoteDataset

almazan:
This PR refactors the function `get_annotations` (previously `get_annotation_dicts`) in order to improve its usability and readability. It also adds a new function `get_classes` that given a Darwin dataset and a class type it returns the list of classes in the dataset.

Finally, it also extends `RemoteDataset` with three new functions, `classes`, `split`, and `annotations`, which call the respective util functions.


darwin/dataset/remote_dataset.py
```diff
@@ -330,6 +339,97 @@ def get_release(self, name: str = "latest"):
                 return release
         raise NotFound(self.identifier)
 
+    def split(
+        self,
+        val_percentage: float = 0.1,
+        test_percentage: float = 0,
+        force_resplit: bool = False,
+        split_seed: int = 0,
+        make_default_split: bool = True,
```
andreaazzini: These defaults values here make a split not being generated silently, which was unexpected to both me and @simedw when we first used the function.
I would, if you agree (later):
1. Generate a split every time (splits may have a uuid)
2. The default split is just a link to the latest split

@almazan you also mentioned we have a bug when generating splits with same seed? Could you elaborate more on how to replicate the issue?

setup.py
```diff
@@ -15,14 +15,15 @@
...
         "humanize",
         "pyyaml>=5.1",
         "requests",
+        "scikit-learn",
         "sh",
         "tqdm",
-        "factory_boy",
     ],
-    packages=["darwin"],
+    packages=setuptools.find_packages(),
```
andreaazzini: Without this line, our nested packages (`dataset` and `torch`) won't be installed

darwin/utils.py
```diff
@@ -3,7 +3,7 @@
 
 from darwin.config import Config
 
-SUPPORTED_IMAGE_EXTENSIONS = [".png", ".jpeg", ".jpg", ".PNG", ".JPEG", ".JPG"]
+SUPPORTED_IMAGE_EXTENSIONS = [".png", ".jpeg", ".jpg", ".jfif", ".PNG", ".JPEG", ".JPG", ".JFIF"]
```
andreaazzini: Some Android files use this format: https://en.wikipedia.org/wiki/JPEG_File_Interchange_Format

darwin/dataset/remote_dataset.py
simedw: how about adding the export name/version to the split name, so that when we download a new version of the same dataset we get a new split.

darwin/dataset/remote_dataset.py
```diff
@@ -330,6 +339,97 @@ def get_release(self, name: str = "latest"):
...
+        force_resplit : bool
+            Discard previous split and create a new one
+        split_seed : int
+            Fix seed for random split creation
+        make_default_split: bool
+            Makes this split the default split
+        """
+        if not self.local_path.exists():
+            raise NotFound(
+                "Local dataset not foungd: the split is perfomed on the local copy of the dataset. \
```
simedw:
```suggestion
                "Local dataset not found: the split is performed on the local copy of the dataset. \
```

darwin/dataset/remote_dataset.py
```diff
@@ -330,6 +339,97 @@ def get_release(self, name: str = "latest"):
...
+        ----------
+        annotation_type
+            The type of annotation classes, e.g. 'tag' or 'polygon'
+
+        Returns
+        -------
+        classes: list
+            List of classes in the dataset of type `class_type`
+        """
+        assert self.local_path.exists()
```
simedw: don't really want to crash here? rather raise an error or re-generate the class list.

darwin/dataset/upload_manager.py
```diff
@@ -351,7 +339,7 @@ def upload_annotations(
                             client=client,
                             team=team,
                             annotation_type_ids=[
-                                "1"
+                                "3"
```
simedw: no magic numbers :)

darwin/torch/dataset.py
```diff
@@ -32,16 +33,16 @@ def __init__(self, root: Path, split: Path, transform: Optional[List] = None):
...
         # Compose the transform if necessary
         if self.transform is not None and isinstance(self.transform, list):
-            self.transform = T.Compose(transform)
+            self.transform = Compose(transform)
 
         # Populate internal lists of annotations and images paths
         if not self.split.exists():
             raise FileNotFoundError(f"Could not find partition file: {self.split}")
-        extensions = [".jpg", ".jpeg", ".png"]
+        extensions = SUPPORTED_IMAGE_EXTENSIONS
```
simedw: only use `SUPPORTED_IMAGE_EXTENSIONS` ?

darwin/torch/dataset.py
```diff
@@ -330,7 +331,7 @@ def measure_weights(self, **kwargs) -> np.ndarray:
         """
         # Collect all the labels by iterating over the whole dataset
         labels = []
-        for i, filename in enumerate(self.images_path):
+        for i in range(len(self.images_path)):
```
simedw:
```suggestion
        for i, _filename in enumerate(self.images_path):
```

darwin/torch/transforms.py
```diff
@@ -1,24 +1,74 @@
...
+        return image, target
+
+
+class ToTensor(transforms.ToTensor):
+    def __call__(self, image, target: Optional[TargetType] = None):
+        if target is None:
+            return super(ToTensor, self).__call__(image)
+
+        image = F.to_tensor(image)
+        return image, target
```
simedw:
```suggestion
        image = super(ToTensor, self).__call__(image)
        if target is None:
             return image
         return image, target
```

darwin/torch/transforms.py
```diff
@@ -1,24 +1,74 @@
...
+
+
+class ColorJitter(transforms.ColorJitter):
+    def __call__(self, image, target: Optional[TargetType] = None):
+        if target is None:
+            return super(ColorJitter, self).__call__(image)
+
+        transform = self.get_params(self.brightness, self.contrast, self.saturation, self.hue)
+        image = transform(image)
+        return image, target
```
simedw: same here

darwin/torch/transforms.py
```diff
@@ -1,24 +1,74 @@
...
+    def __call__(self, image: Image, target: Optional[TargetType] = None):
+        for transform in self.transforms:
+            image, target = transform(image, target)
+        return image, target
 
 
-class Compose(object):
-    def __init__(self, transforms):
-        self.transforms = transforms
+class RandomHorizontalFlip(transforms.RandomHorizontalFlip):
```
simedw: how about def __init__(self, p) and then not using RandomHorizontalFlip as base class?

darwin/utils.py
simedw:
```
def allowed_extension(ext):
    return ext.lower() in SUPPORTED_IMAGE_EXTENSIONS
```

darwin/dataset/remote_dataset.py
almazan: I don't think we need a uuid for each split, since the percentages and the seed are enough to differentiate the splits. If you want a different split using the same percentage you just need to use a different seed: it will create a new folder with the format name `split_vX_tY_sZ` where X is the training percentage, Y the test percentage, and Z the seed. If the seed is 0 then the name is `split_vX_tY`. The default split `split` always points to the latest split created unless `make_default_split` is `False`.

So what I would do is to remove `force_resplit`. I will also make sure that the split is consistent when using the same seed, since I think I saw once an unexpected behaviour.

In order to work with different dataset versions I would also add the export name to the split name, as Simon suggests.

almazan: > how about adding the export name/version to the split name, so that when we download a new version of the same dataset we get a new split.

I just gave this solution a try and it's a bit tricky. The problem is that if we want different export versions to coexist locally, we will also need different folders for their annotations, eg `annotations_0.1` and `annotations_0.2`. Or for example, what if we use the following structure?

dataset_name:
  - images
  - 0.1:
    - annotations
    - lists
  - 0.2:
    - annotations
    - lists

simedw: I think the images would be different to per version, you might have removed or added images. 
That's why we talked about having sym links before:

```
images/
    0.jpg
    1.jpg
    2.jpg
version/
   0.1/
        images/
              0.jpg (sym link)
              1.jpg (sym link)
       annotations/
              ....
   0.1/
        images/
              0.jpg (sym link)
              1.jpg (sym link)
              2.jpg (sym link)
       annotations/
              ....
```

almazan: We can add those symlinks but I don't think they would be necessary since you know what images belong to a given export from the list of annotations.

andreaazzini: You're talking of something like: the images directory is a sort of pool, and the versions only include annotations which point to existing images? I think that should work, I can't think of any edge case, and might look cleaner that having symlinks all around
```
images/
    0.jpg
    1.jpg
    2.jpg
    3.jpg
versions/
    0.1/
        annotations/
            0.json
            1.json
    0.2/
        annotations/
            0.json
            1.json
            2.json
            3.json
```

almazan: Yeah, exactly like that. I'll give it a try then.

almazan: These changes are now part of a new PR: https://github.com/v7labs/darwin-py/pull/67

darwin/dataset/upload_manager.py
andreaazzini: A way of avoiding magic numbers is fetching annotation types from the backend, but since they're _rather_ static, we could just replicate an annotation type id/name mapping and use it.
```python
annotation_types = {..., 'polygon': 3, ...}
```

PR #65, opened 2020-02-21
New token info render + format

andreaazzini:


darwin/client.py
```diff
@@ -381,8 +381,7 @@ def from_api_key(cls, api_key: str, datasets_dir: Optional[Path] = None):
         if response.status_code != 200:
             raise InvalidLogin()
         data = response.json()
-        team_id = data["selected_team"]["id"]
-        team = [team["slug"] for team in data["teams"] if team["id"] == team_id][0]
+        team = data["selected_team"]["slug"]
```
andreaazzini: This is actually the only change that matters, but black formatted other parts of the codebase

PR #70, opened 2020-04-22
[HOTFIX] Fix doubling paths to be uploaded on Windows

andreaazzini:


darwin/cli.py
```diff
@@ -44,7 +44,7 @@ def run(args, parser):
             f.list_teams()
     # Version
     elif args.command == "version":
-        print("0.3")
+        print("0.4.2")
```
andreaazzini, line 47: I noticed we hadn't changed this, so I fixed it

PR #75, opened 2020-05-15
Support for local releases

almazan:
This PR adds support in order to be able to have different local releases of a dataset. The local dataset structure is the following:

```
team-slug/
  dataset-slug/
    images/
      0.jpg
      1.jpg
      2.jpg
      3.jpg
    releases/
      0.1/
        annotations/
            0.json
            1.json
        lists/
      0.2/
        annotations/
            0.json
            1.json
            2.json
            3.json
        lists/
      latest -> 0.2 (symlink)
```

It also includes a migration tool to migrate old-formatted datasets to the new format.


darwin/cli_functions.py
```diff
@@ -145,23 +146,54 @@ def local():
...
+            return p
+
+    for p in client.list_deprecated_local_datasets(team=identifier.team_slug):
+        if identifier.dataset_slug == p.name:
+            print(
+                f"{p} (deprecated format) \n"
+                f"Warning: found local version of the dataset {identifier.dataset_slug} which uses a deprecated format. "
+                f"Run `darwin dataset migrate {identifier}` if you want to be able to use it in darwin-py."
+            )
+            return
```
simedw, line 190: should we return something or error out? since it's printed in cli.py

darwin/cli_functions.py
```diff
@@ -237,6 +269,45 @@ def pull_dataset(dataset_slug: str):
...
+    """Migrates an outdated local dataset to the latest format.
+
+    Parameters
+    ----------
+    dataset_slug: str
+        Slug of the dataset to which we perform the operation on
+    """
+    identifier = DatasetIdentifier.parse(dataset_slug)
+    client = _load_client(offline=True)
+    for p in client.list_deprecated_local_datasets(identifier.team_slug):
```
simedw, line 282: while this check is great, what about just checking if the dataset is already migrated first, so we can say "already migrated".

darwin/dataset/utils.py
```diff
@@ -33,6 +80,10 @@ def extract_classes(annotations_path: Path, annotation_type: str):
     Dictionary where keys are image indices and values are all classes
     contained in that image
     """
+    assert annotation_type in ["tag", "polygon", "box", "bounding_box"]
+    if annotation_type == "box":
```
simedw, line 84: when do we have "box" as an argument?

PR #76, opened 2020-05-18
Complex polygons

andreaazzini:


darwin/torch/dataset.py
```diff
@@ -389,14 +389,12 @@ def _map_annotation(self, index: int):
...
         for annotation in annotations:
-            assert "name" in annotation
-            assert "polygon" in annotation
-            # Extract the sequences of coordinates from the polygon annotation
-            sequences = convert_polygons_to_sequences(annotation["polygon"]["path"])
-            # Discard polygons with less than three points
-            sequences[:] = [s for s in sequences if len(s) >= 6]
-            if not sequences:
+            if "polygon" not in annotation and "complex_polygon" not in annotation:
+                print(f"Warning: missing polygon in annotation {self.annotations_path[index]}")
```
andreaazzini, line 393: I don't think assertions belong here tbh, I'd much rather print a warning and continue.
Let me know if you agree

darwin/torch/dataset.py
```diff
@@ -406,12 +404,11 @@ def _map_annotation(self, index: int):
             max_y = np.max([np.max(y_coord) for y_coord in y_coords])
             w = max_x - min_x + 1
             h = max_y - min_y + 1
-            bbox_area = w * h
             # Compute the area of the polygon
+            # TODO fix with addictive/subtractive paths in complex polygons
```
andreaazzini, line 408: Do you have any suggestions? Is `area` useful? If so, we better fix it sooner rather than later

darwin/types.py
```diff
@@ -0,0 +1,6 @@
+from typing import Dict, List
+
+Point = Dict[str, float]
+BoundingBox = Dict[str, float]
+Polygon = List[Point]
+ComplexPolygon = List[Polygon]
```
andreaazzini, line 6: Do you like something like this? I used it in `darwin/torch/utils.py`, and to me it looks a bit better than `Union[List[Point], List[List[Point]]`

andreaazzini: Unfortunately, `TypedDict` would have been better, but it's only available in `typing` since Python 3.8

darwin/types.py
```diff
@@ -0,0 +1,6 @@
+from typing import Dict, List
+
+Point = Dict[str, float]
+BoundingBox = Dict[str, float]
```
andreaazzini, line 4: `TypedDict` would have been better, but it's only available in typing since Python 3.8

darwin/torch/__init__.py
```diff
@@ -1,4 +1,4 @@
-# Requirements: pytorch, torchvision, pycocotools, sklearn
+# Requirements: pytorch, torchvision, sklearn
```
andreaazzini, line 1: I don't think this line is useful. Can I get rid of it?

darwin/types.py
simedw: I like it, I feel however that types.py and datatypes.py should be merged and more interlinked

darwin/torch/dataset.py
```diff
@@ -389,14 +389,12 @@ def _map_annotation(self, index: int):
...
-            sequences = convert_polygons_to_sequences(annotation["polygon"]["path"])
-            # Discard polygons with less than three points
-            sequences[:] = [s for s in sequences if len(s) >= 6]
-            if not sequences:
+            if "polygon" not in annotation and "complex_polygon" not in annotation:
+                print(f"Warning: missing polygon in annotation {self.annotations_path[index]}")
                 continue
+            # Extract the sequences of coordinates from the polygon annotation
+            annotation_type = "polygon" if "polygon" in annotation else "complex_polygon"
+            sequences = convert_polygons_to_sequences(annotation[annotation_type]["path"])
```
almazan, line 397: We should also add support for complex polygons in the `SemanticSegmentationDataset` (line 494).

PR #77, opened 2020-05-20
Local dataset

almazan:
This PR adds a new class `LocalDataset` that can be used as an API for the local versions of datasets and their json annotations.

Now, Pytorch dataloaders make use of this API and extend it for the different CV tasks.


darwin/torch/dataset.py
```diff
@@ -63,194 +68,7 @@ def get_dataset(
...
-        if self.classes is not None:
-            annotation = [a for a in annotation if a["name"] in self.classes]
-        return {
-            "image_id": index,
-            "original_filename": self.images_path[index],
-            "annotations": annotation,
-        }
+class DarwinDataset(LocalDataset):
 
     def measure_mean_std(self, multi_threaded: bool = True):
```
andreaazzini, line 73: Looks like `DarwinDataset` is just a `LocalDataset` with `measure_mean_std` in addition to it.
Why don't we use `DarwinDataset` right away instead? In other words, is there a reason to keep `LocalDataset` separate from this `dataset.py`?

almazan: The main idea for `LocalDataset` is to be a parser of the json files with some basic functionality that can be used in the different modules in `darwin-py`: right now it's only used by the `torch` module, but eventually could be used by the `tf` module, `dl` module, or any other function that needs to access to the json annotations. That's why `LocalDataset` is in `darwin.dataset` and not in `darwin.torch`.

The reason I created `DarwinDataset` in `darwin.torch` is because `measure_mean_std` requires the `pillow` module (or `cv2`) to load images, so I didn't want to add that dependency outside of `darwin.torch`. If you think it's okay adding this dependency to `darwin-py` then I can move those functions to `LocalDataset` and get rid of `DarwinDataset`.

andreaazzini: No, I think yours is a valid point, let's keep it this way then!

darwin/dataset/local_dataset.py
```diff
@@ -0,0 +1,192 @@
+import json
+import multiprocessing as mp
```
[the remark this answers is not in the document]
almazan: Yeah, you're right, `LocalDataset` doesn't use `multiprocessing`. Just run auto formatting and got rid of it (plus many more).

PR #79, opened 2020-05-26
Convert masks

almazan:
This PR adds two new convert formats: `semantic-mask` and `instance-mask`. It also organises all the converted formats of a dataset under `DATASET/releases/RELEASE_NAME/other_formats`.

```
darwin dataset convert v7-demo/bird-species instance-mask
```

```
darwin dataset convert v7-demo/bird-species semantic-mask
```


darwin/dataset/local_dataset.py
```diff
@@ -169,6 +173,93 @@ def parse_json(self, index: int):
...
+        """
+        class_support = np.unique(labels, return_counts=True)[1]
+        class_frequencies = class_support / len(labels)
+        # Class weights are the inverse of the class frequencies
+        class_weights = 1 / class_frequencies
+        # Normalize vector to sum up to 1.0 (in case the Loss function does not do it)
+        class_weights /= class_weights.sum()
+        return class_weights
+
+    # Loads an image with OpenCV and returns the channel wise means of the image.
```
simedw, line 249: Correct me if I'm wrong but we aren't using OpenCV, right?

darwin/exporter/formats/instance_mask.py
```diff
@@ -0,0 +1,99 @@
...
+                sequence = convert_polygons_to_sequences(annotation.data["path"], height, width)
+                mask = convert_polygons_to_mask(sequence, height, width) * 255
+                mask = Image.fromarray(mask.astype(np.uint8))
+                mask_id = f"{image_id}_{instance_id}"
+                mask.save(masks_dir / f"{mask_id}.png")
+                f.write(f"{image_id},{mask_id},{cat}\n")
+        print(f"Dataset format saved at {output_dir}")
+
+
+def convert_polygons_to_sequences(
```
simedw, line 49: we have this function in both utils and in exporter/formats/coco

darwin/options.py
```diff
@@ -131,7 +131,7 @@ def __init__(self):
         )
         parser_convert.add_argument("format", type=str, help="Annotation import to convert to")
 
-        parser_convert.add_argument("output_dir", type=str, help="Where to store output files")
+        parser_convert.add_argument("-o", "--output_dir", type=str, help="Where to store output files")
```
simedw, line 134: this is a rather major change as it would break existing usage the conversion in cli, let's discuss offline

darwin/dataset/local_dataset.py
```diff
@@ -169,6 +173,93 @@ def parse_json(self, index: int):
             "annotations": annotations,
         }
 
+    def measure_mean_std(self, multi_threaded: bool = True):
```
andreaazzini, line 176: I saw this function used elsewhere, if it's exactly the same let's make sure we are smart with code duplication

darwin/dataset/utils.py
```diff
@@ -130,7 +131,7 @@ def make_class_lists(release_path: Path):
 
 
 def get_classes(
-    dataset_path: Path,
+    dataset_path: Union[Path, str],
```
andreaazzini, line 134: Are we sure we need `str` here? I see you're rightly doing `dataset_path = Path(dataset_path)`

darwin/dataset/utils.py
```diff
@@ -667,3 +668,40 @@ def get_annotations(
             with annotation_path.open() as f:
                 record = json.load(f)
             yield record
+
+
+def load_pil_image(path: Path):
```
andreaazzini, line 673: It's not the first time I see this function (I reckon we use it in Gale). Again, let's make sure we don't replicate this code elsewhere, otherwise this function looks good

darwin/exporter/formats/instance_mask.py
```diff
@@ -0,0 +1,99 @@
...
+    masks_dir = output_dir / "masks"
+    if masks_dir.exists():
+        shutil.rmtree(masks_dir)
+    masks_dir.mkdir(parents=True)
+    instance_ids = set()
+    with open(output_dir / "instance_mask_annotations.csv", "w") as f:
+        f.write("image_id,mask_id,class_name\n")
+        pbar = tqdm(list(annotation_files))
+        pbar.set_description(desc="Processing annotations", refresh=True)
+        for annotation_file in pbar:
```
andreaazzini, line 31: I like this! Perhaps we could wrap this logic around a util function, so we can reuse it elsewhere

darwin/exporter/formats/semantic_mask.py
```diff
@@ -0,0 +1,121 @@
...
+def export(annotation_files: Generator[dt.AnnotationFile, None, None], output_dir: Path):
+    masks_dir = output_dir / "masks"
+    masks_dir.mkdir(exist_ok=True, parents=True)
+    annotation_files = list(annotation_files)
+    categories = calculate_categories(annotation_files)
+    ignore_idx = 255
+    pbar = tqdm(annotation_files)
+    pbar.set_description(desc="Processing annotations", refresh=True)
+    for annotation_file in pbar:
+        outfile = masks_dir / (annotation_file.path.stem + ".png")
```
andreaazzini, line 21: I always prefer string interpolation over concatenation, but not too important
```suggestion
        outfile = masks_dir / f'{annotation_file.path.stem}.png')
```

setup.py
```diff
@@ -24,6 +24,7 @@
         "scikit-learn",
         "sh",
         "tqdm",
+        "pillow",
```
andreaazzini, line 27: Warning: very annoying suggestion

Let's order these alphabetically 😇

darwin/dataset/utils.py
almazan: Since I'm also allowing `str` (because I'm immediately converting it to `Path`) don't I need to specify that in the header?

darwin/dataset/utils.py
almazan: It's exactly the same function we use in Gale. But since we're not going to have Gale as a dependency of `darwin-py` that's why I copied here. Does Gale have `darwin-py` as a dependency already?

darwin/dataset/local_dataset.py
almazan: I think you saw that function in the class `DarwinDataset` in `darwin.torch.dataset`, which has now become `LocalDataset` in `darwin.dataset`. I'll double check anyways.

darwin/dataset/local_dataset.py
almazan: Yeah, we're using Pillow now. Forgot to update the comment.

darwin/exporter/formats/instance_mask.py
almazan: Good idea! Done!

darwin/exporter/formats/instance_mask.py
almazan: Sorry, didn't realise. I re-factor it and only left the one at `darwin.utils`.

darwin/utils.py
```diff
@@ -63,6 +67,12 @@ def is_deprecated_project_dir(project_path: Path) -> bool:
     return (project_path / "annotations").exists() and (project_path / "images").exists()
 
 
+def get_progress_bar(array: List, description: Optional[str] = None):
+    pbar = tqdm(array)
```
andreaazzini, line 71: Does `tqdm()` work with generators as well (not just lists)? If so, no reason to convert it to `List` before using it as an argument for `get_progress_bar`. Otherwise, it's fine.
Anyway, not blocking!

PR #82, opened 2020-06-02
Add PyTorch bindings documentation

almazan:


simedw:
Shouldn't we be able to get rid of darwin/torch/README.md now since it's in the docs?

darwin/torch/README.md
```diff
@@ -11,7 +11,7 @@ Here is an example of how to load the `bird-species` dataset ready to be used in
 ```python
 from darwin.torch import get_dataset
 
-db = get_dataset("/datasets/bird-species", dataset_type="instance_segmentation")
+db = get_dataset("/my-datasets/bird-species", dataset_type="instance_segmentation")
```
andreaazzini, line 14: I don't really like the `my-` in `my-datasets`, because of two reasons:
1. The default directory name is `datasets`, so this code won't be copy-pasteable right away for many users
2. A few lines above we wrote `get_dataset("/PATH/TO/YOUR/DATASET", DATASET_TYPE [, PARTITION, SPLIT_TYPE, RELEASE_NAME, TRANSFORMS])`, so it's already clear from the context that the path should be changed if you are not using `datasets`

docs/torch.md
```diff
@@ -1 +1,40 @@
 # PyTorch bindings
+
```
andreaazzini, line 2: Warning: dumb question 😊

Why are we keeping two identical files for the pytorch bindings?

simedw: I also commented on that, We should keep the torch.md and remove the torch/README.md

We should also thin out the main README.md and make it link to the documentation, just include the most important pieces (installation, authentication, upload / download).

darwin/torch/README.md
almazan: And what about `$DATASETS` instead of `my-datasets`? Because I don't think many users will have a folder called `/datasets` like we do.

docs/torch.md
almazan: Yeah, makes total sense. I'll remove the one in `torch`.

darwin/torch/README.md
almazan: Sorry, I thought those were cli commands. I changed it to `/datasets/`.

simedw:
we need to remove the reference to the deleted torch md file

PR #90, opened 2020-06-10
Darwin importer

almazan:


simedw:
looks good, let's just rename `_parse_darwin_json` to `parse_darwin_json` (I only use _ in the beginning of function names if I don't want them to be used outside the module, which was the case before moving it to utils.)

PR #95, opened 2020-06-29
CLI 'dataset pull' options: 'only-annotations' and 'remove-extra'

almazan:


simedw:
I'm not a big fan of 'remove-extra' since the name is very none descriptive, and it only looks at the current pull, there could be an old version (with annotations) pulled that is referencing images that will now be deleted. Let's not included it or change the name.

almazan:
I removed the flag `--remove-extra` and changed the option in `dataset.pull()` to `False` as default.

PR #97, opened 2020-07-07
cvat attribute convert

zacwellmer:
converts darwin attributes to something like cvat attributes. They're a bit different b/c darwin tags only have a name


simedw:
Hi @zacwellmer thank you for your PR, we introduced something very similar but also added support for __text and __instance_ids

zacwellmer:
hmm not sure what you mean? This PR does not change the functionality of darwin's attributes. This PR just makes sure the cvat conversion includes the attribute information

PR #100, opened 2020-07-10
Measure class weights in LocalDataset

almazan:
This PR fixes a bug in the `measure_weights` functions in `LocalDataset`.


darwin/dataset/local_dataset.py
```diff
@@ -275,3 +281,6 @@ def __str__(self):
             f"  Number of images: {len(self.images_path)}\n"
             f"  Number of classes: {len(self.classes)}"
         )
+
+    def __add__(self, dataset):
+        return self.extend(dataset, extend_classes=True)
```
andreaazzini, line 286: I'm not sure I like the idea of using `__add__`, which is less explicit than `self.extend(dataset, extend_classes=True)`

darwin/torch/dataset.py
```diff
@@ -96,23 +95,35 @@ def __getitem__(self, index: int):
...
         if self.transform is not None:
             img = self.transform(img)
 
+        target = self.get_target(index)
+
+        return img, target
+
+    def get_target(self, index: int):
+        """Returns the classification target
+        """
```
andreaazzini, line 108: Honestly not sure if a formatter would make this change. In any case, let's make sure we run the formatter!

```suggestion
        """
        Returns the classification target
        """
```

andreaazzini (changes requested):
Just a question: do we need to override `LocalDataset.__add__`? Do we actually use that anywhere?

Otherwise, just make sure to run a formatter and I'll approve

darwin/dataset/local_dataset.py
almazan: I think I used it in the auto-annotate code, but not anymore, so I can safely remove it.

andreaazzini (approved):

PR #101, opened 2020-07-11
Replace tqdm with rich.progress

almazan:
This PR replaces `tqdm`'s with `rich.progress` progress bars. It also adds a new helper function in `darwin.utils` called `get_progress_bar` that abstracts the progress bar API.

### DIFF
--- a/darwin/utils.py
+++ b/darwin/utils.py
@@ -3,7 +3,6 @@ from pathlib import Path
 from typing import TYPE_CHECKING, List, Optional, Union
 
 import numpy as np
-from tqdm import tqdm
 from upolygon import draw_polygon
 
 import darwin.datatypes as dt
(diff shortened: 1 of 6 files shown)


darwin/utils.py
```diff
@@ -67,9 +66,14 @@ def is_deprecated_project_dir(project_path: Path) -> bool:
...
 
 
-def get_progress_bar(array: List, description: Optional[str] = None):
-    pbar = tqdm(array)
-    pbar.set_description(desc=description, refresh=True)
+def get_progress_bar(array: List, description: str = "Progress", total: int = None):
+    try:
+        from rich.progress import track
+        pbar = track(array, description=description, total=total)
+    except ImportError:
```
almazan, line 74: Probably we don't need this try-except clause since `tqdm` is a dependency anyways.

PR #107, opened 2020-08-25
support for uploading tags to videos

simedw:


darwin/datatypes.py
```diff
@@ -82,3 +82,25 @@ def make_attributes(attributes):
 
 def make_text(text):
     return SubAnnotation("text", text)
+
+
+def make_keyframe(annotation, idx):
```
andreaazzini, line 87: It doesn't look we're using these new functions here, am I missing anything?

simedw: good catch missed a file.

darwin/importer/formats/csvtagsvideo.py
```diff
@@ -0,0 +1,36 @@
...
+    files = []
+
+    file_annotation_map = {}
+    with path.open() as f:
+        reader = csv.reader(f)
+        for row in reader:
+            try:
+                filename, tag, start_frame, end_frame = map(lambda s: s.strip(), row)
+            except ValueError:
+                continue
```
andreaazzini, line 21: In the future, we could log warnings and errors in a log file, which would be useful when debugging.
Right now, we can simply add a TODO comment since we're probably in a rush

darwin/importer/formats/csvtagsvideo.py
```diff
@@ -0,0 +1,36 @@
...
+            if filename == "":
```
andreaazzini, line 22: I'd add a comment about why we would end up here, and not in the `except` branch above (i.e. why is `filename` empty in some files)

PR #110, opened 2020-10-06
Get annotations in Darwin format

almazan:
Adds an option to `RemoteDataset.annotations` to be able to retrieve the annotations using Darwin format.


simedw:
Already implemented

PR #114, opened 2020-10-22
Parse Darwin Video

almazan:


darwin/cli_functions.py
```diff
@@ -546,7 +546,8 @@ def dataset_convert(dataset_slug, format, output_dir):
         annotations_path = release_path / "annotations"
         if output_dir is None:
             output_dir = release_path / "other_formats" / f"{format}"
-            output_dir.mkdir(parents=True, exist_ok=True)
+        output_dir = Path(output_dir)
```
andreaazzini, line 549: `output_dir` should already be a `Path`, or at least that's what we seem to do in line 548. 

If that's not the case, it's because we call `dataset_convert` with an `output_dir` of a different type (likely a string)

I would suggest:
1. annotate the function definition with `typing`
2. make sure we always call this function with the right types (`mypy` would help at finding mistakes on our end)

darwin/datatypes.py
```diff
@@ -32,6 +32,37 @@ def get_sub(self, annotation_type: str) -> Optional[SubAnnotation]:
...
+@dataclass(frozen=True, eq=True)
+class VideoAnnotation:
+    annotation_class: AnnotationClass
+    frames: Annotation
+    keyframes: List[bool]
+    segments: List[List[int]]
+    interpolated: bool
+
+    def get_frame(self, frame_index):
+        return frames[frame_index]
```
andreaazzini, line 44: It would be awesome if we could annotate types here, too.
It's also useful for review and documentation.

For example: is `frame_index` an integer or a string? I guess a string, but I would have to check that we are consistent every time we call this function to be 100% sure.

darwin/datatypes.py
```diff
@@ -43,6 +74,8 @@ class AnnotationFile:
     image_url: Optional[str] = None
     workview_url: Optional[str] = None
     seq: Optional[int] = None
+    is_video: Optional[bool] = None
```
andreaazzini, line 77: I think we can safely make this required, and default it to `false`
```suggestion
    is_video: bool = false
```
Let me know if you think that would be a wrong idea

darwin/exceptions.py
```diff
@@ -45,3 +45,6 @@ class ValidationError(Exception):
 
 class Unauthorized(Exception):
     pass
+
+class OutdatedDarwinJSONFormat(Exception):
+    pass
```
andreaazzini, line 50: Let's add a newline here (I think the formatter would do it, so if we haven't formatted the whole pr, this is our chance 😊)

darwin/datatypes.py
```diff
@@ -32,6 +32,37 @@ def get_sub(self, annotation_type: str) -> Optional[SubAnnotation]:
...
+    keyframes: List[bool]
+    segments: List[List[int]]
+    interpolated: bool
+
+    def get_frame(self, frame_index):
+        return frames[frame_index]
+
+    def get_data(self, only_keyframes=True, post_processing=None):
+        if not post_processing:
+            post_processing = id
```
andreaazzini, line 48: I find this function a bit difficult to read (it might be because I'm not the most pythonic reviewer in the world).

However, I think it would be extremely beneficial to add a comment or two to explain why this `if` branch is needed.

Documenting the function parameters would also be a plus (e.g. what is `post_processing` supposed to do?)

darwin/datatypes.py
```diff
@@ -32,6 +32,37 @@ def get_sub(self, annotation_type: str) -> Optional[SubAnnotation]:
...
+        return {
+            "frames": {
+                frame: {
+                    **post_processing(
+                        self.frames[frame],
```
andreaazzini, line 53: We use `self.frames[frame]` three times here, it's making things a bit difficult to read.

I'm not sure if I have any particular suggestion here, because we are using a dict comprehension and that's fine by me.

Maybe having a separate function that builds this dictionary would help

darwin/datatypes.py
```diff
@@ -104,3 +141,11 @@ def make_video(keyframes, start, end):
             "segments": [[start, end]],
         },
     )
+
+
+def make_video_annotation(frames, keyframes, segments, interpolated):
+    first_annotation = list(frames.values())[0]
+    if not all(frame.annotation_class.name == first_annotation.annotation_class.name for frame in frames.values()):
+        raise ValueError("invalid argument to make_video_annotation")
```
andreaazzini, line 149: Makes sense. Two questions:

1. Has this inconsistency ever happened to our data? That'd be pretty bad.
2. Do we gain anything by having yet another custom exception for these cases? If so, it's worth defining one.

darwin/exporter/formats/instance_mask.py
```diff
@@ -17,7 +18,7 @@ def export(annotation_files: Generator[dt.AnnotationFile, None, None], output_di
     with open(output_dir / "instance_mask_annotations.csv", "w") as f:
         f.write("image_id,mask_id,class_name\n")
         for annotation_file in get_progress_bar(list(annotation_files), "Processing annotations"):
-            image_id = annotation_file.path.stem
+            image_id = os.path.splitext(annotation_file.filename)[0]
```
andreaazzini, line 21: I have two questions here:

1. If `path.stem` doesn't work, is it because `annotation_file.path` is wrong, or not a path?
2. `os.path` is a bit old, and pretty much anything can now be delegated to `pathlib` https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

darwin/exporter/formats/semantic_mask.py
```diff
@@ -36,7 +37,9 @@ def export(annotation_files: Generator[dt.AnnotationFile, None, None], output_di
         RGB_colors = [c for e in RGB_colors for c in e]
 
     for annotation_file in get_progress_bar(list(annotation_files), "Processing annotations"):
-        outfile = masks_dir / f"{annotation_file.path.stem}.png"
+        image_id = os.path.splitext(annotation_file.filename)[0]
```
andreaazzini, line 40: Same comment about `os.path` vs `pathlib` here

darwin/importer/importer.py
```diff
@@ -112,35 +112,45 @@ def import_annotations(
...
     serialized_annotations = []
     for annotation in annotations:
         annotation_class = annotation.annotation_class
         annotation_class_id = remote_classes[annotation_class.annotation_type][annotation_class.name]
-        if "frames" in annotation.data:
+        if isinstance(annotation, dt.VideoAnnotation):
+            data = annotation.get_data(
+                only_keyframes=True, post_processing=lambda annotation, data: _handle_subs(annotation, data, attributes)
+            )
+        elif "frames" in annotation.data:
```
andreaazzini, line 143: Is this branch still needed?

darwin/importer/importer.py
```diff
@@ -112,35 +112,45 @@ def import_annotations(
...
-            else:
-                data[sub.annotation_type] = sub.data
+            data = _handle_subs(annotation, data)
+
         serialized_annotations.append({"annotation_class_id": annotation_class_id, "data": data})
 
     if client.feature_enabled("WORKFLOW", dataset.team):
         res = client.post(f"/dataset_items/{id}/import", payload={"annotations": serialized_annotations})
-        if res["status_code"] != 200:
+        if res.get("status_code") != 200:
```
andreaazzini, line 153: 👍

darwin/utils.py
```diff
@@ -168,25 +169,62 @@ def persist_client_configuration(
...
-            data["image"]["original_filename"],
-            annotation_classes,
-            annotations,
-            data["image"]["width"],
-            data["image"]["height"],
-            data["image"]["url"],
-            data["image"].get("workview_url"),
-            data["image"].get("seq", count),
-        )
+        if "fps" in data["image"] or "frame_count" in data["image"]:
```
andreaazzini, line 184: This makes perfect sense with the way we currently export videos.
As we discussed separately, it's probably worse including the item type within the format, so we don't have to do this.

darwin/utils.py
```diff
@@ -207,6 +245,8 @@ def parse_darwin_annotation(annotation: dict):
         main_annotation = dt.make_line(name, annotation["line"]["path"])
     elif "keypoint" in annotation:
         main_annotation = dt.make_keypoint(name, annotation["keypoint"]["x"], annotation["keypoint"]["y"])
+    elif "skeleton" in annotation:
+        main_annotation = dt.make_skeleton(name, annotation["skeleton"]["nodes"])
```
andreaazzini, line 249: This is totally fine by me. Remember we don't have the annotation class metadata, where the definitions of nodes and edges are stored. For that, we had a manifest PR, which I don't think we ever got to merge @simedw

andreaazzini (changes requested):

darwin/datatypes.py
simedw: It's more to protect users of the SDK to build invalid annotations

darwin/importer/importer.py
simedw: Good point, I think we need to migrate the cvs video tag importer to use the new format and then we could drop that Elif branch

darwin/datatypes.py
simedw: I agree, however, we should then imo reorder it so the optional fields are last

darwin/datatypes.py
simedw: Maybe we can make the default value for post_processing id (identity function) in the function header, maybe that's clearer?

darwin/cli_functions.py
almazan: Yes, you're right. I annotated the function and removed that line.

darwin/datatypes.py
almazan: Annotated this function too. `frame_index` are ints but I'd also need to check that we are consistent with that.

darwin/exporter/formats/instance_mask.py
almazan: Two reasons why I'm using `os.path`:

1. To infer image_id now I'm using filename instead of path, which is a string, so I'd had to convert it to Path anyways
2. If the annotation is part of a video annotation, `filename` has the format `{video_filename}/{frame:07d}.jpg`, so if I use `stem` I would discard the first part, which I need to keep.

So the `pathlib` solution would be: `Path(annotation_file.filename).with_suffix('')`. I thought the `os.path` way looked a bit less convoluted, but I can change it if you like this way better.

darwin/datatypes.py
almazan: Agree too. I changed it and reordered.

darwin/importer/importer.py
almazan: Migrated and removed.

darwin/datatypes.py
```diff
@@ -32,17 +32,50 @@ def get_sub(self, annotation_type: str) -> Optional[SubAnnotation]:
...
+    keyframes: List[bool]
+    segments: List[List[int]]
+    interpolated: bool
+
+    def get_frame(self, frame_index: int):
+        return frames[frame_index]
+
+    def get_data(self, only_keyframes=True, post_processing=None):
+        if not post_processing:
+            post_processing = lambda annotation, data: data
```
andreaazzini, line 48: Makes much more sense now, let's add a comment about why we need it

darwin/datatypes.py
```diff
@@ -32,17 +32,50 @@ def get_sub(self, annotation_type: str) -> Optional[SubAnnotation]:
...
+    annotation_class: AnnotationClass
+    frames: Annotation
+    keyframes: List[bool]
+    segments: List[List[int]]
+    interpolated: bool
+
+    def get_frame(self, frame_index: int):
+        return frames[frame_index]
+
+    def get_data(self, only_keyframes=True, post_processing=None):
```
andreaazzini, line 46: It would be great if we could add type annotations here as well. I think the following makes sense:
```suggestion
    def get_data(
      self,
      only_keyframes: bool = True,
      post_processing: Optional[Callable[[Annotation, dict], dict]] = None,
    ):
```

darwin/importer/formats/csvtagsvideo.py
```diff
@@ -25,12 +25,16 @@ def parse_file(path: Path) -> Optional[List[dt.AnnotationFile]]:
             start_frame = int(start_frame)
             end_frame = int(end_frame)
 
-            annotation = dt.make_video([dt.make_keyframe(dt.make_tag(tag), start_frame)], start_frame, end_frame)
+            annotation = dt.make_tag(tag)
+            frames = {i: annotation for i in range(start_frame, end_frame + 1)}
+            keyframes = {i: i == start_frame for i in range(start_frame, end_frame + 1)}
```
andreaazzini, line 30: Are we sure about the `+ 1` here?

andreaazzini (changes requested):
I think we're very close! Only a couple of comments about type annotations and comments, only one comment about potential logic errors

darwin/importer/formats/csvtagsvideo.py
almazan: I'm not sure about that. @simedw ?

darwin/cli_functions.py
```diff
@@ -546,7 +546,7 @@ def dataset_convert(dataset_slug, format, output_dir):
         annotations_path = release_path / "annotations"
         if output_dir is None:
             output_dir = release_path / "other_formats" / f"{format}"
-            output_dir.mkdir(parents=True, exist_ok=True)
+        output_dir.mkdir(parents=True, exist_ok=True)
```
andreaazzini, line 549: I think we have a bug here. We added the type annotation for `output_dir`, but that doesn't guarantee that the argument is actually of that type.

Make sure that each time we call `dataset_convert` we actually make `output_dir` a `Path` (I don't think we're doing that, but I might be wrong)

andreaazzini (changes requested):
I found another potential bug

darwin/importer/formats/csvtagsvideo.py
simedw: yeah, otherwise it's not inclusive:
```py
>>> list(range(1,3))
[1, 2]
```

almazan: What I didn't know is if the end frame indicated in the csvtag is included or not. It it like this `[start, end]` or like this `[start, end)`?

darwin/utils.py
```diff
@@ -255,8 +253,13 @@ def parse_darwin_annotation(annotation: dict):
...
         main_annotation = dt.make_keypoint(name, annotation["keypoint"]["x"], annotation["keypoint"]["y"])
-    elif "skeleton" in annotation:
-        main_annotation = dt.make_skeleton(name, annotation["skeleton"]["nodes"])
+    elif "ellipse" in annotation:
+        main_annotation = dt.make_ellipse(name, annotation["ellipse"])
+    elif "cuboid" in annotation:
+        main_annotation = dt.make_cuboid(name, annotation["cuboid"])
+    # TODO
+    # elif "skeleton" in annotation:
+    #     main_annotation = dt.make_skeleton(name, annotation["skeleton"]["nodes"])
```
andreaazzini, line 262: Fair enough, I guess we need to wait for the annotation class metadata to be included within exports

andreaazzini (approved):
Approved, the only comment I have is about a few lines that were commented in about handling skeletons, but it's not blocking

darwin/utils.py
almazan: There was also an issue when importing skeletons: fetching the annotation type id in `RemoteDataset` returns an error because it cannot find the `skeleton` type.

PR #120, opened 2020-11-03
add option to pull video frames intead of video file

almazan:
This PR adds an optional flag (`--video-frames`) to pull video frame images instead of video files.

```
darwin dataset pull v7/test --video-frames
```

Frames for each video are organised in different folders, with the name of each folder corresponding to the name of the video file (eg `myvideo1/0000001.jpg`).


darwin/dataset/download_manager.py
```diff
@@ -85,14 +88,25 @@ def download_all_images_from_annotations(
...
+    # for annotation_path in annotations_to_download_path:
+    #     download_image_from_annotation(
+    #         api_key,
+    #         api_url,
+    #         annotation_path,
+    #         images_path,
+    #         annotation_format,
+    #         use_folders,
+    #         video_frames,
+    #     )
```
simedw, line 104: remove?

darwin/dataset/download_manager.py
```diff
@@ -138,19 +154,31 @@ def download_image_from_json_annotation(
...
         Recreate folders
+    video_frames: bool
+        Pulls video frames images instead of video files
     """
     annotation = json.load(annotation_path.open())
 
-    # Make the image file name match the one of the JSON annotation
-    original_filename_suffix = Path(annotation["image"]["original_filename"]).suffix
+    filename = annotation_path.stem
+    if video_frames:
```
simedw, line 163: shouldn't we first check if what you are extracting is a video? 

imagine an export that has one video and one image in it.

darwin/dataset/download_manager.py
almazan: removed!

andreaazzini:
I read through it, I saw your comments @simedw, I'm afraid I wouldn't have much to add to your review 😊

darwin/dataset/download_manager.py
almazan: True, I didn't think about that situation. Just added a check.

simedw (approved):

PR #129, opened 2020-11-24
add support to export ellipses and bounding boxes

almazan:


darwin/utils.py
```diff
@@ -3,6 +3,7 @@
 from typing import TYPE_CHECKING, List, Optional, Union
 
 import numpy as np
+from skimage import draw
```
simedw, line 6: not too thrilled by using skimage for drawing...

simedw (changes requested):

simedw:
Can not include skimage as a dependency for now

PR #136, opened 2021-02-01
Only fetch needed files when importing annotations

simedw:


andreaazzini (approved):
This looks perfectly fine, so I'm approving.

Two things:
- remember to use the formatter
- if I remember correctly, we show one progress bar per annotation file when we import annotations. It's unrelated here but I think we should instead show a global progress bar for the whole import task

PR #139, opened 2021-03-06
[HOTFIX] Avoid stripping class names

andreaazzini:


andreaazzini:
@simedw This works, so you can merge it in any time

simedw (approved):

PR #145, opened 2021-03-25
Fixes for unicode decoding and missing file closing

tkupek:


darwin/dataset/remote_dataset.py
```diff
@@ -249,7 +253,7 @@ def pull(
 
         if release.latest:
             latest_dir = self.local_releases_path / "latest"
-            if latest_dir.exists():
+            if os.path.islink(latest_dir):
```
andreaazzini, line 256: I think we can leverage the `pathlib` equivalent
```suggestion
            if latest_dir.is_symlink():
```
We can avoid importing `os` that way

setup.py
```diff
@@ -1,7 +1,7 @@
 import setuptools
 
-with open("README.md", "r") as f:
-    long_description = f.read()
+with open("README.md", "rb") as f:
+    long_description = f.read().decode('utf-8')
```
andreaazzini, line 4: I would add a comment on top of this line, explaining why decoding is important here

darwin/dataset/download_manager.py
```diff
@@ -59,7 +59,9 @@ def download_all_images_from_annotations(
 
     annotations_to_download_path = []
     for annotation_path in annotations_path.glob(f"*.{annotation_format}"):
-        annotation = json.load(annotation_path.open())
+        with annotation_path.open() as file:
+            annotation = json.load(file)
+            file.close()
```
andreaazzini, line 64: As far as I'm aware, files are automatically closed when control leaves the `with` block, so these `file.close()` calls can be avoided

andreaazzini (changes requested):
Summary:
- Closing files is redundant at the end of `with` blocks
- We can avoid importing `os` and use its `pathlib` counterparts
- It would be nice to add a comment on why decoding is needed for future reference

darwin/dataset/remote_dataset.py
tkupek: Pathlib `is_symlink()` breaks the code on my Windows 10 machine. Can you verify if it works for you?

setup.py
tkupek: @florian-fischer-swarm can you do that?

darwin/dataset/download_manager.py
tkupek: Done.

darwin/dataset/remote_dataset.py
tkupek: Done.

darwin/dataset/remote_dataset.py
```diff
@@ -249,7 +250,7 @@ def pull(
 
         if release.latest:
             latest_dir = self.local_releases_path / "latest"
-            if latest_dir.exists():
+            if latest_dir.is_symlink():
```
andreaazzini, line 253: This change seems to be working on macOS, but I don't have access to a Windows machine. Do you happen to have further information on why it breaks on Windows? That would be super helpful 🙂 

Also, is there a reason why we didn't like `latest_dir.exists()`?

setup.py
andreaazzini: I would also use double quotes `"`, so the linter does not complain 🙂 
```suggestion
    long_description = f.read().decode("utf-8")
```

andreaazzini (changes requested):
Thank you for the quick updates!

The PR is looking pretty close to be mergeable, but I added a couple other comments.

darwin/dataset/remote_dataset.py
tkupek: `latest_dir.exists()` breaks on my Windows machine, `is_symlink` and the os alternative work.
I think `exists()` is meant for directories and files only, Windows handles symlinks a bit different as Unix systems, wherefore it does not work.

tkupek: Maybe you can add some unit-tests and schedule them on different machines via GitHub actions.

setup.py
```diff
@@ -1,7 +1,7 @@
 import setuptools
 
 with open("README.md", "rb") as f:
-    long_description = f.read().decode('utf-8')
+long_description = f.read().decode("utf-8")
```
andreaazzini, line 4: Indentation seems off here

andreaazzini (changes requested):
One more thing 🙂 

There seems to be an indentation problem. After that's resolved, I'll approve and merge this PR.

Thank you!

andreaazzini (approved):

PR #146, opened 2021-03-29
Helper to compute class and instance distributions

andreaazzini:


darwin/dataset/utils.py
```diff
@@ -768,3 +769,45 @@ def compute_max_density(annotations_dir: Path):
...
+                found_classes = []
+                for annotation in annotation_file.annotations:
+                    annotation_class = annotation.annotation_class.name
+                    # Count it in the class distribution, only if not found already
+                    if annotation_class not in found_classes:
+                        if annotation_class in class_distribution[partition]:
+                            class_distribution[partition][annotation_class] += 1
+                        else:
+                            class_distribution[partition][annotation_class] = 1
+                        found_classes.append(annotation_class)
```
simedw, line 806: This feels simpler
```suggestion
                for annotation in annotation_file.annotations:
                    annotation_class = annotation.annotation_class.name
                    # Count it in the class distribution, only once
                    class_distribution[partition][annotation_class] = 1
```

andreaazzini: It works only in case `annotation_class not in class_distribution[partition]`, becase otherwise we need to increment by one. But there's room for simplification, on it.

simedw (approved):

PR #150, opened 2021-03-31
[HOTFIX] handle different coco segmentation formats

simedw:


andreaazzini (approved):
This one makes sense to me: `dict` and `list` discrimination is safer, and the change in the comprehension on the list iteration looks more robust.

PR #156, opened 2021-05-13
[V7-491] Prepare video frame annotations for training

andreaazzini:


darwin/dataset/local_dataset.py
```diff
@@ -77,11 +77,12 @@ def __init__(
...
                 raise FileNotFoundError(
                     f"could not find a dataset partition. "
-                    f"Split the dataset using `split_dataset()` from `darwin.dataset.utils`"
+                    f"Split the dataset using `split_dataset()` from `darwin.dataset.split_manager`"
                 ) from None
         else:
             # If the partition is not specified, get all the annotations
-            stems = [e.stem for e in annotations_dir.glob("*.json")]
+            prefix = f"{annotations_dir}/"
+            stems = [str(e).replace(prefix, "").split(".json")[0] for e in annotations_dir.glob("**/*.json")]
```
simedw, line 85:
```suggestion
            stems = [e.stem for e in annotations_dir.glob("**/*.json")]
```

andreaazzini: This was the original code I had. Unfortunately, that won't work. The key is in the `"**"` wildcard in `glob`.
That would match something like `<annotations_dir>/parent/child.json`, but if we only do `e.stem` we lose the `parent/` bit.

simedw: Ah I missed that, how about `e.relative_to(annotations_dir)` ? it would give you `parent/child.json`, but since you want to drop the `.json`

```suggestion
            stems = [e.relative_to(annotations_dir).parent / e.stem for e in annotations_dir.glob("**/*.json")]
```

andreaazzini: Nice! I wasn't aware of that

simedw (approved):

PR #161, opened 2021-06-11
[V7-994] Parse remote path information at upload time

andreaazzini:


darwin/importer/importer.py
```diff
@@ -81,7 +81,8 @@ def import_annotations(
     print("Fetching remote file list...")
     remote_files = get_remote_files(dataset, filenames)
     for parsed_file in parsed_files:
-        if parsed_file.filename not in remote_files:
+        filename = f"{parsed_file.remote_path}/{parsed_file.filename}"
+        if filename not in remote_files:
```
simedw, line 85:
```suggestion
        if parsed_file.full_path not in remote_files:
```

darwin/item.py
```diff
@@ -13,6 +13,12 @@ class DatasetItem:
     dataset_slug: str
     seq: int
     current_workflow_id: Optional[int]
+    path: str
+
+    @property
+    def full_path(self) -> str:
+        path_items = filter(lambda item: item is not None, [self.path, self.filename])
+        return "/".join(path_items)
```
simedw, line 21:
```suggestion
        return "/".join(filter(None, [self.path, self.filename])) 
```

darwin/datatypes.py
```diff
@@ -77,6 +77,12 @@ class AnnotationFile:
     workview_url: Optional[str] = None
     seq: Optional[int] = None
     frame_urls: Optional[List[str]] = None
+    remote_path: Optional[str] = None
+
+    @property
+    def full_path(self) -> str:
+        path_items = filter(lambda item: item is not None, [self.remote_path, self.filename])
+        return "/".join(path_items)
```
simedw, line 85:
```suggestion
        return "/".join(filter(None, [self.path, self.filename]))
```

darwin/item.py
```diff
@@ -26,4 +31,5 @@ def parse_dataset_item(raw) -> DatasetItem:
         "n/a",
         raw["seq"],
         raw.get("current_workflow_id"),
+        raw["path"],
```
andreaazzini, line 34: @simedw is path always going to be there?

simedw: I think we always give it in our current api / export format

simedw (approved):

PR #162, opened 2021-06-24
[V7-1044] Make DatasetItem path not start with a root character

andreaazzini:


andreaazzini:
@simedw removed 3.6 pytest support because it was strangely complain about a `datatypes` import that is totally legit on other python versions for some reason. Now we can merge the PR if we want, but let's make sure to look into that later

begedin:
> @simedw removed 3.6 pytest support because it was strangely complain about a `datatypes` import that is totally legit on other python versions for some reason. Now we can merge the PR if we want, but let's make sure to look into that later

Only way to make sure is to have a ticket, so I recommend creating that, assigning it to someone and even putting it into a cycle. YOu can always move it into next cycle if other things are higher in prio, but it will never get tackled if it doesn't exist.

simedw:
The issue is circular dependencies:

utils import darwin.datatypes 
and
darwin.datatypes import utils

PR #164, opened 2021-06-30
[V7-1011] New upload api

simedw:


darwin/dataset/upload_manager.py
```diff
@@ -1,528 +1,610 @@
...
+        self.data["filename"] = args.get("filename") or self.local_path.name
+        self.data["type"] =  args.get("type") or self._infer_type()  
+
+    def _infer_type(self):
+        extension = Path(self.data["filename"]).suffix
+        if is_image_extension_allowed(extension):
+            return "image"
+        if is_video_extension_allowed(extension):
+            return "video"
+        raise UnsupportedFileType(self.data["filename"])
```
andreaazzini, line 27: Do we really need type checking? The backend is inferring the type already. (Still haven't read the whole PR while writing this, the type might be useful for something else)

darwin/dataset/upload_manager.py
```diff
@@ -1,528 +1,610 @@
...
+
+class UploadHandler:
+    def __init__(self, client, local_files, dataset_identifier):
+        self.client = client 
+        self.local_files = local_files
+        self.dataset_identifier = dataset_identifier
+
+        self.blocked_items, self.pending_items = request_upload(client, local_files, dataset_identifier)
+        print(self.blocked_items, self.pending_items)
```
andreaazzini, line 36: Is this line staying here, or a leftover from debug?

darwin/dataset/upload_manager.py
```diff
@@ -1,528 +1,610 @@
...
+            raise ValueError(f"Can not match {item['filename']} from payload with files to upload")
+        yield lambda: _upload_file(client, item["dataset_item_id"], file.local_path, team_slug=team_slug)
+
+def _upload_file(client: "Client", dataset_item_id: int, file_path: Path, team_slug: str):
+    sign_response = client.get(f"/dataset_items/{dataset_item_id}/sign_upload", team=team_slug)
+    signature = sign_response["signature"]
+    end_point = sign_response["postEndpoint"]
+
+    upload_response = requests.post("http:" + end_point, data=signature, files={"file": file_path.open("rb")})
+    upload_response.raise_for_status()
```
andreaazzini, line 78: If we raise here, does that get caught for us to print a more descriptive error message?

darwin/dataset/upload_manager.py
```diff
@@ -1,528 +1,610 @@
...
+
+    client.put(endpoint=f"/dataset_items/{dataset_item_id}/confirm_upload", payload={}, team=team_slug)
```
andreaazzini, line 80: Maybe payload should be empty by default?

darwin/dataset/upload_manager.py
```diff
@@ -1,528 +1,610 @@
...
+        if not file:
+            raise ValueError(f"Can not match {item['filename']} from payload with files to upload")
+        yield lambda: _upload_file(client, item["dataset_item_id"], file.local_path, team_slug=team_slug)
+
+def _upload_file(client: "Client", dataset_item_id: int, file_path: Path, team_slug: str):
+    sign_response = client.get(f"/dataset_items/{dataset_item_id}/sign_upload", team=team_slug)
+    signature = sign_response["signature"]
+    end_point = sign_response["postEndpoint"]
+
+    upload_response = requests.post("http:" + end_point, data=signature, files={"file": file_path.open("rb")})
```
andreaazzini, line 77:
```suggestion
    upload_response = requests.post(f"http:{end_point}" data=signature, files={"file": file_path.open("rb")})
```

darwin/dataset/upload_manager.py
```diff
@@ -1,528 +1,610 @@
...
+def _upload_files(client: "Client", files: List[LocalFile], items_pending_upload, team_slug: str):
+    file_lookup = { file.data["filename"]: file for file in files }
+    for item in items_pending_upload:
+        file = file_lookup.get(item['filename'])
+        if not file:
+            raise ValueError(f"Can not match {item['filename']} from payload with files to upload")
+        yield lambda: _upload_file(client, item["dataset_item_id"], file.local_path, team_slug=team_slug)
+
+def _upload_file(client: "Client", dataset_item_id: int, file_path: Path, team_slug: str):
+    sign_response = client.get(f"/dataset_items/{dataset_item_id}/sign_upload", team=team_slug)
```
andreaazzini, line 73: What happens if this request is not successful?

andreaazzini (changes requested):
Do we have an idea about how we want to specify tags here? It's still WIP, but I left a few comments anyway.

darwin/dataset/remote_dataset.py
```diff
@@ -69,7 +69,7 @@ def push(
         self,
         files_to_upload: List[str],
         blocking: bool = True,
-        multi_threaded: bool = True,
+        multi_threaded: bool = not True,
```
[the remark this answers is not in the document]
simedw: this was just for debugging :)

simedw:
is missing `path` support

darwin/dataset/upload_manager.py
```diff
@@ -1,528 +1,613 @@
...
+    
+#     print(data)
+
+#     if 'items' in data and data['items']:
+#         _upload_files(client, files, data['items'], team_slug)
+
+def _upload_files(client: "Client", files: List[LocalFile], items_pending_upload, team_slug: str):
+    file_lookup = { file.data["filename"]: file for file in files }
+    for item in items_pending_upload:
+        file = file_lookup.get(item['filename'])
```
simedw, line 67: This is broken if a user uploads two files with the same name but different dest folders, since we would always match to just one of them.

Solutions:
1) introduce the request_id we talked about (Cons: fair bit of work)
2) use the dest path + filename as the key (Cons: make sure to normalize that path, and are we even getting the path back in the response? if not we need to add it!)

darwin/dataset/remote_dataset.py
```diff
@@ -80,7 +80,7 @@ def push(
         self,
         files_to_upload: List[str],
         blocking: bool = True,
-        multi_threaded: bool = not True,
+        multi_threaded: bool = False,
```
simedw, line 83: let's make it true by default

darwin/dataset/remote_dataset.py
```diff
@@ -106,60 +117,39 @@ def push(
...
-                responses = [{k: str(v) for k, v in response.items()} for response in responses]
-                if resume:
-                    responses.extend(logged_responses)
-                with responses_path.open("w") as f:
-                    json.dump(responses, f)
-            return None, count
-        else:
-            return progress, count
+        if blocking and handler.pending_count:
+            import concurrent.futures
```
FrancescoSaverioZuppichini: import is not at the top of the file

darwin/dataset/remote_dataset.py
```diff
@@ -106,60 +117,39 @@ def push(
...
+            with concurrent.futures.ThreadPoolExecutor(max_workers=5) as executor:
+                future_to_progress = {executor.submit(f): f for f in handler.progress}
+                for future in concurrent.futures.as_completed(future_to_progress):
+                    try:
+                        future.result()
+                    except Exception as exc:
+                        print('%r generated an exception: %s' % (url, exc))
```
FrancescoSaverioZuppichini: There are a lot of `print` in the code, what if we use the [logging module](https://docs.python.org/3/howto/logging.html), so we can also persist `cli` logs for better debugging/support.

darwin/dataset/remote_dataset.py
```diff
@@ -106,60 +117,39 @@ def push(
...
-                if resume:
-                    responses.extend(logged_responses)
-                with responses_path.open("w") as f:
-                    json.dump(responses, f)
-            return None, count
-        else:
-            return progress, count
+        if blocking and handler.pending_count:
+            import concurrent.futures
+            with concurrent.futures.ThreadPoolExecutor(max_workers=5) as executor:
```
FrancescoSaverioZuppichini: `max_workers` is hardcoded, maybe this is intended behavior

FrancescoSaverioZuppichini (commented):
I hope the review was okay, let me know if this is the correct way to do it. 

There are missing types and return types are not always specified and sometimes a function can return different types based on the parameter passed

darwin/cli_functions.py
```diff
@@ -380,7 +399,89 @@ def upload_data(
...
+                        pending_item.filename,
+                        pending_item.path,
+                        error.stage.name,
+                        str(error.error),
+                    )
+                    break
+
+        console.print(error_table)
     except NotFound as e:
         _error(f"No dataset with name '{e.name}'")
```
myFree, line 486: Not directly on related to these changes, but perhaps we can add suggestions on error messages? For example this could be updated to be: 
`No dataset with name '{e.name}', use `darwin dataset remote` to see all available datasets`

Of course; this could introduce maintenance issues if commands change over time, so maybe we could centralize the error messages for slightly easier tracking

And maybe changes like this would be best tackled in their own ticket

andreaazzini (approved):
I think this is ready to be merged and used.

At the same time, I think we could clean the callback logic a bit.
When writing the docs, it would be nice to split into simple and advanced (progress + chunking etc), but the API we have currently makes it a little difficult to parse IMHO.

With that said, it's better to merge now and then improve it in a minor release than keeping adding stuff on top of this PR

PR #166, opened 2021-07-13
adding sorting to list_files from dataset

Fl4m3Ph03n1x:
What
---

Adding capability to sort items when listing them. 

How
---

The functions creating the requests were not taking into account sorting, so I added that. 

Questions
---

This is my first PR in this project and also the first time I have touched Python in over a decade. I expect this PR to be of poor quality. I also expect that I need to do additional work that I missed because I am not familiar with the project. 

So, given this context:

 - How can I improve this code?
 - Am I missing something?, aka, Did I forget to change something else for this to work?
 - Should I add tests for this functionality? (If so, how?)


darwin/cli.py
```diff
@@ -72,7 +72,7 @@ def run(args, parser):
         elif args.action == "export":
             f.export_dataset(args.dataset, args.include_url_token, args.annotation_class, args.name)
         elif args.action == "files":
-            f.list_files(args.dataset, args.status, args.path, args.only_filenames)
+            f.list_files(args.dataset, args.status, args.path, args.only_filenames, args.sort, args.direction)
```
andreaazzini, line 75: Being a CLI argument, I would call it `sort_by`. This should make the command a bit more readable
```
darwin dataset files v7/test-dataset complete --sort-by updated_at --direction desc
```

darwin/cli_functions.py
```diff
@@ -400,21 +400,28 @@ def dataset_import(dataset_slug, format, files, append):
...
-            filters["path"] = path
-        for file in dataset.fetch_remote_files(filters):
+            parameters["path"] = path
+        if sort: 
+            if sort not in ["inserted_at", "updated_at", "file_size", "filename", "priority"]:
+                _error(f"Invalid sort parameter '{sort}', available sort parameters: inserted_at, updated_at, file_size, filename, priority")
+            if direction not in ["asc", "desc"]:
+                _error(f"Invalid direction for sort '{direction}', available directions: asc, desc")
+            parameters["sort"] = sort
+            parameters["direction"] = direction
```
andreaazzini, line 423: I would say errors are brilliant, but direction is disregarded if sort isn't set. Not sure if that's what we want, but I think the following should work (listing files in descending order, sorting by "updated_at")
```
darwin dataset files v7/test-dataset complete --direction desc
```

andreaazzini (changes requested):
Pretty nice, I have a couple of change requests, but overall that would be a nice small feature to support.

On another topic, we have started enforcing tests on new PRs. I would really love to see a few test for `list_files` and a couple more for `fetch_remote_files`.

We're using `pytest`, and you should be able to find inspiration from the few tests already existing in `master` and on the upload API PR.

Finally, I have included a `.vscode/settings.json` file in the upload API PR, that should help formatting the files in the way we want. Please, let me know if that doesn't work or if you have feedback regarding those settings 🙂

darwin/cli_functions.py
Fl4m3Ph03n1x: The intention was to do just that: don't validate direction unless we have sort specified. However, now that I see your use case, I believe you are correct. I will update the code accordingly.

darwin/cli.py
Fl4m3Ph03n1x: Will update accordingly.

myFree (changes requested):

Fl4m3Ph03n1x:
> Pretty nice, I have a couple of change requests, but overall that would be a nice small feature to support.
> 
> On another topic, we have started enforcing tests on new PRs. I would really love to see a few test for `list_files` and a couple more for `fetch_remote_files`.
> 
> We're using `pytest`, and you should be able to find inspiration from the few tests already existing in `master` and on the upload API PR.
> 
> Finally, I have included a `.vscode/settings.json` file in the upload API PR, that should help formatting the files in the way we want. Please, let me know if that doesn't work or if you have feedback regarding those settings slightly_smiling_face

I really do think I will need a hand with them tests. I have checked some of the existing tests but I was unable to adapt them to my use case.

Yeah, I am not sure the formatter is working for my vscode setup. I keep getting errors about some dependencies missing and VSCode can't find them.

andreaazzini (changes requested):
Ok so your changes make perfect sense, but after looking at what @simedw suggested in the thread, I think
```
--sort-by updated_at:desc
```
makes it for a nicer CLI

Basically, we need an extra step to parse the `sort` string, but that should be simple enough. If no `:` is there, we could keep `asc` as default?

darwin/dataset/remote_dataset.py
```diff
@@ -317,25 +317,27 @@ def remove_remote(self):
         """Archives (soft-deletion) the remote dataset"""
         self.client.put(f"datasets/{self.dataset_id}/archive", payload={}, team=self.team)
 
-    def fetch_remote_files(self, filters: Optional[dict] = None):
+    def fetch_remote_files(self, parameters: Optional[dict] = None):
```
myFree, line 320: This could be a breaking change for any clients who currently invoke this function with "named parameters" (this might not be the correct Python naming standard)

We could add an additional parameter instead; so we end up with ```filters: Optional[dict] = None, sort: ....)

Fl4m3Ph03n1x: If we go by this rule, then we lock ourselves from ever changing the names of parameters ever again, but client's _might_ be using our functions with named parameters. 

The project is still in alpha, I don't see this as an issue. Clients should expect changes, that's what we defined.

andreaazzini: I tend to agree with @Fl4m3Ph03n1x. The argument is defined as positional, the project is not stable, I don't see argument name consistency as a major issue at this stage.

With that said, I like @myFree's suggestion about using an additional argument for sorting parameters, as the change in the `filters` name was made to make the code more readable/maintainable, and I reckon that suggestion is also pointing in that direction

simedw: What I would like is to do is not to change the positional parameters (usually the first 1-3 args), if there are parameters that are optional or might change we first add `*` in the function header like this:

```python
def example_function(filter, *, batch_size): 
```

This was batch_size can only be called as a named parameter. 

if batch_size is no longer needed, we can first keep accepting it but give out a warning, and then later drop it. 

Sometimes we do big changes and it doesn't make sense to be backwards compatible, and in those situations we can break function signatures, names etc. But it's always nice if we can avoid them when possible. 


In this specific function, I don't exactly see the benefit of merging the filters and sort into the same argument to the function, doesn't that make things less clear and less typed compared to 

```python
def get_remote_files(self, filters: Optional[dict] = None, sort: Optional[str] = None):
    pass

# .... 

dataset.get_remote_files(sort="updated:desc") 
```

thoughts?  (we could introduce a Sort class that can parse sort strings)

Fl4m3Ph03n1x:
> Ok so your changes make perfect sense, but after looking at what @simedw suggested in the thread, I think
> 
> ```
> --sort-by updated_at:desc
> ```
> 
> makes it for a nicer CLI
> 
> Basically, we need an extra step to parse the `sort` string, but that should be simple enough. If no `:` is there, we could keep `asc` as default?

Done.

darwin/cli_functions.py
```diff
@@ -561,7 +562,27 @@ def list_files(dataset_slug: str, statuses: str, path: str, only_filenames: bool
...
         if path:
             filters["path"] = path
-        for file in dataset.fetch_remote_files(filters):
+
+        if not sort_by:
+            sort_by = "updated_at:desc"
+
+        if not _has_valid_format(sort_by):
+            _error(
+                f"Invalid sort_by parameter '{sort_by}'. Correct format is 'attribute:direction', i.e., 'updated_at:desc'."
```
myFree, line 571: Potential typo - I don't think a `,` is needed after `i.e.`

darwin/options.py
```diff
@@ -80,7 +80,7 @@ def __init__(self):
             default="native",
             help="Frames per second for video split (recommended: 1), use 'native' to use the videos intrinsic fps.",
         )
-        parser_push.add_argument("--frames", action="store_true", help="Annotate a video as independent frames.")
+        parser_push.add_argument("--frames", action="store_true", help="Annotate a video as independent frames")
```
myFree, line 83: Is this change intentional? Seems to just be removing a `.`

darwin/path_utils.py
```diff
@@ -1,6 +1,7 @@
 from pathlib import PurePosixPath
 from typing import Optional
 
+
```
myFree, line 4: Redundant new line added?

myFree (approved):
Some minor comments regarding potential typos; but other than that looks good to me.
I'll "fearlessly" approve; but would recommend we get another approver before shipping

I added the steps for creating a new python release https://www.notion.so/v7labs/Darwin-Py-534ed10bbc3b4c7b8b577120baa68be6 just FYI

darwin/cli_functions.py
Fl4m3Ph03n1x: > Potential typo - I don't think a `,` is needed after `i.e.`
> But YAY and +1+1+1+1 for helpful error messages with an example fix suggestion; I think that can help end users along their journey nicely
> (Not for this PR; but might be worth centralizing error messages; so its easier to review and keep them up to date over time.)

Yes please, I would love a place for centralized error messages. Right now I have about the same error messages showing the same default, possible values for sort in 3 different places. If tomorrow we add a new sorting attribute, I bet I (or whoever is working here) will forget to update all of 3 of them.

simedw:
Made a small push regarding what I mentioned on slack earlier, let me know what you think. 
Not really sold on the name item_sorter to be honest.

darwin/path_utils.py
Fl4m3Ph03n1x: Formatter doing formatter things :D

tests/darwin/dataset/remote_dataset_test.py
```diff
@@ -63,6 +64,237 @@ def create_annotation_file(
         json.dump(annotation_content, f)
 
 
+@pytest.fixture()
+def files_content() -> dict:
+    return {
+        "items": [
```
andreaazzini, line 70: We seem to be leaking some valid signed URLs here, are we ok with it?

On a separate note, should we keep these large payloads in actual json files, so we only open and decode them here? That would make this code cleaner, and a centralized place for example payloads to reuse in tests might be beneficial.

tests/darwin/item_sorter_test.py
```diff
@@ -0,0 +1,72 @@
+import pytest
+
+from darwin.item_sorter import ItemSorter, SortDirection
+
+
+def describe_item_sorter():
+    def describe_parse():
+        def test_works_when_sort_is_complete_and_valid():
```
andreaazzini, line 8: I would suggest getting rid of these `test_` prefixes, now we can 🙂 
```suggestion
        def works_when_sort_is_complete_and_valid():
```

andreaazzini (changes requested):

darwin/item.py
```diff
@@ -22,7 +22,7 @@ def full_path(self) -> str:
         return construct_full_path(self.path, self.filename)
 
 
-def parse_dataset_item(raw) -> DatasetItem:
+def parse_dataset_item(raw: Dict[str, Any]) -> DatasetItem:
```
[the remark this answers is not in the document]
Fl4m3Ph03n1x: Why dont we do that? What are the blockers?

darwin/cli.py
```diff
@@ -1,5 +1,9 @@
+__all__ = ["main"]
```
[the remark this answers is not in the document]
Fl4m3Ph03n1x: They are complementary. Using the `_private` will not keep people from using our private functions if they do `import *`. You might say "_this is Python, every one is responsible_". But would you really apply the mathematical principle of generalization to big carbon bags such as Humans? I think that would be unwise.

On the other hand, `__all__` will only export the functions listed when you do `import *`. There is more to it though:
https://stackoverflow.com/a/35710527/1337392

I personally believe this is a good addition. We could talk about it.

darwin/options.py
Fl4m3Ph03n1x: Fixed!

tests/darwin/dataset/remote_dataset_test.py
Fl4m3Ph03n1x: This file is the decoded (into Python format) json, not a _real_ json, file. 
We could have a folder for that, I am OK with either approach.

tests/darwin/item_sorter_test.py
Fl4m3Ph03n1x: Done

andreaazzini (approved):

PR #167, opened 2021-07-14
Pass RemoteDataset object to initialize UploadHandler objects

andreaazzini:


darwin/dataset/remote_dataset.py
```diff
@@ -128,11 +128,15 @@ def push(
 
         local_files = []
         for file in files_to_upload:
-            local_files.append(LocalFile(file, fps=fps, as_frames=as_frames, path=path))
+            local_files.append(LocalFile(Path(file), fps=fps, as_frames=as_frames, path=path))
```
simedw, line 131: aren't they already Paths there, looking at the file signature for `find_files`

simedw (approved):

darwin/dataset/remote_dataset.py
```diff
@@ -78,61 +83,79 @@ def __init__(
 
     def push(
         self,
-        files_to_upload: List[str],
+        files_to_upload: Optional[List[Union[str, Path, LocalFile]]],
```
andreaazzini, line 86: This is debatable, but I would force all arguments but `files_to_upload` to be named.

darwin/dataset/remote_dataset.py
```diff
@@ -78,61 +83,79 @@ def __init__(
...
+        if any(isinstance(item, LocalFile) for item in files_to_upload) and generic_parameters_specified:
+            raise ValueError("Cannot specify a path when uploading a LocalFile object.")
 
-        if not files_to_upload:
-            raise ValueError("No files to upload, check your path, exclusion filters and resume flag")
+        uploading_files: List[LocalFile] = []
+        for item in files_to_upload:
+            if isinstance(item, LocalFile):
+                uploading_files.append(item)
+                continue
```
simedw, line 139:
```suggestion
        uploading_files = [item for files_to_upload if isinstance(item, LocalFile)] 
        search_files = [item for files_to_upload if not isinstance(item, LocalFile)]
        
        if uploading_files and generic_parameters_specified:
            raise ValueError("Cannot specify a path when uploading a LocalFile object.")
        
        for item in find_files(files=search_files, recursive=True, files_to_exclude=files_to_exclude):
            uploading_files.append(LocalFile(item, fps=fps, as_frames=as_frames, path=path))
        
        
        for item in files_to_upload:
            if isinstance(item, LocalFile):
                uploading_files.append(item)
                continue
```

darwin/dataset/remote_dataset.py
```diff
@@ -78,61 +83,80 @@ def __init__(
...
         if files_to_exclude is None:
             files_to_exclude = []
+
         if files_to_upload is None:
             raise ValueError("No files or directory specified.")
 
-        files_to_upload = find_files(files=files_to_upload, recursive=True, files_to_exclude=files_to_exclude)
+        generic_parameters_specified = path is not None or fps != 0 or as_frames
+        if any(isinstance(item, LocalFile) for item in files_to_upload) and generic_parameters_specified:
+            raise ValueError("Cannot specify a path when uploading a LocalFile object.")
```
simedw, line 133: you don't need this anymore

darwin/dataset/remote_dataset.py
Fl4m3Ph03n1x: Given that push is a public function part of darwin's public API, wouldn't such a change possibly break clients that are already invoking it with positional parameters only?

darwin/dataset/identifier.py
```diff
@@ -8,8 +9,11 @@ def __init__(self, dataset_slug: str, team_slug: Optional[str] = None, version:
         self.version = version
 
     @classmethod
-    def parse(cls, identifier: str):
-        team_slug, dataset_slug, version = parse(identifier)
+    def parse(cls, identifier: Union[str, "DatasetIdentifier"]) -> "DatasetIdentifier":
```
Fl4m3Ph03n1x, line 12: Can't we use:

```
from __future__ import annotations

def parse(cls, identifier: Union[str, "DatasetIdentifier"]) -> DatasetIdentifier:
```

instead?

darwin/dataset/remote_dataset.py
andreaazzini: Yes, that's why I wanted to bring this conversation up. We are changing the upload API entirely, and this belongs to a `0.6` release, so I'd be fine with that, but let's hear what @simedw or @myFree think

darwin/dataset/identifier.py
andreaazzini: I think it's not going to work for Python 3.6, but let me try that

andreaazzini: As I anticipated 😢 
https://github.com/v7labs/darwin-py/pull/167/checks?check_run_id=3086770021

darwin/dataset/identifier.py
```diff
@@ -22,18 +26,15 @@ def __str__(self):
...
-        else:
-            dataset = slug
-            version = None
-        return team, dataset, version
-    except ValueError as e:
-        raise type(e)(f"Invalid dataset identifier {slug}")
+def _parse(slug: str) -> Tuple[str, str, Optional[str]]:
+    version: Optional[str] = None
+
+    if not re.fullmatch(r"[\w\-]+/[\w\-]+(:[\w\.]+)?", slug):
```
Fl4m3Ph03n1x, line 32: For me, personally, this is hard to understand. Perhaps making this _explicit_ could be an option?

```
if not _is_dataset_valid(slug):
  raise ValueError(f"Invalid dataset identifier {slug}")
```
```

File: darwin/dataset/remote_dataset.py [hunk 7]
```diff
@@ -78,61 +83,74 @@ def __init__(
...
 
-        if not files_to_upload:
-            raise ValueError("No files to upload, check your path, exclusion filters and resume flag")
+        uploading_files = [item for item in files_to_upload if isinstance(item, LocalFile)]
```
Fl4m3Ph03n1x, line 131: would this be correct?
```
uploading_files: LocalFile = [item for item in files_to_upload if isinstance(item, LocalFile)]
```

darwin/utils.py
```diff
@@ -84,41 +84,39 @@ def prompt(msg: str, default: Optional[str] = None) -> str:
 
 
 def find_files(
-    files: List[Union[str, Path]] = [], recursive: bool = True, files_to_exclude: List[Union[str, Path]] = []
+    files: List[Union[str, Path]], *, files_to_exclude: List[Union[str, Path]] = [], recursive: bool = True
```
Fl4m3Ph03n1x, line 87: Wont this break backwards compatibility with previous clients now that we are forcing some of the arguments to be named?

darwin/dataset/remote_dataset.py
andreaazzini: I think so! What's your opinion about explicit type annotations on variables? Your linter should be able to figure it out anyway, and I have a feeling that being stricter would cause a lot of code to be unnecessarily split into multiple lines, thus making it a bit less readable

tests/darwin/dataset/identifier_test.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+from darwin.dataset.identifier import DatasetIdentifier
+
+
+class TestDatasetIdentifier:
```
Fl4m3Ph03n1x, line 5: Instead of defining a new class I have installed ```pytest-describe``` and I am using describe blocks in my tests. Perhaps we should have a discussion to see which convention we ought to adopt.

darwin/utils.py
andreaazzini: Yes, but I'm like 99.9% positive this function is not called explicitly by anybody, and even if it is, it's a helper function, not really meant to be front facing. Let's see if @simedw and @myFree agree

tests/darwin/dataset/identifier_test.py
andreaazzini: Interesting, let me have a look

darwin/dataset/remote_dataset.py
Fl4m3Ph03n1x: I personally like static typing. For me it was not obvious that `uploading_files` was a list of `LocalFile`. I had to spend time analyzing the code and then try to make an educated guess of what the type was. This could have been avoided if the type was explicit. 

I dont mind the extra boiler plate, which imo, is actually useful documentation and not really useless boiler plate. The linter is also not that smart most times, so it needs a hand. 

I personally prefer having things with types. If this means more lines, that's ok, that's why we use code formatters and good practices to make the code readable.

tests/darwin/dataset/identifier_test.py
andreaazzini: Adopted because:
- classes is imho a terrible means of structuring/organizing tests
- `pytest-describe` allows you to get rid of the `test_` naming nonsense

darwin/dataset/remote_dataset.py
simedw: If we are changing it, now is the time.
Let's do it.

darwin/dataset/identifier.py
```diff
@@ -22,18 +26,19 @@ def __str__(self):
...
+    team, dataset = slug.split("/")
+
+    if ":" in dataset:
+        dataset, version = dataset.split(":")
+
+    return team, dataset, version
+
+
+def _is_slug_valid(slug: str) -> bool:
+    return re.fullmatch(r"[\w\-]+/[\w\-]+(:[\w\.]+)?", slug) is not None
```
simedw, line 44: I think this blocks us from doing just "dataset name" which we could before! 

```suggestion
    return re.fullmatch(r"([\w\-]+/)?[\w\-]+(:[\w\.]+)?", slug) is not None
```

PR #173, opened 2021-07-22
New bounding box format for coco

simedw:
- preserve folder structure on upload
- Fix tests
- Coco bounding box new format


andreaazzini (approved):
Merged master into it, then formatted it. Looks good otherwise

PR #178, opened 2021-07-30
Check darwin-py version

simedw:
- send darwin-py version as header
- fetch latest version

### DIFF
--- a/darwin/cli_functions.py
+++ b/darwin/cli_functions.py
@@ -144,6 +144,7 @@ def create_dataset(dataset_slug: str):
         print(
             f"Dataset '{dataset.name}' ({dataset.team}/{dataset.slug}) has been created.\nAccess at {dataset.remote_path}"
         )
+        print_new_version_info(client)
     except NameTaken:
         _error(f"Dataset name '{identifier.dataset_slug}' is already taken.")
     except ValidationError:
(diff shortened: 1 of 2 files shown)


darwin/client.py
```diff
@@ -513,5 +521,24 @@ def _decode_response(response, debug: bool = False):
...
+
+            current_version = parse_version(__version__)
+            latest_version = parse_version(server_latest_version)
+            if current_version >= latest_version:
+                return
+            self._newer_version = latest_version
+        except:
+            pass
+
+    def newer_darwin_version(self):
```
andreaazzini, line 540: I'd make this a `property`

darwin/cli_functions.py
```diff
@@ -719,3 +729,12 @@ def _load_client(
...
 
+
+def print_new_version_info(client):
+    if client and not client.newer_darwin_version():
+        return
+    (a, b, c) = client.newer_darwin_version()
+    print(
+        f"A newer version of darwin-py is available [{a}.{b}.{c}]\ninstall with: pip install darwin-py=={a}.{b}.{c}",
+        file=sys.stderr,
+    )
```
andreaazzini, line 740: Why don't we use `rich` to print a `warning` message here?
Also, why are we writing into `sys.stderr`?

PR #179, opened 2021-08-04
support for importing data with team wide classes

simedw:


darwin/dataset/remote_dataset.py
```diff
@@ -374,15 +374,42 @@ def create_annotation_class(self, name: str, type: str):
...
+
+        datasets = match[0]["datasets"]
+        # check that we are not already part of the dataset
+        for dataset in datasets:
+            if dataset["id"] == self.dataset_id:
+                return
+        datasets.append({"id": self.dataset_id})
+        return self.client.put(f"/annotation_classes/{match[0]['id']}", {"datasets": datasets, "id": match[0]["id"]})
+
+    def fetch_remote_classes(self, all=False):
```
andreaazzini, line 403: What do you think of `team_wide` instead of `all`? Otherwise, that might read a bit confusing
```suggestion
    def fetch_remote_classes(self, team_wide=False):
```

darwin/importer/importer.py
```diff
@@ -65,14 +75,39 @@ def get_remote_files(dataset, filenames):
     return remote_files
 
 
+def _resolve_annotation_classes(annotation_class: List[dt.AnnotationClass], classes_in_dataset, classes_in_team):
```
andreaazzini, line 78: Can we include the missing types here?

darwin/dataset/remote_dataset.py
```diff
@@ -374,15 +374,42 @@ def create_annotation_class(self, name: str, type: str):
                 "name": name,
                 "metadata": {"_color": "auto"},
                 "annotation_type_ids": [type_id],
+                "datasets": [{"id": self.dataset_id}],
```
andreaazzini, line 377: Do we still need `"dataset_id"` at line 373?

darwin/importer/importer.py
andreaazzini:
```suggestion
def _resolve_annotation_classes(annotation_classes: List[dt.AnnotationClass], classes_in_dataset, classes_in_team):
```

darwin/importer/importer.py
```diff
@@ -65,14 +75,39 @@ def get_remote_files(dataset, filenames):
...
+        if cls.name in [missing_class.name for missing_class in local_classes_not_in_dataset]:
+            continue
+        if cls.name in [missing_class.name for missing_class in local_classes_not_in_team]:
+            continue
+
+        if annotation_type in classes_in_team and cls.name in classes_in_team[annotation_type]:
+            local_classes_not_in_dataset.add(cls)
+        else:
+            local_classes_not_in_team.add(cls)
+    return (local_classes_not_in_dataset, local_classes_not_in_team)
```
andreaazzini, line 98:
```suggestion
    return local_classes_not_in_dataset, local_classes_not_in_team
```

andreaazzini (changes requested):
Minor-ish comments, looks good otherwise. I'd really love not to leave this untested though 🤔

darwin/dataset/remote_dataset.py
simedw: strangely enough both are needed. @begedin maybe we could change that?

andreaazzini (approved):

PR #180, opened 2021-08-04
adding tests and typing to utils_py

Fl4m3Ph03n1x:


tests/darwin/utils_test.py
```diff
@@ -1,6 +1,41 @@
...
+    def it_returns_true_for_allowed_extensions():
+        assert is_video_extension_allowed(".mp4")
+
+    def it_returns_false_for_unknown_extensions():
+        assert not is_video_extension_allowed(".not_video")
+
+
+def describe_urljoin():
+    def it_returns_an_url():
+        assert urljoin("api", "teams") == "api/teams"
```
andreaazzini, line 38: I would add an assertion to see if `strip` works as expected

andreaazzini (changes requested):

tests/darwin/utils_test.py
Fl4m3Ph03n1x: done !

andreaazzini (approved):

PR #182, opened 2021-08-10
One fix, and one variable rename, and one unfix'd issue; tbc.

myFree:


Fl4m3Ph03n1x (changes requested):
Do we have a test that covers this change to the if statement?

darwin/dataset/remote_dataset.py
```diff
@@ -387,7 +387,10 @@ def add_annotation_class(self, annotation_class):
             cls
             for cls in all_classes
             if cls["name"] == annotation_class.name
```
Fl4m3Ph03n1x, line 389: Might be a good idea to put this if condition inside a function that makes it clear that it does (with the correct name) to go along the lines of the Python Good code practices.

darwin/importer/importer.py
```diff
@@ -75,26 +75,26 @@ def get_remote_files(dataset, filenames):
     return remote_files
 
 
-def _resolve_annotation_classes(annotation_classes: List[dt.AnnotationClass], classes_in_dataset, classes_in_team):
+def _resolve_annotation_classes(local_annotation_classes: List[dt.AnnotationClass], classes_in_dataset, classes_in_team):
     local_classes_not_in_dataset = set()
```
Fl4m3Ph03n1x, line 79:
```
local_classes_not_in_dataset: Set[dt.AnnotationClass] = set()
```

darwin/importer/importer.py
```diff
@@ -75,26 +75,26 @@ def get_remote_files(dataset, filenames):
...
     local_classes_not_in_team = set()
```
Fl4m3Ph03n1x, line 80: Same as above.

Fl4m3Ph03n1x (changes requested):

darwin/dataset/remote_dataset.py
myFree: Could you share an example code snippet you feel would make this line clearer?

To me; putting it in a function would require going into that function to see what it does; which is an extra step when trying to understand the code.
That line of code only does one very obvious thing - it compares two values.

Fl4m3Ph03n1x: I see you are comparing values, but I dont understand what is the semantic meaning you are trying to convey. 
Are you trying to check the class is the same or is compatible?

myFree: Discussed on a productive call; understand Pedro's input - leaving as is for now

simedw (approved):

PR #187, opened 2021-08-20
changed max_workers

FrancescoSaverioZuppichini:
Changed `max_workers` in upload manager to be `multiprocessing.cpu_counts()`

### DIFF
--- a/darwin/dataset/upload_manager.py
+++ b/darwin/dataset/upload_manager.py
@@ -10,6 +10,8 @@ from darwin.path_utils import construct_full_path
 from darwin.utils import chunk
 from requests_toolbelt import MultipartEncoder, MultipartEncoderMonitor
 
+MIN_WORKERS = 5
+
 if TYPE_CHECKING:
     from darwin.client import Client
     from darwin.dataset import RemoteDataset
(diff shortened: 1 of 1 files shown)


Fl4m3Ph03n1x:
May I suggest we use instead `len(os.sched_getaffinity(0))` to get the number of usable CPUs, since `cpu_count` does not return that?

https://docs.python.org/3/library/multiprocessing.html#multiprocessing.cpu_count

Also, since `cpu_count` can throw an exception, I believe it would be better to account for that and to have a default number of CPUs (5) should the error occur.

FrancescoSaverioZuppichini:
Thank you for your comment @simedw , I was not aware that `multiprocessing.cpu_count()` doesn't return the global number of cpus. I'll change it

FrancescoSaverioZuppichini:
btw, doc from `ThreadPoolExecutor`. Maybe we can just set it to None since the machine will at least have 1 core resulting in `num_workers=5`.

> Changed in version 3.5: If max_workers is None or not given, it will default to the number of processors on the machine, multiplied by 5, assuming that ThreadPoolExecutor is often used to overlap I/O instead of CPU work and the number of workers should be higher than the number of workers for ProcessPoolExecutor.

simedw (approved):

PR #188, opened 2021-08-23
create annotation classes with sub annotations

simedw:
- support for creating annotations with sub types
- better error message

### DIFF
--- a/darwin/dataset/remote_dataset.py
+++ b/darwin/dataset/remote_dataset.py
@@ -79,6 +79,7 @@ class RemoteDataset:
         self.image_count = image_count
         self.progress = progress
         self.client = client
+        self.annotation_types = None
 
     def push(
         self,
@@ -119,7 +120,7 @@ class RemoteDataset:
             Optional callback, called every time the progress of an uploading files is reported.
         file_upload_callback: Optional[FileUploadCallback]
             Optional callback, called every time a file chunk is uploaded.
-        
+
         Returns
         -------
         handler : UploadHandler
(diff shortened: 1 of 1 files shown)


darwin/dataset/remote_dataset.py
```diff
@@ -366,35 +367,30 @@ def restore_archived(self, items) -> None:
...
-        name : str
-            The name of the annotation class.
-        type : str
-            The type of the annotation class.
+    def create_annotation_class(self, name: str, type: str, subtypes: List[str] = []):
+        type_ids = []
+        for annotation_type in [type] + subtypes:
+            type_id = self.fetch_annotation_type_id_for_name(annotation_type)
+            if not type_id:
+                list_of_annotation_types = ", ".join([type["name"] for type in self.annotation_types])
```
FrancescoSaverioZuppichini, line 381: I have a doubt about this part. So, we are searching for `annotation_type` that may be `type` or something inside `subtypes`. 

Then, if we fail (`fetch_annotation_type_id_for_name` returns `None`) we are going to raise a ValueError saying that specific `annotation_type` was `Unknown` and we add `valid values: {list_of_annotation_types}"`. But if the only `valid values` are the ones inside `list_of_annotation_types` (computed from `self.annotation_types` could we just use the intersection between `subtypes` and `self.annotation_types`?

FrancescoSaverioZuppichini (commented):
Great work! Just a little comment

darwin/dataset/remote_dataset.py
simedw: But we also want to get all the type ids?

FrancescoSaverioZuppichini: After yesterday discussion, LGTM

FrancescoSaverioZuppichini (approved):
👍

PR #195, opened 2021-09-03
[HOTFIX] Avoid unnecessary stem stripping

andreaazzini:


simedw (commented):
Not sure if you also want \r (I think I used rstrip("\n\r") in a couple of places)

Fl4m3Ph03n1x (approved):

PR #200, opened 2021-09-10
Upload files with PUT instead of POST requests

rslota:
We've added `upload_url` field to `/sign_upload` endpoint on backend which contains PUT URL for data upload. This is now preferred method of uploading which is in line with what we're on our UI. Old fields for POST uploads are gonna stay for a while for backwards compatibility.


andreaazzini (changes requested):
The PR looks good to me.

If something has changed on the backend that has made this PR possible, I think we should write a couple of lines in the PR's description.

Additionally, let's make sure there's not gonna be outdated documentation on readme.com!

rslota:
> The PR looks good to me.
> 
> If something has changed on the backend that has made this PR possible, I think we should write a couple of lines in the PR's description.
> 
> Additionally, let's make sure there's not gonna be outdated documentation on readme.com!

@andreaazzini I'm gonna update docs as soon as backend changes hit production, because those didn't just yet. That's also why I haven't finished this PR (fixing tests). I'll update both PR and readme.com soon ™️ .

darwin/dataset/upload_manager.py
```diff
@@ -44,6 +44,35 @@ def full_path(self):
...
+    """
+
+    def __init__(self, io: BinaryIO, file_size: int, callback):
+        #: Instance of the :class:`BufferedReader` being monitored
+        self.io = io
+
+        #: Optionally function to call after a read
+        self.callback = callback
+
+        #: Number of bytes already read from the :class:`BufferedReader`
```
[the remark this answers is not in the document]
rslota: I'm not really a python dev, so I may have some weird biases as i developed in this language so little. 
Here, this is just a habbit of mine that may not make sense here - `":` comments are special comments that are picked up by `Sphinx` when generating docs.

rslota:
@andreaazzini PR description updated, documentation (recipe https://docs.v7labs.com/recipes/upload-visual-data-via-http) updated. URL in tests has been expanded to make sure no part of it gets lost between requests.

darwin/dataset/upload_manager.py
```diff
@@ -44,6 +44,35 @@ def full_path(self):
...
 
+class FileMonitor(object):
+    """
+    An object used to monitor the progress of a :class:`BufferedReader`.
+
+    To use this monitor, you construct your :class:`BufferedReader` as you
+    normally would, then construct this object with it as argument.
+    """
+
+    def __init__(self, io: BinaryIO, file_size: int, callback):
```
andreaazzini, line 55: any chance we could add a type annotation to `callback`?

andreaazzini (changes requested):
Very easily approvable once we make sure we add all type annotations. Otherwise, LGTM!

darwin/dataset/upload_manager.py
rslota: That weirdly required `__future__` 🤷 
https://github.com/v7labs/darwin-py/pull/200/commits/b362f33d402e24131fef5c7d54640a92ac64909e

rslota: Oh, this works only on python 3.7+... @andreaazzini Do you have other ideas?

darwin/dataset/upload_manager.py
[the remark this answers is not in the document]
Fl4m3Ph03n1x: Here we prefer the rst documentation style instead of the `#:`. Please see our Python guidelines for an example:

 - https://www.notion.so/v7labs/Python-Code-Style-351a7831c2074e78bf4e21ec13777542

darwin/dataset/upload_manager.py
```diff
@@ -44,6 +43,35 @@ def full_path(self):
...
+        #: Optionally function to call after a read
+        self.callback = callback
+
+        #: Number of bytes already read from the :class:`BufferedReader`
+        #: instance
+        self.bytes_read = 0
+
+        self.len = file_size
+
+    def read(self, size=-1):
```
Fl4m3Ph03n1x, line 67: Please add typing:

```
def read(self, size: int = -1) -> Any
```

As well as function documentation.

Fl4m3Ph03n1x (changes requested):

darwin/dataset/upload_manager.py
```diff
@@ -44,6 +53,44 @@ def full_path(self):
...
+
+        self.bytes_read = 0
+
+        self.len = file_size
+
+    def read(self, size: int = -1) -> Any:
+        """
+        Reads given amount of bytes from configured IO and calls the configured callback for each
+        block read. The callback is passed a reference this object that can be used to get current
+        self.bytes_read.
```
Fl4m3Ph03n1x, line 80: Missing documentation parameter `size`.

darwin/dataset/upload_manager.py
```diff
@@ -189,34 +236,29 @@ def _do_upload_file(
         except Exception as e:
             raise UploadRequestError(file_path=file_path, stage=UploadStage.REQUEST_SIGNATURE, error=e)
 
-        signature = sign_response["signature"]
-        end_point = sign_response["postEndpoint"]
+        upload_url = sign_response["upload_url"]
```
Fl4m3Ph03n1x, line 239: `upload_url: str = sign_response["upload_url"]`

Fl4m3Ph03n1x (changes requested):

Fl4m3Ph03n1x (approved):

Fl4m3Ph03n1x:
I made a commit with the documentation changes I thought were necessary just so we can approve this. 
Other than that, LGTM.

PR #201, opened 2021-09-14
V7 1539 - Multi Labels Dataset

FrancescoSaverioZuppichini:
Since the APIs are not test-friendly, the following code is not tested. It should work, please have a look. `ClassificationDataset` now returns a Tensor of size 1 for single-label classification and a Tensor of size `[C,1]` for multi-label classification where `C` is the number of classes


darwin/torch/dataset.py
```diff
@@ -108,17 +118,31 @@ def get_target(self, index: int):
...
+            # one hot encode all the targets
+            for tag in tags:
+                idx = self.classes.index(tag)
+                target[idx] = 1
+        else:
+            target = self.classes.index(tags[0])
+
         return target
 
+    def check_if_multi_label(self):
```
Fl4m3Ph03n1x, line 133: def check_if_multi_label(self) -> None:

darwin/torch/dataset.py
```diff
@@ -195,7 +219,9 @@ def get_target(self, index: int):
             # Extract the sequences of coordinates from the polygon annotation
             annotation_type = "polygon" if "polygon" in annotation else "complex_polygon"
```
Fl4m3Ph03n1x, line 220: `annotation_type: str`

Fl4m3Ph03n1x (commented):
Some documentation could be added. Other than that, LGTM

darwin/torch/dataset.py
FrancescoSaverioZuppichini: thanks, done!

darwin/torch/dataset.py
FrancescoSaverioZuppichini: thanks, done!

PR #206, opened 2021-09-21
[V7-1647] Make sure to include subs when converting bounding boxes to coco

andreaazzini:


darwin/datatypes.py
```diff
@@ -86,43 +87,44 @@ def full_path(self) -> str:
...
 
-def make_keypoint(class_name, x, y):
-    return Annotation(AnnotationClass(class_name, "keypoint"), {"x": x, "y": y})
+def make_keypoint(class_name: str, x: float, y: float, subs: List[SubAnnotation] = []):
+    return Annotation(AnnotationClass(class_name, "keypoint"), {"x": x, "y": y}, subs)
 
 
-def make_line(class_name, path):
-    return Annotation(AnnotationClass(class_name, "line"), {"path": path})
+def make_line(class_name: str, path: List[Point], subs: List[SubAnnotation] = []):
```
FrancescoSaverioZuppichini, line 114: Please remove the default parameter `=[]`. Default list (/dict/tuple) in parameters are evaluate ad function definition time, so they points always to the same instance creating weird side effects:

```python
>>> def foo(a=[]):
...     a.append('a')
...     return a
... 
>>> foo()
['a']
>>> foo()
['a', 'a']
>>> foo()
['a', 'a', 'a']
```
We can make them `Optional` and `=None`

FrancescoSaverioZuppichini (changes requested):

darwin/datatypes.py
andreaazzini: Great point, will change that

FrancescoSaverioZuppichini (approved):
LGTM! 🚀

PR #209, opened 2021-09-22
[ERRATA] [V7-1539] Change default behaviour in ClassificationDataset

FrancescoSaverioZuppichini:
- Better naming, before `target` was being overwritten.
- `assert` if no `tags`
- default behaviour is to the first tag, if `multi_label` one-hot them


darwin/torch/dataset.py
```diff
@@ -115,18 +115,20 @@ def get_target(self, index: int):
         Returns the classification target
         """
 
-        target = self.parse_json(index)
-        annotations = target.pop("annotations")
+        data = self.parse_json(index)
+        annotations = data.pop("annotations")
         tags = [a["name"] for a in annotations if "tag" in a]
 
+        assert len(tags) >= 1, f"No tags were found for index={index}"
```
Fl4m3Ph03n1x, line 122: I am confused. If `tags` has content, does it not mean we actually did find tags for the index?
Also, I can't really remember regarding the use of asserts in functions, we did mention this topic last week. 
@andreaazzini  sorry to bug you, but can you pop in and remind us of what the conclusion was? Were we OK with it?

Fl4m3Ph03n1x (changes requested):
I have a small question. Other than that, would it be feasible to add a test?

darwin/torch/dataset.py
FrancescoSaverioZuppichini: assert will raise an exception if `len(tags) >=1` is `False`, so no tag were found

Fl4m3Ph03n1x: Good point !

Fl4m3Ph03n1x (approved):

PR #210, opened 2021-09-22
Bounding Box Dataset

FrancescoSaverioZuppichini:
With this PR we add a new bind for PyTorch for object detection.

`from darwin.torch.dataset import BoundingBoxDetectionDataset`


darwin/torch/dataset.py
```diff
@@ -343,3 +325,86 @@ def measure_weights(self, **kwargs):
...
+class BoundingBoxDetectionDataset(LocalDataset):
+    def __init__(self, transform: Optional[List] = None, **kwargs):
+        """See `LocalDataset` class for documentation"""
+        super().__init__(annotation_type="bounding_box", **kwargs)
+
+        self.transform = transform
+        if self.transform is not None and isinstance(self.transform, list):
+            self.transform = Compose(self.transform)
+
+    def __getitem__(self, index: int):
```
Fl4m3Ph03n1x, line 339: Nice documentation, but could we also add the Typing for the Return value?

darwin/torch/dataset.py
```diff
@@ -343,3 +325,86 @@ def measure_weights(self, **kwargs):
...
+            image_path: str
+                The path to the image on the file system
+            labels : tensor(n)
+                The class label of each one of the instances
+            boxes : tensor(n, 4)
+                Coordinates of the bounding box enclosing the instances as [x, y, x, y]
+            area : float
+                Area in pixels of each one of the instances
+        """
+        img = self.get_image(index)
```
Fl4m3Ph03n1x, line 355: What is the type of `img`?

darwin/torch/dataset.py
```diff
@@ -343,3 +325,86 @@ def measure_weights(self, **kwargs):
             target = self.get_target(i)
             labels.extend([a["category_id"] for a in target["annotations"]])
         return self._compute_weights(labels)
+
+
+class BoundingBoxDetectionDataset(LocalDataset):
```
Fl4m3Ph03n1x, line 330: I would prefer we have this class in a separate file and accompany it with its own tests. 
The `dataset.py` file is becoming increasingly larger. Dividing into smaller chunks would make it more manageable IMO.

Also tests. We all agreed each new PR should add a test, in this case, the new class does not have tests yet :D

Fl4m3Ph03n1x (changes requested):

darwin/torch/dataset.py
FrancescoSaverioZuppichini: we don't have types in place in this part of the code. Would be better to make a separate PR to fix it

darwin/torch/dataset.py
```diff
@@ -217,7 +199,7 @@ def get_target(self, index: int):
             if "polygon" not in annotation and "complex_polygon" not in annotation:
                 print(f"Warning: missing polygon in annotation {self.annotations_path[index]}")
             # Extract the sequences of coordinates from the polygon annotation
-            annotation_type: str = "polygon" if "polygon" in annotation else "complex_polygon"
+            annotation_type = "polygon" if "polygon" in annotation else "complex_polygon"
```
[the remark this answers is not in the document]
FrancescoSaverioZuppichini: Nice point, idk. Keep in mind that part of the code was not changed by this PR

darwin/torch/dataset.py
FrancescoSaverioZuppichini: I agree different datasets should go in different files, this should be fixed by a different PR. Keep in mind testing dataset, in this case, is not trivial with the current darwin APIs.

I lost the test files, if you remember I had problems with my previous PR because `config.json` was still tracked. I don't think there is a way to get my files back and no time to write the tests from scratch again now. To be fair, nothing is tested there.

darwin/torch/dataset.py
```diff
@@ -118,31 +114,17 @@ def get_target(self, index: int):
...
-        if self.is_multi_label:
-            target = torch.zeros(len(self.classes))
-            # one hot encode all the targets
-            for tag in tags:
-                idx = self.classes.index(tag)
-                target[idx] = 1
-        else:
-            target = torch.tensor([self.classes.index(tags[0])])
-
+        if len(tags) > 1:
```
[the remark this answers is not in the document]
FrancescoSaverioZuppichini: Yes, there must be *at least* two. Will change the condition to make this more explicit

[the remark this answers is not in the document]
FrancescoSaverioZuppichini: that condition is not part of this PR

darwin/torch/dataset.py
```diff
@@ -359,3 +364,91 @@ def measure_weights(self, **kwargs) -> np.ndarray:
...
+        Returns
+        -------
+        class_weights : ndarray[double]
+            Weight for each class in the train set (one for each class) as a 1D array normalized
+        """
+        # Collect all the labels by iterating over the whole dataset
+        labels = []
+        for i, _ in enumerate(self.images_path):
+            target = self.get_target(i)
+            print(target)
```
kamilc, line 452: Is this dev-time and debug only? Should we get rid of this line (or turn it into a debug log message?)

kamilc (approved):
Only two minor comments. I'm assuming that you've tested the stacking - basing on this it LGTM

darwin/torch/dataset.py
```diff
@@ -79,7 +80,7 @@ def get_dataset(
 
 
 class ClassificationDataset(LocalDataset):
-    def __init__(self, transform: Optional[Callable | List] = None, **kwargs):
+    def __init__(self, transform: Optional[List] = None, **kwargs):
```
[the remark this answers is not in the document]
FrancescoSaverioZuppichini: No

darwin/torch/dataset.py
FrancescoSaverioZuppichini: nice catch

PR #212, opened 2021-09-29
added hack to expand left menu sidebar

Fl4m3Ph03n1x:
What
----

Added hack to automatically expand left menu Packages list when opening a page.

Why
----

We agreed on a meeting that it would be more convenient and explicit for our users to have the list of package and subpackages expanded automatically upon opening the page. 
 
How
----

In a rather surprising twist, the option to have the left menu bar expanded is bugged and DOES NOT WORK. 
I had to resort to hacking the crap out of the DOM to make it happen. I tried making the code readable. Sorry for the suffering caused to any reviewers/victims.

How can I test this?
----

1. Download branch
2. Install dependencies
3. Run `rm -f README.rst && m2r README.md && mv README.rst source/ && rm -rf docs/*  && sphinx-apidoc -f -o source darwin && sphinx-build -b html source/ docs/`
4. Go to `docs/index.html` and open it in your browser

Additional Info
-----

Linear ticket: https://linear.app/v7labs/issue/V7-1718/prevent-left-menu-items-from-being-collapsed

### DIFF
--- a/source/_static/js/custom.js
+++ b/source/_static/js/custom.js
@@ -1,42 +1,40 @@
-window.addEventListener('load', (_event) => {
-    var menu = document.querySelector(".wy-menu ul li:first-child")
-    recurse(menu)
+window.addEventListener("load", (_event) => {
+  const menu = document.querySelector(".wy-menu ul li:first-child")
+  recurse(menu)
 });
 
 /**
  * Given a Node, it recursively goes through every child and checks if the child is expandable, it
  * expands it unless it is already expanded.
- * 
- * @param {Node} node 
+ *
+ * @param {Node} node   The node to evaluate.
  */
-function recurse(node) {
-    if (is_expandable(node) && !is_expanded(node)) {
-        node.classList.add("current")
-    }
+const recurse = (node) => {
+  if (isExpandable(node) && !isExpanded(node)) {
+    node.classList.add("current")
+  }
 
-    // By default, children are not arrays, so we need to convert them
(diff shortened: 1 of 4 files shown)


source/_static/js/custom.js
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Returns whether or not the given expandable node is already expanded.
+ * Nodes are considered expandaded if they are 'current'ly selected, so we take advantage of this.
+ * 
+ * @param {Node} node 
+ * @returns {boolean} true if the node is already expanded, false otherwise.
+ */
+function is_expanded(node) {
+    return node.classList.contains("current")
+}
```
begedin, line 42: This is not frontend, so I don't think it matters, but if you want to follow the conventions we have there, then

- we use camelCase to name things
- we prefer arrow over regular functions
- we ident using spaces, 2 spaces per indent level
- we make the line limit 100 chars, for comments and code.

To use this function as example

```suggestion
const isExpanded = (node) => node.classList.contains("current")
```

Posting this as a comment because I really want to stress out that this is just a note and in no way a valid review.

andreaazzini: @Fl4m3Ph03n1x let's stick with the convention Nikola mentioned

Fl4m3Ph03n1x: done!

begedin (approved):
Approved, but I don't think my vote should be the deciding one.

andreaazzini (approved):

PR #214, opened 2021-10-01
[DRAFT] [V7-1732] Possibly block unsuccessful responses

andreaazzini:


darwin/client.py
```diff
@@ -130,6 +132,9 @@ def put(
             time.sleep(10)
             return self.put(endpoint, payload=payload, retry=False)
 
+        if response.status_code != 200 and block_unsuccessful_requests:
+            raise UnsuccessfulResponse()
```
Fl4m3Ph03n1x: Could we add the response body or more information to the error?
Logging something could also be useful.

Fl4m3Ph03n1x (changes requested):

darwin/client.py
```diff
@@ -130,6 +131,9 @@ def put(
             time.sleep(10)
             return self.put(endpoint, payload=payload, retry=False)
 
+        if block_unsuccessful_requests:
+            response.raise_for_status()
```
Fl4m3Ph03n1x, line 135: Are you sure you dont want to log the response body or other information here?

Fl4m3Ph03n1x (changes requested):

tests/darwin/client_test.py
```diff
@@ -0,0 +1,117 @@
...
+import pytest
+import responses
+from requests.exceptions import HTTPError
+
+from darwin.client import Client
+from darwin.config import Config
+from darwin.exceptions import InsufficientStorage, Unauthorized
+
+
+def describe_put():
```
Fl4m3Ph03n1x, line 13: While not an extensive list of tests (there are several use cases missing) it does cover for the very basic usage of the PUT method. 

Also, first set of tests for client !

tests/darwin/client_test.py
```diff
@@ -0,0 +1,117 @@
+from pathlib import Path
+from typing import Any, Dict
+
+import pytest
+import responses
+from requests.exceptions import HTTPError
+
+from darwin.client import Client
+from darwin.config import Config
+from darwin.exceptions import InsufficientStorage, Unauthorized
```
FrancescoSaverioZuppichini, line 10: IMHO not so pythonic names, `InsufficientStorageError, UnauthorizedError` are better

tests/darwin/client_test.py
FrancescoSaverioZuppichini: love them

darwin/client.py
```diff
@@ -121,20 +126,28 @@ def put(
...
+        if block_unsuccessful_requests:
+            self._print_debug(response, endpoint, payload)
+            response.raise_for_status()
+
         if raw:
             return response
         else:
             return self._decode_response(response, debug)
 
+    def _print_debug(self, response: Response, endpoint: str, payload: Dict) -> None:
```
FrancescoSaverioZuppichini, line 143: I think this is not best practice, we should start using logging at some point, for now just print. The function name is not clear, considering the inputs should be something like `self._print_response` or something like that. I think there is no need to have a separate method (assuming this is used only once). What do you think?

FrancescoSaverioZuppichini (changes requested):

tests/darwin/client_test.py
Fl4m3Ph03n1x: Yes, but I didn't create them, I am just going with the flow here :P

darwin/client.py
Fl4m3Ph03n1x: It is used twice in the code, but yes. This function is (hopefully) not permanent and I added it to make sure the code is DRY and that I don't forget to revert it back to its original state. 

This branch is supposed to be temporary (for the most part), so after we find out what the issue is, I will likely remove this function and add the fix together with the necessary tests.

FrancescoSaverioZuppichini (approved):

Fl4m3Ph03n1x (approved):

PR #215, opened 2021-10-04
Added try except for symlink creation.

Fl4m3Ph03n1x:
Why?
---

When creating a symlink in an Unix system in Azure, the creation of the symlink fails because Azure machines do not have the necessary permissions to do it. Thus code the raises and everything dies.

We previously thought this issue was only in Windows machines, but now we realize it can happen in any environment and that being a Unix machine is only bu a small factor.

What
----

We want to make it so that symlink creation does not raise under any circumstances. TO achieve this I added a try/except block.

I am not really that happy with this solution for 2 reasons:

 - I cannot test it in an _easy_ way. If you see the test and think there are too many nesting levels... You are right!
 - I am using a print instead of a Logger. Python does have a logger: https://docs.python.org/3/howto/logging.html . I defend we give it a look.

How can I test this?
----

Just simple `pytest` should do it. The test should cover this use case.

### DIFF
--- a/darwin/dataset/remote_dataset.py
+++ b/darwin/dataset/remote_dataset.py
@@ -282,12 +282,15 @@ class RemoteDataset:
         make_class_lists(release_dir)
 
         if release.latest and is_unix_like_os():
-            latest_dir: Path = self.local_releases_path / "latest"
-            if latest_dir.is_symlink():
-                latest_dir.unlink()
+            try:
+                latest_dir: Path = self.local_releases_path / "latest"
+                if latest_dir.is_symlink():
+                    latest_dir.unlink()
 
-            target_link: Path = self.local_releases_path / release_dir.name
-            latest_dir.symlink_to(target_link)
+                target_link: Path = self.local_releases_path / release_dir.name
+                latest_dir.symlink_to(target_link)
+            except OSError:
+                print("WARNING: Tried to create Symlink and failed. Continuing ...")
 
         if only_annotations:
             # No images will be downloaded
--- a/tests/darwin/dataset/remote_dataset_test.py
+++ b/tests/darwin/dataset/remote_dataset_test.py
@@ -1,6 +1,7 @@
 import json
 import types
 from datetime import datetime
+from pathlib import Path
 from typing import Any, Dict
 from unittest.mock import MagicMock, patch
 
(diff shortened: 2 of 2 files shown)


darwin/dataset/remote_dataset.py
```diff
@@ -282,12 +282,15 @@ def pull(
...
-            latest_dir.symlink_to(target_link)
+            try:
+                latest_dir: Path = self.local_releases_path / "latest"
+                if latest_dir.is_symlink():
+                    latest_dir.unlink()
+
+                target_link: Path = self.local_releases_path / release_dir.name
+                latest_dir.symlink_to(target_link)
+            except OSError:
+                print("WARNING: Tried to create Symlink and failed. Continuing ...")
```
andreaazzini, line 293: I wouldn't go with a log because we're not logging anything, we are warning our users about something.

I think we could use `rich.console` to print a message here. 

```python
console.print(f"Could not mark release {release.name} as latest.", style="warning")
```

What do you think?

andreaazzini (changes requested):

darwin/dataset/remote_dataset.py
Fl4m3Ph03n1x: Given this is part of the SDK, I lean towards logging instead of printing. 

Quoting an SO answer:

```
(...) if your project is meant to be imported by other python tools, it's bad practice for your package to print things to stdout, since the user likely won't know where the print messages are coming from. With logging, users of your package can choose whether or not they want to propogate logging messages from your tool or not. (...)
```

I believe this quote fits our use case, since this module is technically part of the SDK, which will be imported by someone at some point.

It is also because of this general rule that I defend the SDK code should never print anything, but either raise Exceptions or return ERROR CODES. It would then be the job of the CLI layer to convert those into meaningful `print` messages. 

But this falls out of the scope of this PR.

I can use `rich.console` if you'd like, let me know on the comments below !

Biblio:

- https://stackoverflow.com/a/6918596/1337392

andreaazzini: `rich.console` has a `log` function in addition to `print` https://rich.readthedocs.io/en/stable/console.html

shall we try with that one?

Fl4m3Ph03n1x: Done, let me know what you think !

Fl4m3Ph03n1x: I would like to make a comment that I am not a fan of how I am using `console` in this example. 

Console is a dependency with a side effect (that of writing output to a terminal). We should do dependency injection by constructor, and inject this dependency when creating the remote dataset. However, I have not found the right place to do that in the project unfortunately and I don't think we will be able to take proper advantage of dependency injection without a rewrite of some of core parts of darwin. 

So I guess we'll have to save this one for later....

FrancescoSaverioZuppichini: I feel both of you, we should use `logging` and inject custom handlers

FrancescoSaverioZuppichini:
@Fl4m3Ph03n1x great PR, I don't get why we need to create `symlink` then I don't understand why we don't want the code to fail

andreaazzini (approved):

PR #216, opened 2021-10-05
V7 1686 find a place to host our documentation

Fl4m3Ph03n1x:


.github/workflows/tests.yml
```diff
@@ -1,4 +1,4 @@
-name: Python package
```
FrancescoSaverioZuppichini, line 1: should we run test when we do PR?

.github/workflows/document.yml
```diff
@@ -1,4 +1,4 @@
-name: Python package
+name: Documentation
```
FrancescoSaverioZuppichini, line 1: I think we need to build the doc only when we push on master

FrancescoSaverioZuppichini (changes requested):
I think we should change when we run the actions

.github/workflows/document.yml
FrancescoSaverioZuppichini: We don't want to build the doc any time a new pull request is made to master but we want to re-build the doc when the code on master changes. We should change `pull_request` to `push`

Fl4m3Ph03n1x: changed

PR #219, opened 2021-10-05
[V7-1774] Add class to dataset based on annotation class id

simedw:


darwin/dataset/remote_dataset.py
```diff
@@ -441,13 +441,13 @@ def create_annotation_class(self, name: str, type: str, subtypes: List[str] = []
...
 
-    def add_annotation_class(self, annotation_class: AnnotationClass) -> Union[Dict, None]:
+    def add_annotation_class(self, annotation_class: Union[AnnotationClass, int]) -> Union[Dict, None]:
         """
         Adds an annotation class to this dataset.
 
         Parameters
         ----------
-        annotation_class : AnnotationClass
+        annotation_class : AnnotationClass or annotation_class_id (int)
```
andreaazzini, line 450: Not sure if this is the right syntax, this should be a type. The description belongs to the line below. @Fl4m3Ph03n1x could you confirm?

andreaazzini (changes requested):
Small request, we are making a change to the SDK without changing any CLI function, is that intended?

darwin/dataset/remote_dataset.py
Fl4m3Ph03n1x: The typing is correct!

Fl4m3Ph03n1x (approved):

Fl4m3Ph03n1x:
I approve this change as I understand we need it right away to get HP as a client. 
However, I defend we should add a test to check this functionality.

darwin/dataset/remote_dataset.py
andreaazzini: What I meant is: shouldn't those lines be
```
        Parameters
        ----------
        annotation_class : Union[AnnotationClass, int]
            The annotation class object or ID to add.
```

PR #221, opened 2021-10-07
fixed check on filename not existing

Fl4m3Ph03n1x:
Why
----

Upon reviewing the code for `pascal_voc` files I realized that the `<filename></filename>` child element could be empty or simply contain a `' '` space. 

I fixed this issue to avoid future problems. 

How can I test this?
-----------------------------
Since I added a test, running `pytest` should do it.

### DIFF
--- a/darwin/importer/formats/pascalvoc.py
+++ b/darwin/importer/formats/pascalvoc.py
@@ -1,11 +1,11 @@
 import xml.etree.ElementTree as ET
 from pathlib import Path
-from typing import List, NoReturn, Union
+from typing import List, Union
 
 import darwin.datatypes as dt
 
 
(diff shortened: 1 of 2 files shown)


darwin/importer/formats/pascalvoc.py
```diff
@@ -44,6 +44,6 @@ def _find_element(source: ET.Element, name: str) -> Union[ET.Element, NoReturn]:
 # Private
 def _find_text_value(source: ET.Element, name: str) -> Union[str, NoReturn]:
     element = _find_element(source, name)
-    if element is None or element.text is None:
+    if not element.text.strip():
```
andreaazzini, line 47: `element` might be `None` right? In that case, `element.text.strip()` should raise?

PR #222, opened 2021-10-07
Add DARWIN_UPLOAD_CHUNK_SIZE environment variable

andreaazzini:
# Why

We are currently struggling with large dataset upload tasks. That mostly depends on our query plans being extremely nondeterministic. Our investigation revealed that acting on the upload chunk size may lead our db to choose a favourable query plan in some cases.

# How

This PR adds a `DARWIN_UPLOAD_CHUNK_SIZE` environment variable, which can be set by the user to change the chunk size during upload. It defaults at `500`.

### DIFF
--- a/darwin/dataset/upload_manager.py
+++ b/darwin/dataset/upload_manager.py
@@ -287,6 +287,15 @@ DEFAULT_UPLOAD_CHUNK_SIZE: int = 500
 
 
 def _upload_chunk_size() -> int:
+    """
+    Gets the chunk size to be used from the OS environment, or uses the default one if that is not
+    possible. The default chunk size is 500.
+
+    Returns
+    -------
+    int
+        The chunk size to be used.
+    """
     env_chunk: Optional[str] = os.getenv("DARWIN_UPLOAD_CHUNK_SIZE")
     if env_chunk is None:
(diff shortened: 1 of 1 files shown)


darwin/dataset/upload_manager.py
```diff
@@ -278,3 +280,17 @@ def callback(monitor):
...
+DEFAULT_UPLOAD_CHUNK_SIZE: int = 500
+
+
+def _upload_chunk_size() -> int:
+    env_chunk: Optional[str] = os.getenv("DARWIN_UPLOAD_CHUNK_SIZE")
+    if env_chunk is None:
+        return DEFAULT_UPLOAD_CHUNK_SIZE
+
+    try:
+        return int(env_chunk)
```
Fl4m3Ph03n1x, line 294: Can we add a log/rich message saying that the ENV var is not properly set?

Fl4m3Ph03n1x (changes requested):

darwin/dataset/upload_manager.py
andreaazzini: Done, I couldn't manage to make it work with `rich` and we need this pushed, so I used good ol' `print`

PR #224, opened 2021-10-08
improved docs

Fl4m3Ph03n1x:
Why
----

One of our users is having trouble importing a `pascalvoc` file. Upon further analysis, I concluded his file did not have the format we expected to find.

What
---

I have asked the client to change the format of his `pascalvoc` file. However I also understand the format we expected was not documented at all. 

I documented the expected format.

### DIFF
--- a/darwin/importer/formats/pascalvoc.py
+++ b/darwin/importer/formats/pascalvoc.py
@@ -8,6 +8,23 @@ import darwin.datatypes as dt
 def parse_file(path: Path) -> Optional[dt.AnnotationFile]:
     """
     Parses the given pascalvoc file and maybe returns the corresponding annotation.
+    The file must have the following structure:
+    
+    ```xml
+    <filename>SOME_FILE_NAME</filename>
+    <object>
+        <name>CLASS_NAME</name>
+        <bndbox>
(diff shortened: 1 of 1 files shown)


darwin/importer/formats/pascalvoc.py
```diff
@@ -8,6 +8,23 @@
 def parse_file(path: Path) -> Optional[dt.AnnotationFile]:
     """
     Parses the given pascalvoc file and maybe returns the corresponding annotation.
+    The file must have the following structure:
```
FrancescoSaverioZuppichini, line 11: I think the <annotation> tag is missing. The [official website](http://host.robots.ox.ac.uk/pascal/VOC/voc2007/) is down, I would like to add a link to it in the doc string

FrancescoSaverioZuppichini:
@Fl4m3Ph03n1x pls show the client file

darwin/importer/formats/pascalvoc.py
andreaazzini: https://gist.github.com/Prasad9/30900b0ef1375cc7385f4d85135fdb44

PR #225, opened 2021-10-08
Use Thread Pool to upload annotations

FrancescoSaverioZuppichini:
# Why
Uploading annotations take way too long
# How
I added a thread pool in `importer.py`. By default we have 4 workers, we got a linear speed up increase.

With 1 worker
image

4
image

8
image


darwin/importer/importer.py
```diff
@@ -205,20 +210,41 @@ def import_annotations(
...
+                dataset.client,
+                image_id,
+                remote_classes,
+                attributes,
+                parsed_file.annotations,
+                dataset,
+                append,
             )
 
+        print(f"Uploading annotations with {max_workers} workers")
```
Fl4m3Ph03n1x, line 236: Could we use `rich.console` here?

darwin/importer/importer.py
```diff
@@ -205,20 +210,41 @@ def import_annotations(
...
+            if type(parsed_files) is not list:
+                parsed_files = [parsed_files]
+            # remove files missing on the server
+            missing_files = [missing_file.full_path for missing_file in local_files_missing_remotely]
+            parsed_files = [parsed_file for parsed_file in parsed_files if parsed_file.full_path not in missing_files]
+            return parsed_files
+
+        def import_annotation(parsed_file):
             image_id = remote_files[parsed_file.full_path]
             _import_annotations(
```
Fl4m3Ph03n1x, line 226: How is this going to behave if the `post` in the function fails and we raise?

FrancescoSaverioZuppichini: I don't know, I just refactor the code that was there

.vscode/settings.json
```diff
@@ -22,5 +22,6 @@
     "--profile",
     "black"
   ],
-  "python.testing.pytestEnabled": true
+  "python.testing.pytestEnabled": true,
+  "python.pythonPath": "/home/zuppif/miniconda3/bin/python"
```
kamilc, line 26: Can you unstage this local dev-env specific line? Thanks!

kamilc (changes requested):
One tiny request - otherwise it looks good

.vscode/settings.json
FrancescoSaverioZuppichini: yes

FrancescoSaverioZuppichini (commented):
Done!

darwin/importer/importer.py
```diff
@@ -105,6 +108,7 @@ def import_annotations(
     importer: Callable[[Path], Union[List[dt.AnnotationFile], dt.AnnotationFile, None]],
     file_paths: List[Union[str, Path]],
     append: bool,
+    max_workers: Optional[int] = 4,
```
andreaazzini, line 111: This argument is not optional actually, as it defaults to an integer
```suggestion
    max_workers: int = 4,
```

darwin/importer/importer.py
```diff
@@ -120,6 +124,8 @@ def import_annotations(
         A list of `Path`'s or strings containing the Annotations we wish to import.
     append : bool
         If `True` appends the given annotations to the datasets. If `False` will override them.
+    max_workers: Optional[int]
```
andreaazzini, line 127:
```suggestion
    max_workers: int
```

darwin/importer/importer.py
```diff
@@ -205,20 +210,41 @@ def import_annotations(
...
+        data = set(local_file.path for local_file in local_files)
+        # let's get the files we need to process
+        stage = executor.map(get_parsed_files, data)
+        out = list(stage)
+        parsed_files = chain.from_iterable(out)
+        # we have to get the total size, so let's convert parsed_files to list
+        parsed_files = list(parsed_files)
+        # process the file
+        stage = track(executor.map(import_annotation, parsed_files), total=len(parsed_files))
+        list(stage)
```
andreaazzini, line 246: It's not immediately easy for me to understand this code. That's probably a sign it needs a few comments to document the steps we're taking?

FrancescoSaverioZuppichini: updated the comments

darwin/importer/importer.py
FrancescoSaverioZuppichini: updated

.gitignore
```diff
@@ -1,7 +1,7 @@
 __pycache__/
 .pytest_cache/
 .cache/
-
+.vscode/
```
pfac, line 4: Do we really want to do this? It won't remove existing settings but it will ignore any further changes right?

I'm assuming you added this because of [Kamil's comment](https://github.com/v7labs/darwin-py/pull/225#discussion_r725867447). If that's the case, then I don't think this is the correct solution.

Instead of adding `"python.pythonPath": "/home/zuppif/miniconda3/bin/python"` to the project's VSCode settings, you can add it to your own VSCode's settings for Python.

Alternatively, if you want to set the Python path for this project only, then you can add this line to a [global `.gitignore` file in your system](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer), or [to a file under `.git/info/exclude`](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#excluding-local-files-without-creating-a-gitignore-file). Both these situations will ignore changes to these files in your system, but still allow others to keep changing the file in the repo.

PR #226, opened 2021-10-12
[HOTFIX] Strip file stems correctly

andreaazzini:


kamilc (approved):

darwin/dataset/local_dataset.py
```diff
@@ -280,3 +263,38 @@ def __str__(self):
...
+) -> Iterator[str]:
+    # If no partition is specified, then take all json files in the annotations directory.
+    # The resulting generator prepends parent directories relative to the main annotation directory.
+    #
+    # E.g.: ["annotations/test/1.json", "annotations/2.json", "annotations/test/2/3.json"]:
+    #     - annotations/test/1
+    #     - annotations/2
+    #     - annotations/test/2/3
+    if partition is None:
+        return (str(e.relative_to(annotations_dir).parent / e.stem) for e in annotations_dir.glob("**/*.json"))
```
FrancescoSaverioZuppichini, line 284: could we add a comment here explaining what are we doing? Btw we are returning a set, while the return type is `Iterator`, It is true that a set is an iterator but we could just return

```python
map(lambda e: str(e.relative_to(annotations_dir).parent / e.stem), annotations_dir.glob("**/*.json")`
```

that is a real iterator

PR #227, opened 2021-10-13
adding typing to cli_functions

Fl4m3Ph03n1x:


darwin/cli_functions.py
```diff
@@ -5,8 +5,8 @@
 import sys
 from itertools import tee
 from pathlib import Path
-from typing import Dict, List, NoReturn, Optional, Union
-
+from typing import Callable, Dict, Iterator, List, NoReturn, Optional, Union, Tuple
+from darwin.dataset.release import Release
```
andreaazzini, line 9: I'm pretty sure this shouldn't belong here. Are you not automatically re-ordering import statements?

darwin/client.py
```diff
@@ -1,7 +1,7 @@
 import os
 import time
 from pathlib import Path
-from typing import Any, Dict, Iterator, Optional, Union
+from typing import Any, Dict, Iterator, Optional, Union, Tuple
```
andreaazzini, line 4: And ordering seems wrong, so yeah definitely something that doesn't work on your end when it comes to imports

darwin/client.py
```diff
@@ -570,7 +570,7 @@ def parse_version(version_str):
             pass
 
     @property
-    def newer_darwin_version(self):
+    def newer_darwin_version(self) -> Optional[Tuple[int, int, int]]:
```
andreaazzini, line 573: Not sure what you think, but I'd be super happy to introduce our own `typings.py`. In this case
```python
DarwinVersionNumber = Tuple[int, int, int]
```
so here we would do `Optional[DarwinVersionNumber]`

andreaazzini: When do I think these are necessary? Basically, every time the time is not self-documenting the function signature

darwin/client.py
Fl4m3Ph03n1x: I believe this was fixed.

darwin/cli_functions.py
```diff
@@ -189,30 +193,30 @@ def path(dataset_slug: str) -> Path:
...
+        report: str = remote_dataset.get_report(granularity)
         print(report)
     except NotFound:
         _error(f"Dataset '{dataset_slug}' does not exist.")
 
 
 def export_dataset(
-    dataset_slug: str, include_url_token: bool, annotation_class_ids: Optional[List] = None, name: Optional[str] = None
-):
+    dataset_slug: str, include_url_token: bool, name: str, annotation_class_ids: Optional[List] = None
```
andreaazzini, line 218: The way I see it is: `List`, `Tuple`, `Dict`, etc should always specify its items' types.
In this case, I believe this would be
```suggestion
    dataset_slug: str, include_url_token: bool, name: str, annotation_class_ids: Optional[List[int]] = None
```

darwin/cli_functions.py
```diff
@@ -568,12 +576,12 @@ def file_upload_callback(file_name, file_total_bytes, file_bytes_sent):
         _error(f"No files found")
 
 
-def dataset_import(dataset_slug, format, files, append):
-    client = _load_client(dataset_identifier=dataset_slug)
-    parser = find_supported_format(format, darwin.importer.formats.supported_formats)
+def dataset_import(dataset_slug, format, files, append) -> None:
```
andreaazzini, line 579: args are not typed here

darwin/cli_functions.py
```diff
@@ -637,27 +645,41 @@ def set_file_status(dataset_slug: str, status: str, files: List[str]):
         _error(f"No dataset with name '{e.name}'")
 
 
-def find_supported_format(query, supported_formats):
+def find_import_supported_format(
+    query: str,
+    supported_formats: List[ImporterFormat],
+) -> ImportParser:
```
andreaazzini, line 651: Nice, this is very readable to me 👍

darwin/cli_functions.py
```diff
@@ -637,27 +645,41 @@ def set_file_status(dataset_slug: str, status: str, files: List[str]):
...
             return fmt_parser
     list_of_formats = ", ".join([fmt for fmt, _ in supported_formats])
-    _error(f"Unsupported format, currently supported: {list_of_formats}")
+    _error(f"Unsupported export format, currently supported: {list_of_formats}")
 
 
-def dataset_convert(dataset_slug: str, format: str, output_dir: Union[str, Path, None] = None):
-    client = _load_client()
-    parser = find_supported_format(format, darwin.exporter.formats.supported_formats)
+def dataset_convert(dataset_slug: str, format: str, output_dir: Union[str, Path, None] = None) -> None:
```
andreaazzini, line 670: What do you think of making a new type
```python
PathLike = Union[str, Path]
```
so here would be
```suggestion
def dataset_convert(dataset_slug: str, format: str, output_dir: Optional[PathLike] = None) -> None:
```

The reason why I think it's important is that we have `Union[str, Path]` all over the place

darwin/cli_functions.py
```diff
@@ -669,12 +691,12 @@ def dataset_convert(dataset_slug: str, format: str, output_dir: Union[str, Path,
...
 
-def convert(format: str, files: List[Union[str, Path]], output_dir: Path):
-    parser = find_supported_format(format, darwin.exporter.formats.supported_formats)
+def convert(format: str, files: List[Union[str, Path]], output_dir: Path) -> None:
+    parser: ExportParser = find_export_supported_format(format, ExportSupportedFormats)
     exporter.export_annotations(parser, files, output_dir)
 
 
-def help(parser, subparser: Optional[str] = None):
+def help(parser, subparser: Optional[str] = None) -> None:
```
andreaazzini, line 699: `parser` is untyped

darwin/config.py
```diff
@@ -119,6 +119,6 @@ def get_default_team(self, raise_on_invalid_team: bool = True):
                 return None
         return self.get_team(teams[0])
 
-    def get_all_teams(self):
+    def get_all_teams(self) -> List[Optional[Dict[str, Any]]]:
```
andreaazzini, line 122: I'm wondering: what do we do with `None` teams?

Is there a chance we could filter `None` teams out of the returned list?

darwin/exporter/exporter.py
```diff
@@ -25,7 +25,7 @@ def darwin_to_dt_gen(file_paths):
 
 
 def export_annotations(
-    exporter: Callable[[Generator[dt.AnnotationFile, None, None], Path], None],
+    exporter: ExportParser,
     file_paths: List[Union[str, Path]],
     output_directory: Union[str, Path],
```
andreaazzini, line 30: Here's an example of where `PathLike` would be useful to have

darwin/importer/formats/__init__.py
```diff
@@ -1,6 +1,10 @@
+from typing import List
+
+from darwin.types import ImporterFormat
+
 from . import coco, csvtags, csvtagsvideo, darwin, dataloop, pascalvoc
 
-supported_formats = [
+supported_formats: List[ImporterFormat]  = [
```
andreaazzini, line 7: Not sure if this belongs to the scope of this PR (probably not), but I was wondering if `supported_formats` could be turned into a dictionary.

After all, it seems to be used to map a type name to a function.

darwin/cli_functions.py
andreaazzini: Any reasons why we swapped `name` and `annotation_class_ids`?

andreaazzini (changes requested):
This is a great start!

I found a few things that need attention, and made a few suggestions.
I might have missed a few other places but I think running `mypy` should help find those cases.

(Also, maybe it's time to add `mypy` to our CI!)

darwin/cli_functions.py
```diff
@@ -73,8 +77,8 @@ def authenticate(api_key: str, default_team: Optional[bool] = None, datasets_dir
         API key to use for the client login
     default_team: bool
```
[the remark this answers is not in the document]
Fl4m3Ph03n1x: This PR focuses on types. The documentation will be fixed later on, once I have all the types working properly (or at least working to the extent where I can, because we know there are type inconsistencies that can only be fixed by re-writes in some areas).

It is in that second pass that I will pay more attention to documentation and enforce a standard. Right now I am mostly ignoring it.

darwin/cli_functions.py
Fl4m3Ph03n1x: I like it !

darwin/cli_functions.py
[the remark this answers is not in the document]
Fl4m3Ph03n1x: I have still updated it.

darwin/config.py
Fl4m3Ph03n1x: FIxed, I think :D

darwin/cli_functions.py
```diff
@@ -189,30 +193,30 @@ def path(dataset_slug: str) -> Path:
     )
 
 
-def url(dataset_slug: str) -> Path:
-    """Returns the url of the specified dataset"""
-    client = _load_client(offline=True)
+def url(dataset_slug: str) -> None:
+    """Prints the url of the specified dataset"""
```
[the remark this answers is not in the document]
Fl4m3Ph03n1x: https://linear.app/v7labs/issue/V7-1977/use-richconsole-in-darwin-py

darwin/importer/formats/__init__.py
Fl4m3Ph03n1x: I agree, but for the time being, I feel this PR is already big as it is. 
It definitely grew beyond what I was expecting, as it turns out, it looks like the Typing issue we had with Darwin-py was considerably larger than I initially estimated.

Which means it is a good thing we are working on it now. This work will not only benefit the docs, it has already resulted in several pre-emptive bugfixes. 

As for the dictionary, I defend we create a separate PR for it.

darwin/cli_functions.py
FrancescoSaverioZuppichini: I think this is an antipattern, the whole goal of types is to provide a unique way to identify what is what, by doing this you violate it. A better strategy is to always use Path where we need a Path instance and string where we need a string.

andreaazzini (approved):
This looks good to me. Keep in mind I've pushed a commit from my end (`format`) that includes formatting for all the files in the project. I think your configuration of VSCode needs a few adjustments I'm afraid 😕

PR #228, opened 2021-10-15
Fixed wrong typing

Fl4m3Ph03n1x:


Fl4m3Ph03n1x:
I don't know how many more `Any`s we have, but I am guessing I will find them eventually. 
The one thing where `Any` is rather prevalent (and I don't really like it) is in `Dict[str, Any]` which is basically everything we return from `get`, `put`, `post` and `delete` functions in `client`. 

I would much rather have the classes for the structures we use instead of simply have dictionaries all over the place, but this discussion does not belong in here.

PR #229, opened 2021-10-15
fixed types and bugs

Fl4m3Ph03n1x:
- fixed types
- fixed bugs regarding teams being none and causing crashes

### DIFF
--- a/darwin/client.py
+++ b/darwin/client.py
@@ -1,7 +1,7 @@
 import os
 import time
 from pathlib import Path
-from typing import Any, Callable, Dict, Iterator, List, Optional, Union
+from typing import Any, Dict, Iterator, List, Optional, Union
 
 import requests
 
@@ -16,7 +16,7 @@ from darwin.exceptions import (
     NotFound,
     Unauthorized,
 )
-from darwin.types import DarwinVersionNumber, Team
+from darwin.types import DarwinVersionNumber, ErrorHandler, Team
 from darwin.utils import is_project_dir, urljoin
 from darwin.validators import name_taken, validation_error
 
@@ -84,7 +84,7 @@ class Client:
     def put(
         self,
         endpoint: str,
-        payload: Dict[Any, Any],
+        payload: Dict[str, Any],
         team: Optional[str] = None,
         retry: bool = False,
         debug: bool = False,
@@ -203,7 +203,7 @@ class Client:
         endpoint: str,
         team: Optional[str] = None,
         retry: bool = False,
-        error_handlers: Optional[List[Callable[[int, str], None]]] = None,
+        error_handlers: Optional[List[ErrorHandler]] = None,
         debug: bool = False,
     ) -> Dict[str, Any]:
         """Delete something new on the server trough HTTP
--- a/darwin/types.py
+++ b/darwin/types.py
@@ -18,3 +18,5 @@ ImporterFormat = Tuple[str, ImportParser]
 
 PathLike = Union[str, Path]
 Team = Dict[str, Any]
+
+ErrorHandler = Callable[[int, str], None]


darwin/client.py
```diff
@@ -1,10 +1,10 @@
 import os
 import time
 from pathlib import Path
-from typing import Any, Dict, Iterator, Optional, Union
+from typing import Any, Dict, Iterator, Optional, Tuple, Union, List, Callable
+from darwin.types import Team
```
andreaazzini, line 5: Hate to say this, but you have some issues with your IDE which I think require immediate attention, so I don't bother you all the time with this kind of comments.

darwin/client.py
```diff
@@ -23,16 +23,16 @@
...
-        self.config = config
-        self.url = config.get("global/api_endpoint")
-        self.base_url = config.get("global/base_url")
-        self.default_team = default_team or config.get("global/default_team")
-        self.features: dict = {}
+        self.config: Config = config
+        self.url: str = config.get("global/api_endpoint")
+        self.base_url: str = config.get("global/base_url")
+        self.default_team: str = default_team or config.get("global/default_team")
+        self.features: Dict[str, Any] = {}
```
andreaazzini, line 30: I wonder if we can do better than `Any`.

Can we define something like this in our `types.py`
```python
from dataclasses import dataclass

@dataclass
class FeaturePayload:
    """Structured payload of a Feature record on V7 Darwin"""
    name: str
    enabled: boolean
```

This way, we can define `self.features: Dict[str, FeaturePayload]`.

This method is very similar to what we do on the FE side (TypeScript there).

Note: the reason why I'm using a `dataclass` here is that we can't use `TypedDict`, which I believed was introduced in Python 3.8 if I'm not mistaken.

Note: we probably need to edit `load_feature_flags`. We currently do:
```python
self.features[team_slug] = self.get(f"/teams/{team_slug}/features")
```
which is terrible for many reasons.

darwin/client.py
```diff
@@ -84,12 +84,12 @@ def get(
     def put(
         self,
         endpoint: str,
-        payload: Dict,
+        payload: Dict[Any, Any],
```
andreaazzini, line 87: Can't this be turned at least into:
```suggestion
        payload: Dict[str, Any],
```

Same for other `payload` arguments down below

darwin/client.py
```diff
@@ -197,9 +203,9 @@ def delete(
         endpoint: str,
         team: Optional[str] = None,
         retry: bool = False,
-        error_handlers: Optional[list] = None,
+        error_handlers: Optional[List[Callable[[int, str], None]]] = None,
```
andreaazzini, line 206: `Callable[[int, str], None]` to me is a good candidate to be made into its own type
```python
ErrorHandler = Callable[[int, str], None]
```

Also, can't we make `error_handler` non-optional, defaulting to `[]`? Not sure if that's going to be a problem tbh

darwin/client.py
```diff
@@ -353,30 +366,43 @@ def create_dataset(self, name: str, team: Optional[str] = None) -> RemoteDataset
 
     def fetch_remote_classes(self, team: Optional[str] = None) -> Optional[Dict[str, Any]]:
         """Fetches all remote classes on the remote dataset"""
-        team = self.config.get_team(team or self.default_team)
+        the_team: Optional[Team] = self.config.get_team(team or self.default_team)
```
andreaazzini, line 369: I would actually rename the argument to be `team_name` or `team_slug` (depending on what this string actually represents), so this one can be named `team`.

I know it's breaking it for people who call `fetch_remote_classes(team=team)`, but I think it's fine for a minor upgrade

andreaazzini: Same in other functions below.

Issue #232, opened 2021-10-20
`detectron2_register_dataset` function anormal behaviour when the release_name parameter is defined


sachasamama:
I detect an error when we defined the release_name as parameter (which is optional and take "latest" value by default)

The DatasetCatalog will register a wrong catalog if there is a difference between the "latest" version and the "release_name"

**To solve this issue :**
You need to replace the 103rd row in darwin.torch.utils.py `detectron2_register_dataset()`

classes = get_classes(dataset_path, annotation_type="polygon")

by 

-----> classes = get_classes(dataset_path=dataset_path, release_name=release_name, annotation_type="polygon")

PR #233, opened 2021-10-20
fix: Specify release_name parameter value in detectron2_register_datasets

sachasamama:
PR for the #232 issue.


andreaazzini (approved):
LGTM 👍

PR #235, opened 2021-10-25
Compute weights for Classification Dataset

FrancescoSaverioZuppichini:
## Why
Currently `compute_weights` doesn't work in `ClassificationDataset`s.

```python
from darwin.torch.dataset import ClassificationDataset
from pathlib import Path

root = Path(....)
ds = ClassificationDataset(dataset_path=root, release_name="latest")

print(ds.measure_weights())

```

## How
Correctly add the the `labels` list the labels based on single-multi label classification

Moreover, I used `pets` to test `ClassificationDataset` both in single and multi label tasks


FrancescoSaverioZuppichini:
I've added tests for the torch bindings. I've created three dummies datasets for single label, multi label classification and instance segmentation. They are very small, only 5 `50x50` images and corresponding annotation files. They are unzipped at runtime with a custom fixture

.vscode/settings.json
```diff
@@ -14,13 +14,9 @@
...
-  "python.sortImports.args": [
-    "--profile",
-    "black"
-  ],
-  "python.testing.pytestEnabled": true
-}
\ No newline at end of file
+  "python.sortImports.args": ["--profile", "black"],
+  "python.testing.pytestEnabled": true,
+  "python.pythonPath": "/home/zuppif/miniconda3/bin/python"
```
andreaazzini, line 21: ⚠️

andreaazzini: `git checkout origin/master .vscode/settings.json`?

FrancescoSaverioZuppichini: done

andreaazzini (approved):
LGTM!

PR #242, opened 2021-10-26
[V7-2084] SuperAnnotate vector JSON format importer

andreaazzini:


Fl4m3Ph03n1x:
Outdated by more recent PR that accomplish the same and then some.

PR #269, opened 2021-11-05
[V7-2192] Allow for item deletion in CLI

andreaazzini:
# Why

We don't support file deletion from the CLI

# How

Add argument parser option and functions, which are pretty similar to the ones for setting an item status.
Note: this PR is changing `delete` APIs, and is not retro-compatible. With that said, the new API is aligned to `post` and `put`, which I think is important.

Because of the aforementioned reasons, I would release a new minor version: `0.7.0`

### DIFF
--- a/darwin/cli_functions.py
+++ b/darwin/cli_functions.py
@@ -654,13 +654,13 @@ def set_file_status(dataset_slug: str, status: str, files: List[str]) -> None:
         _error(f"No dataset with name '{e.name}'")
 
 
-def delete_files(dataset_slug: str, files: List[str], yes: bool = False) -> None:
+def delete_files(dataset_slug: str, files: List[str], skip_user_confirmation: bool = False) -> None:
     client: Client = _load_client(dataset_identifier=dataset_slug)
     try:
         console = Console()
         dataset: RemoteDataset = client.get_remote_dataset(dataset_identifier=dataset_slug)
         items: Iterator[DatasetItem] = dataset.fetch_remote_files({"filenames": ",".join(files)})
-        if not yes and not secure_continue_request():
+        if not skip_user_confirmation and not secure_continue_request():
             console.print("Cancelled.")
             return
 
--- a/darwin/cli_functions.py
+++ b/darwin/cli_functions.py
@@ -657,6 +657,10 @@ def delete_files(dataset_slug: str, files: List[str]) -> None:
     try:
         dataset: RemoteDataset = client.get_remote_dataset(dataset_identifier=dataset_slug)
         items: Iterator[DatasetItem] = dataset.fetch_remote_files({"filenames": ",".join(files)})
+        if not secure_continue_request():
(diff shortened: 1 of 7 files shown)


Fl4m3Ph03n1x (approved):

darwin/cli_functions.py
```diff
@@ -652,6 +652,20 @@ def set_file_status(dataset_slug: str, status: str, files: List[str]) -> None:
         _error(f"No dataset with name '{e.name}'")
 
 
+def delete_files(dataset_slug: str, files: List[str]) -> None:
+    client: Client = _load_client(dataset_identifier=dataset_slug)
+    try:
+        dataset: RemoteDataset = client.get_remote_dataset(dataset_identifier=dataset_slug)
+        items: Iterator[DatasetItem] = dataset.fetch_remote_files({"filenames": ",".join(files)})
+        if not secure_continue_request():
+            print("Cancelled.")
```
andreaazzini, line 661: I personally think we're fine without this `print` call

andreaazzini (commented):
Can't request changes since I'm the author of the PR, but:
- I think we can get rid of a `print` call
- We're missing a `-y` flag, which would circumvent `secure_continue_request()`

darwin/cli_functions.py
Fl4m3Ph03n1x: The print call is standard in other places where we ask for confirmation as well, that's why I kept it.
I will look into the `-y` flag.

Fl4m3Ph03n1x:
- Added `-y` flag to bypass user prompt
- Added rich console spinner while we wait to delete files
- Added more tests
- Fixed incorrect behavior of DELETE and POST functions in client.py

Note
-----

The behavior of `client.py` DELETE and POST function is now **inconsistent** with that of PUT and GET. 
DELETE and POST will raise an exception if an unknown error is received by the server, while PUT and GET will do something else. 

I personally don't like this approach, but I suggest we start thinking about a proper interface for `client.py` because I only see this inconsistency causing more issues in the future.

darwin/cli_functions.py
```diff
@@ -652,18 +654,24 @@ def set_file_status(dataset_slug: str, status: str, files: List[str]) -> None:
         _error(f"No dataset with name '{e.name}'")
 
 
-def delete_files(dataset_slug: str, files: List[str]) -> None:
+def delete_files(dataset_slug: str, files: List[str], yes: bool = False) -> None:
```
andreaazzini, line 657: While I think it's a good idea to call this argument `yes` in the parser, I think this should be named differently here. Something like `skip_confirmation`?

Fl4m3Ph03n1x: Done !

darwin/cli_functions.py
Fl4m3Ph03n1x: Resolved over Slack call.

PR #270, opened 2021-11-10
[V7-2273] Item Count should consider both images and videos

andreaazzini:


FrancescoSaverioZuppichini (commented):
LGTM!

Fl4m3Ph03n1x (approved):

Fl4m3Ph03n1x:
Would be nice if we could test this somehow, but given the complexity and the setup needed I think we can skip it for now.

darwin/client.py
```diff
@@ -360,7 +360,7 @@ def create_dataset(self, name: str, team: Optional[str] = None) -> RemoteDataset
             team=team or self.default_team,
             slug=dataset["slug"],
             dataset_id=dataset["id"],
-            item_count=dataset["num_images"] + dataset["num_videos"],
+            item_count=dataset["num_images"],
```
Fl4m3Ph03n1x, line 363: having "num_videos" made sense to me, what was the issue this was causing?

Fl4m3Ph03n1x (changes requested):

darwin/client.py
andreaazzini: There's no `num_videos` in the `show.json` render view in backend. `create_dataset` fails if I include `num_videos`

Fl4m3Ph03n1x: oki!

Fl4m3Ph03n1x (approved):

PR #274, opened 2021-11-16
V7 2297 labelbox format importer

Fl4m3Ph03n1x:
Background
---

We are trying to add compatibility with LabelBox: 
https://docs.labelbox.com/docs/annotation-types-1

This PR includes:
 - Importation of bounding box **Images** from LabelBox into our system
 - Renamed some test files to avoid naming conflicts with pytest
 - A documentation typo fix

How can I test this
----

Just run `pytest`. 
You can check the test file, which should cover all the cases for this PR.

I would say, if you only have time for 1 thing, it would be to review the test named `test_it_imports_bboxes`, as it has the fully working example and it would be nice to have someone else verify my bbox calculations are OK.

Extra Info
----

Linear Parent ticket: https://linear.app/v7labs/issue/V7-2297/labelbox-format-importer
Linear Task ticket: https://linear.app/v7labs/issue/V7-2321/import-bounding-box-image


darwin/importer/formats/label_box.py
```diff
@@ -0,0 +1,111 @@
...
+
+    raise ValueError(f"Unsupported object type {obj}")
+
+
+def _to_bbox(obj: Dict[str, Any], title: str) -> Annotation:
+    bbox = obj.get("bbox")
+    assert bbox
+
+    x = bbox.get("top")
+    y = bbox.get("left")
```
andreaazzini, line 102:
```suggestion
    x = bbox.get("left")
    y = bbox.get("top")
```

darwin/importer/formats/label_box.py
```diff
@@ -0,0 +1,111 @@
...
+    width = bbox.get("width")
+    height = bbox.get("height")
+
+    return make_bounding_box(title, x, y, width, height)
```
andreaazzini, line 106: Does this work if `x`, `y`, `width` or `height` are `None`?

darwin/importer/formats/label_box.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def _get_classes(annotation: Annotation) -> AnnotationClass:
```
andreaazzini, line 109: This function returns a single class, so either the name should not be plural or something is not right here

andreaazzini (changes requested):
Some changes needed, but I like the overall structure.

What I'd do from here is:
1. add `labelbox` to the supported import formats
2. actually try it with `darwin dataset import <dataset-identifier> labelbox /path/to/labelbox.json`
3. see what happens

If it breaks, start with making this work (for each broken piece of functionality, having a test would be golden)
If it does NOT break, then YEY, let's move to polygons!

darwin/importer/formats/label_box.py
```diff
@@ -0,0 +1,111 @@
...
+
+    if obj.get("bbox"):
+        return _to_bbox(obj, title)
+
+    raise ValueError(f"Unsupported object type {obj}")
+
+
+def _to_bbox(obj: Dict[str, Any], title: str) -> Annotation:
+    bbox = obj.get("bbox")
+    assert bbox
```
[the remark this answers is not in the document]
Fl4m3Ph03n1x: This is mainly here to shut up pyright. 
You see, before calling this function, I already know I have a `"bbox"` because I only call this function in the event I have one. 

```
if obj.get("bbox"):
        return _to_bbox(obj, title)
```

However, Pyright is not smart enough to understand this so I had to add a quick "fix" to it.

darwin/importer/formats/label_box.py
Fl4m3Ph03n1x: Fixed!

darwin/importer/formats/label_box.py
Fl4m3Ph03n1x: No, it will absolutely blow up. 
But if we have a bbox without those fields, then it is a malformed one. 

I will add a check for it.

darwin/importer/formats/label_box.py
Fl4m3Ph03n1x: Fixed!

darwin/importer/formats/label_box.py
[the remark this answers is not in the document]
Fl4m3Ph03n1x: Fixed!

darwin/importer/formats/__init__.py
```diff
@@ -2,9 +2,10 @@
 
 from darwin.datatypes import ImporterFormat
 
-from . import coco, csvtags, csvtagsvideo, darwin, dataloop, pascalvoc
+from . import coco, csvtags, csvtagsvideo, darwin, dataloop, label_box, pascalvoc
```
andreaazzini, line 5: this is really really unimportant, but I think we should call that file `labelbox` (no underscore, like the actual startup name) for consistency with other files (`data loop`, `pascalvoc`, etc)

andreaazzini (changes requested):
One very minor change request, looks great otherwise!

darwin/importer/formats/__init__.py
Fl4m3Ph03n1x: done !

andreaazzini (approved):
LGTM ✨

PR #280, opened 2021-11-18
Fix bug with dataset split in darwin-py

Fl4m3Ph03n1x:
Background
----

A user reported in Github that `dataset.split` was not working. 
This PR aims to fix this issue. 

How can I test this
------

You can use this code snippet to run the code:

```

client = Client.local()
dataset = client.get_remote_dataset("cars")
dataset.pull()
dataset.split()
```

### DIFF
--- a/darwin/dataset/remote_dataset.py
+++ b/darwin/dataset/remote_dataset.py
@@ -646,8 +646,8 @@ class RemoteDataset:
 
     def split(
         self,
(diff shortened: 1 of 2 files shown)


FrancescoSaverioZuppichini:
LGTM but missing tests

andreaazzini (approved):

PR #282, opened 2021-11-19
tests for split datasets

FrancescoSaverioZuppichini:
### Background
One of the latest PRs created a fix for split dataset without testing and by wrongly assigning int type to percentage values

### How can I test this

```
python -m pytest ./tests/darwin/dataset/split_manager_test.py
```


darwin/dataset/remote_dataset.py
```diff
@@ -658,9 +658,9 @@ def split(
 
         Parameters
         ----------
-        val_percentage : int
+        val_percentage : float
             Percentage of images used in the validation set
-        test_percentage : int
+        test_percentage : float
```
Fl4m3Ph03n1x, line 663: If we do this (working with `float` instead of `int`) we will be breaking the Public API of several functions. 
The reason I changed everything to `int` its because the code inside was working with integers instead of floats. Not only that, the majority of functions used `int`, so I went with the change that was least disruptive.

I don't oppose the changes, but I would like @andreaazzini opinion, since this merits a 0.8.0 release IMO.

tests/darwin/dataset/split_manager_test.py
```diff
@@ -1,11 +1,37 @@
...
+    def it_should_split_a_dataset(test_datasets_dir: Path, val_percentage: float, test_percentage: float):
+        root = test_datasets_dir / "data" / "sl"
+
+        train_percentage: float = 1 - val_percentage - test_percentage
+
+        tot_size: int = len(list((root / "images").glob("*")))
+        splits: Path = split_dataset(
+            root, release_name="latest", val_percentage=val_percentage, test_percentage=test_percentage
+        )
+        print(splits)
```
Fl4m3Ph03n1x, line 28: Do we need this `print` in the test? Can't we remove it?

darwin/dataset/split_manager.py
```diff
@@ -269,6 +268,7 @@ def _stratify_samples(
     if test_percentage == 0.0:
         return list(set(X_train.astype(np.int))), list(set(X_tmp.astype(np.int))), []
 
+    print((test_percentage / (val_percentage + test_percentage)))
```
Fl4m3Ph03n1x, line 271: If we really want to print something, how feasible/easy would it be to use `rich` here?

darwin/torch/dataset.py
```diff
@@ -331,7 +331,9 @@ def get_target(self, index: int) -> Dict[str, Any]:
         annotations = []
         for obj in target["annotations"]:
             sequences = convert_polygons_to_sequences(
-                obj["polygon"]["path"], height=target["height"], width=target["width"],
+                obj["polygon"]["path"],
```
Fl4m3Ph03n1x, line 334: If "polygon" is None, this will explode right?
Do you think we could do some validation before? Would it be a silly idea?

Fl4m3Ph03n1x (changes requested):

tests/darwin/dataset/split_manager_test.py
FrancescoSaverioZuppichini: removed, thanks nice catch :)

darwin/dataset/split_manager.py
FrancescoSaverioZuppichini: removed, thanks nice catch :)

darwin/torch/dataset.py
FrancescoSaverioZuppichini: idk, I didn't touch that code

darwin/dataset/remote_dataset.py
FrancescoSaverioZuppichini: Then let's change them, float is the correct value to use here and IMHO it will make more sense for all our users.

I don't understand why everytime we need to be scared if we improve our APIs

Fl4m3Ph03n1x (approved):

darwin/dataset/remote_dataset.py
andreaazzini: Mmh I don't think I agree here guys. That's a type annotation, which as you know in Python is not going to change the way you use your API

PR #283, opened 2021-11-23
added JSONschema validation to labelbox

Fl4m3Ph03n1x:
Background
----

I have, for the first time in darwin-py's life, added JSON schema validation, this time to Labelbox. 
However, there are a couple of things I am not a big fan of:
- The error message for `oneOf` is simply wrong. It says `"point" is required` which is not true, `point` is only one of the possible valued (bbox and polygon being the others).
- After double checking, I have realized `jsonschema` is only compatible with Python 3.7 and onwards. I have to use an alternative. 

In the meantime I would like a review of the code. What do you guys think?

How can I test this?
----

Simply run `pytest`.

EDIT
----

Replaced `jsonschema` with `jsonschmea_rs` because the second one is the only one compatible with all the versions of Python `darwin-py` official supports. 

Although the issue with `oneOf` is now fixed (it does not exist in this library`, we have however another, perhaps more important issue:
- None of the missing required properties for internal objects belonging to `oneOf` are shown as missing. Instead we just get a general error "Could not match any of the schemas".

This is not really any better, one might even say the loss of granularity is so severe it can be even worst. However, it is literally the only one compatible with Python version 3.6 - 3.9.

We should probably discuss this.

### DIFF
--- a/darwin/importer/formats/labelbox.py
+++ b/darwin/importer/formats/labelbox.py
@@ -1,9 +1,9 @@
 import json
 from functools import partial
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Set
+from typing import Any, Dict, List, Optional, Set, cast
 
-from jsonschema import validate
+from jsonschema_rs import JSONSchema
 
 from darwin.datatypes import (
     Annotation,
@@ -15,7 +15,7 @@ from darwin.datatypes import (
 )
 
 
-def parse_file(path: Path, schema: Dict[str, Any]) -> Optional[List[AnnotationFile]]:
+def parse_file(path: Path, validator: JSONSchema) -> Optional[List[AnnotationFile]]:
     """
     Parses the given LabelBox file and maybe returns the corresponding annotations.
     The file must have a structure simillar to the following:
@@ -66,7 +66,7 @@ def parse_file(path: Path, schema: Dict[str, Any]) -> Optional[List[AnnotationFi
 
     with path.open() as f:
         data = json.load(f)
-        validate(instance=data, schema=schema)
+        validator.validate(data)
         convert_with_path = partial(_convert, path=path)
 
         return list(map(convert_with_path, data))
@@ -86,11 +86,8 @@ def _convert(file_data: Dict[str, Any], path) -> AnnotationFile:
     )
 
 
-# We ignore the Any | None warning here because the schema has been validated and we know
-# we have the values with the correct types
 def _convert_label_objects(obj: Dict[str, Any]) -> Annotation:
-    title: str = obj.get("title")  # type: ignore
-
+    title: str = str(obj.get("title"))
     bbox: Optional[Dict[str, Any]] = obj.get("bbox")
     if bbox:
         return _to_bbox_annotation(bbox, title)
@@ -102,13 +99,11 @@ def _convert_label_objects(obj: Dict[str, Any]) -> Annotation:
     raise ValueError(f"Unsupported object type {obj}")
 
 
-# We ignore the Any | None warning here because the schema has been validated and we know
-# we have the values with the correct types
 def _to_bbox_annotation(bbox: Dict[str, Any], title: str) -> Annotation:
-    x: float = bbox.get("left")  # type: ignore
-    y: float = bbox.get("top")  # type: ignore
-    width: float = bbox.get("width")  # type: ignore
-    height: float = bbox.get("height")  # type: ignore
+    x: float = cast(float, bbox.get("left"))
+    y: float = cast(float, bbox.get("top"))
+    width: float = cast(float, bbox.get("width"))
+    height: float = cast(float, bbox.get("height"))
 
     return make_bounding_box(title, x, y, width, height)
 
--- a/darwin/importer/formats/labelbox.py
+++ b/darwin/importer/formats/labelbox.py
@@ -1,9 +1,7 @@
 import json
 from functools import partial
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Set, cast
-
-from jsonschema_rs import JSONSchema
+from typing import Any, Callable, Dict, List, Optional, Set, cast
 
 from darwin.datatypes import (
     Annotation,
@@ -15,7 +13,7 @@ from darwin.datatypes import (
 )
 
 
-def parse_file(path: Path, validator: JSONSchema) -> Optional[List[AnnotationFile]]:
+def parse_file(path: Path, validate: Callable[[Any], None]) -> Optional[List[AnnotationFile]]:
     """
     Parses the given LabelBox file and maybe returns the corresponding annotations.
     The file must have a structure simillar to the following:
@@ -38,6 +36,8 @@ def parse_file(path: Path, validator: JSONSchema) -> Optional[List[AnnotationFil
     ]
     ```
 
+    You can check the Labelbox Schemas in `labelbox_schemas.py`.
+
     Currently we support the following annotations:
     - bounding-box `Image`: https://docs.labelbox.com/docs/bounding-box-json
     - polygon `Image`: https://docs.labelbox.com/docs/polygon-json
@@ -47,6 +47,9 @@ def parse_file(path: Path, validator: JSONSchema) -> Optional[List[AnnotationFil
     path: Path
         The path of the file to parse.
 
+    validate: Callable[[Any], None]
+        The validator function that validates the schema.
+
     Returns
     -------
     Optional[List[darwin.datatypes.AnnotationFile]]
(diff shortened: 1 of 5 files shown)


darwin/importer/formats/labelbox_schemas.py
```diff
@@ -0,0 +1,78 @@
...
+}
+
+
+label_object = {
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "type": "object",
+    "required": ["title"],
+    "properties": {"title": {"type": "string"}},
+    "oneOf": [
+        {"type": "object", "required": ["point"], "properties": {"point": point}},
```
kamilc, line 48: You don't need the `"type": "object"` here as it's already included in the level above.

darwin/importer/formats/labelbox_schemas.py
```diff
@@ -0,0 +1,78 @@
...
+    },
+    "required": ["Label", "External ID"],
+}
+
+labelbox_export = {
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://darwin.v7labs.com/schemas/export",
+    "description": "A Labelbox export",
+    "title": "Export",
+    "default": {"x": 1.2, "y": 2.5},
```
kamilc, line 73: Are those defaults meaningful? Does it convey any real logic here?

darwin/importer/formats/labelbox_schemas.py
```diff
@@ -0,0 +1,78 @@
...
+    "$id": "https://darwin.v7labs.com/schemas/bounding_box",
+    "description": "Schema of a Bounding Box",
+    "title": "Bounding Box",
+    "default": {"top": 1.2, "left": 2.5, "height": 10, "width": 20},
+    "examples": [{"top": 0, "left": 0, "height": 10, "width": 20}],
+    "type": "object",
+    "properties": {
+        "top": {"type": "number"},
+        "left": {"type": "number"},
+        "height": {"type": "number"},
```
kamilc, line 12: Could we add some constraints, like e. g.: `"minimum": 0`?

darwin/importer/formats/labelbox_schemas.py
```diff
@@ -0,0 +1,78 @@
...
+    "title": "Polygon",
+    "default": [{"x": 1.2, "y": 2.5}, {"x": 2.5, "y": 3.6}, {"x": 1.2, "y": 2.5}],
+    "examples": [[{"x": 1.2, "y": 2.5}, {"x": 2.5, "y": 3.6}, {"x": 1.2, "y": 2.5}], []],
+    "type": "array",
+    "items": point,
+}
+
+
+label_object = {
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
```
kamilc, line 43: Should it include `"$id"`?

darwin/importer/formats/labelbox.py
```diff
@@ -89,10 +86,10 @@ def _convert(file_data: Dict[str, Any], path) -> AnnotationFile:
     )
 
 
+# We ignore the Any | None warning here because the schema has been validated and we know
+# we have the values with the correct types
 def _convert_label_objects(obj: Dict[str, Any]) -> Annotation:
-    title: Optional[str] = obj.get("title")
-    if not title:
-        raise ValueError(f"LabelBox objects must have a title: {obj}")
+    title: str = obj.get("title")  # type: ignore
```
kamilc, line 92: Would it look nicer to do `str(obj.get("title"))` instead of ignoring the types?

darwin/importer/formats/labelbox.py
```diff
@@ -13,7 +15,7 @@
 )
 
 
-def parse_file(path: Path) -> Optional[List[AnnotationFile]]:
+def parse_file(path: Path, schema: Dict[str, Any]) -> Optional[List[AnnotationFile]]:
```
kamilc, line 18: I think that it now misses `schema` in the docstring.

kamilc (changes requested):

darwin/importer/formats/labelbox_schemas.py
Fl4m3Ph03n1x: I am specifying that the items of the array of of type "object". The above "object" type is for the father object.
Unless there is an implicit rule I am not aware of, I think this is correct.

darwin/importer/formats/labelbox_schemas.py
Fl4m3Ph03n1x: Those defaults are shown when documentation tools are used to pre-fill test fields.

darwin/importer/formats/labelbox_schemas.py
Fl4m3Ph03n1x: I don;t know if we have any constraints for `top` and `left`. I think the `minimum` one is a good start for `height` and `width` though!

darwin/importer/formats/labelbox_schemas.py
Fl4m3Ph03n1x: I am not sure all objects here should have `$id`. Reality is I subdivided this into so many objects because I the structure was quiet complex and big and this made it easier to understand. The only object I am actually reusing is `point`.

We can  have a talk about this if you want.

darwin/importer/formats/labelbox.py
Fl4m3Ph03n1x: Although that would work for `str(None)` it would not for the other cases. 
If you have any suggestions, I am all ears. The only ones I found were:
https://stackoverflow.com/questions/68446642/how-do-i-get-pylance-to-ignore-the-possibility-of-none

Any of them sound better then ignoring type casting, but I have not settled on one. @andreaazzini  for example is not a fan of `assert`s.

darwin/importer/formats/labelbox.py
Fl4m3Ph03n1x: Correct!

darwin/importer/formats/labelbox_schemas.py
kamilc: This isn't how it works. Whatever you specify under `oneOf` gets merged with the level above. So here the level above specifies `type: object` and alternatives under `oneOf` specify the same which result in... `type: object`. Let's take an example (pseudo-code for brevity):

```yaml
type: object
properties:
  title:
    type: string
oneOf:
  - properties:
      one:
        type: string
    required: one
  - properties:
      two:
        type: string
    required: two
```
Makes it resolve to these two alternative schemas:

```yaml
type: object
properties:
  title:
    type: string
  one:
    type: string
required: one
```
and:
```yaml
type: object
properties:
  title:
    type: string
  two:
    type: string
required: two
```

... which are tried one after another by the validator. Notice that adding `type: object` to children of `oneOf` wouldn't change the resulting alternative schemas at all.

Hope this helps.

darwin/importer/formats/labelbox_schemas.py
kamilc: We most probably don't have constraints for top and left but yes, width and height can't really be negative.

darwin/importer/formats/labelbox.py
kamilc: Isn't this `title` coming from the schema that has just been validated? If so, then we know for sure that it's a string-like variable. In this case, doing `str(...)` should be pretty safe. Am I not seeing something here?

darwin/importer/formats/labelbox_schemas.py
Fl4m3Ph03n1x: Fixed !

darwin/importer/formats/labelbox_schemas.py
Fl4m3Ph03n1x: Fixed !

darwin/importer/formats/labelbox.py
Fl4m3Ph03n1x: My comment is in regards to Pylance. Pylance doesnt know there was  a validation and therefore will always report an error. It is this I am trying to address here.

Fl4m3Ph03n1x: Fixed is static `cast`

darwin/importer/formats/labelbox.py
Fl4m3Ph03n1x: Waiting for final approval before fixing the docs :D

kamilc (approved):

darwin/importer/formats/labelbox.py
Fl4m3Ph03n1x: Fixed

kamilc (approved):

PR #284, opened 2021-11-24
Adding support for labelbox Point

Fl4m3Ph03n1x:
What changed
-----

Added support for labelbox Point. 
The schema was already here, so I didnt have to add it.

How can I test it?
----

with `pytest`, as a new test was added !


andreaazzini (approved):
LGTM!

PR #287, opened 2021-11-25
V7 2428 harmonize api for clientpy

Fl4m3Ph03n1x:
Background
-----

Implementation of changes from RFC:
https://www.notion.so/v7labs/API-Harmonization-for-client-py-0248a627bad144109bf64f078134cecf

This PR focuses on:
 - Usage of Python Logger for debugging instead of print
- GET PUT POST DELETE functions are now private
- Created separate functions for RAW
- No error handlers in all functions
- Using `reponse.ok` instead of checking for `response.status_code != 200`
- Fixed tests
- Checked code, no need to return tuples for the time being.

Missing:
- Add Docs for Sphinx


darwin/client.py
```diff
@@ -17,21 +20,43 @@
...
+            f"Request: (endpoint={endpoint})"
+        )
+
+        if response.status_code == 401:
+            raise Unauthorized()
+
+        if response.status_code == 404:
+            raise NotFound(urljoin(self.url, endpoint))
+
+        if response.status_code != 200 and retry:
```
simedw, line 51:
```suggestion
        if not response.ok and retry:
```
^--- I think this is what we want, since 201 is sometimes returned as well

darwin/client.py
```diff
@@ -17,21 +20,43 @@
...
+            f"({response.status_code}). "
+            f"Client: ({self})"
+            f"Request: (endpoint={endpoint})"
+        )
+
+        if response.status_code == 401:
+            raise Unauthorized()
+
+        if response.status_code == 404:
+            raise NotFound(urljoin(self.url, endpoint))
```
simedw, line 49: Do you think it makes sense to have one function that is called by put/get/post that checks the status_code and error messages and then raises the correct Exception, like NotFound or InsufficentStorage etc, so we don't need to duplicate that logic across the various methods?

darwin/client.py
```diff
@@ -17,21 +20,43 @@
     Unauthorized,
 )
 from darwin.utils import is_project_dir, urljoin
-from darwin.validators import name_taken, validation_error
 
 
 class Client:
-    def __init__(self, config: Config, default_team: Optional[str] = None):
+    def __init__(self, config: Config, log: Logger, default_team: Optional[str] = None):
```
FrancescoSaverioZuppichini, line 26: We should have a default logger defined somewhere, e.g. in `logger.py`

FrancescoSaverioZuppichini (commented):
See comments

darwin/client.py
Fl4m3Ph03n1x: I think it is an idea worth exploring.

darwin/client.py
Fl4m3Ph03n1x: And then we would import it using `from darwin.logger import log` or something simillar, right?
I also understand we can test this mocking the logger like we do with requests. 
Plus we would not change the Client's API even further. 

I like the idea, @andreaazzini what do you think?

darwin/client.py
Fl4m3Ph03n1x: Implemented

darwin/client.py
Fl4m3Ph03n1x: I agree with your suggestions, however the PR is already big enough and this change falls outside of the scope.
This PR's main objective was to introduce a rather simple usage of `logging` which it already does.

I created a Linear Ticket to work on this on the future:
https://linear.app/v7labs/issue/V7-2469/have-a-centralized-point-for-logger-configuration

darwin/client.py
```diff
@@ -1,282 +1,58 @@
...
-                raise InsufficientStorage()
-
-        if response.status_code != 200 and retry:
-            time.sleep(10)
-            return self.put(endpoint, payload=payload, retry=False)
-
-        if raw:
-            return response
+        if log is None:
+            self.log: Logger = logging.getLogger()
```
FrancescoSaverioZuppichini, line 38: better to do `logging.getLogger('darwin')`

FrancescoSaverioZuppichini (commented):
Minor comment, great work!

darwin/client.py
Fl4m3Ph03n1x: Done

darwin/dataset/upload_manager.py
```diff
@@ -18,6 +16,7 @@
 
 import requests
 from darwin.datatypes import PathLike
+from darwin.exceptions import UploadRequestError, UploadStage
```
andreaazzini, line 19: `UploadStage` is not an exception, to me it does not belong there. If we really want to move it, we could go for our types file

darwin/client.py
```diff
@@ -468,52 +414,274 @@ def get_datasets_dir(self, team: Optional[str] = None) -> Optional[str]:
...
-        from darwin import __version__
+        response: Dict[str, Any] = cast(
+            Dict[str, Any], self._get(f"/dataset_items/{dataset_item_id}/sign_upload", team=team_slug)
+        )
+        return response
 
-        headers["User-Agent"] = f"darwin-py/{__version__}"
-        return headers
+    def upload_data(
+        self, dataset_slug: str, payload: Dict[str, Any], team: Optional[str] = None
```
andreaazzini, line 478:
```suggestion
        self, dataset_slug: str, payload: Dict[str, Any], team_slug: Optional[str] = None
```

darwin/client.py
```diff
@@ -468,52 +414,274 @@ def get_datasets_dir(self, team: Optional[str] = None) -> Optional[str]:
...
+            of items to upload.
+        team: Optional[str]
+            Team slug of the team the dataset will belong to. Defaults to None.
+
+        Returns
+        ------
+        Optional[Dict[str, Any]]
+            A dictionary with the result of the operation, or None if the Team was not found.
+        """
+        the_team: Optional[Team] = self.config.get_team(team or self.default_team)
```
andreaazzini, line 498:
```suggestion
        team: Optional[Team] = self.config.get_team(team_slug or self.default_team)
```

darwin/client.py
```diff
@@ -468,52 +414,274 @@ def get_datasets_dir(self, team: Optional[str] = None) -> Optional[str]:
...
+
+        if not the_team:
+            return None
```
andreaazzini, line 501: I would honestly raise here, so:
- the return type of this function is not optional
- when we use `client.upload_data` we don't need to check for the result to be `None`

darwin/client.py
andreaazzini: You would have to propagate these name changes all over this function

darwin/dataset/upload_manager.py
```diff
@@ -206,11 +193,13 @@ def _request_upload(self) -> Tuple[List[ItemPayload], List[ItemPayload]]:
...
-                endpoint=f"/teams/{self.dataset_identifier.team_slug}/datasets/{self.dataset_identifier.dataset_slug}/data",
-                payload=upload_payload,
-                team=self.dataset_identifier.team_slug,
-            )
+            dataset_slug: str = self.dataset_identifier.dataset_slug
+            team_slug: Optional[str] = self.dataset_identifier.team_slug
+
+            data = self.client.upload_data(dataset_slug, upload_payload, team_slug)
+            if data is None:
+                raise ValueError(f"Unable to find team {team_slug} for dataset {dataset_slug}")
```
andreaazzini, line 201: No need to check here if you raise in `client.upload_data`

darwin/client.py
```diff
@@ -468,52 +414,274 @@ def get_datasets_dir(self, team: Optional[str] = None) -> Optional[str]:
...
+        the_team: Optional[Team] = self.config.get_team(team or self.default_team)
 
-        if team_config:
-            api_key = team_config.api_key
+        if not the_team:
+            return None
 
-        if api_key and len(api_key) > 0:
-            headers["Authorization"] = f"ApiKey {api_key}"
+        team_slug: str = the_team.slug
```
andreaazzini, line 470: suggestions for `upload_data` also apply here in `sign_upload`

darwin/dataset/upload_manager.py
```diff
@@ -239,12 +228,12 @@ def _upload_file(
...
-        team_slug = self.dataset_identifier.team_slug
+        team_slug: Optional[str] = self.dataset_identifier.team_slug
 
         try:
-            sign_response = self.client.get(f"/dataset_items/{dataset_item_id}/sign_upload", team=team_slug, raw=True)
-            sign_response.raise_for_status()
-            sign_response = sign_response.json()
+            sign_response: Optional[Dict[str, Any]] = self.client.sign_upload(dataset_item_id, team_slug)
+            if sign_response is None:
+                raise ValueError(f"Team not found: {team_slug}")
```
andreaazzini, line 236: No need to check here if you raise in `client.sign_upload`

tests/darwin/cli_functions_test.py
```diff
@@ -1,4 +1,5 @@
 import builtins
+import logging
```
andreaazzini, line 2: I trust you we need this, but why?

tests/darwin/client_test.py
```diff
@@ -1,88 +1,11 @@
...
 from pathlib import Path
-from typing import Any, Dict
 
-import pytest
 import responses
 
 from darwin.client import Client
 from darwin.config import Config
 from darwin.datatypes import Feature
-from darwin.exceptions import InsufficientStorage, Unauthorized
```
andreaazzini, line 10: We are removing some tests, and we're not adding any for the new functions we're introducing. Are we ok with this?

darwin/client.py
```diff
@@ -297,12 +73,19 @@ def list_remote_datasets(self, team: Optional[str] = None) -> Iterator[RemoteDat
...
+
         Returns
         -------
-        list[RemoteDataset]
-        List of all remote datasets
+        Iterator[RemoteDataset]
+            List of all remote datasets
         """
-        for dataset in self.get("/datasets/", team=team):
+        response: List[Dict[str, Any]] = cast(List[Dict[str, Any]], self._get("/datasets/", team=team))
```
andreaazzini, line 86: Could you explain why we need `cast` here? Looks a bit too verbose for my taste

darwin/client.py
```diff
@@ -386,19 +182,152 @@ def create_dataset(self, name: str, team: Optional[str] = None) -> RemoteDataset
...
+        Archive (soft delete) a remote dataset.
+
+        Parameters
+        ----------
+        dataset_id: int
+            Id of the dataset to archive.
+        team_slug: str
+            Team slug of the dataset.
+        """
+        self._put(f"datasets/{dataset_id}/archive", payload={}, team=team_slug)
```
andreaazzini, line 196: We're probably not using the response when we call `archive_remote_dataset`, but I would actually return the response for API consistency. What do you think?

andreaazzini (changes requested):
Requested a few changes already, but pretty much:
- in the new client functions, I would try and be consistent by always returning the `response`
- there are some naming conventions I personally don't like (`the_{}`)
- we are not really introducing any tests for `client.py`
- some things feel a bit weird, like using `cast` or importing `logging` in all tests. Are we sure there is no other way?

darwin/client.py
Fl4m3Ph03n1x: I am making generalized use of cast because I want to tell Pylance that I know the type of variable. Basically, since we have `Union[A, B]` the only way to make Pylance accept that a given variable is of type `A` is to use a cast. 

I am open to alternatives if you know of any. This is what my research found out.

tests/darwin/cli_functions_test.py
Fl4m3Ph03n1x: Old code. Fixed !

darwin/client.py
Fl4m3Ph03n1x: The documentation was incorrect for this case. This is the Team object, not the slug. Fixed it !

darwin/client.py
Fl4m3Ph03n1x: Done !

tests/darwin/client_test.py
Fl4m3Ph03n1x: Removed 2 outdated tests. Added more than that. Obviously every single function should be tested. But if I do that now it might be overkill for a single PR.

darwin/client.py
Fl4m3Ph03n1x: I would go even further and actually use a `DatasetIdentifier` iin every place we are using a slug or an `id`. However, I think I made enough breaking changes as is for 1 PR :D

darwin/dataset/upload_manager.py
Fl4m3Ph03n1x: Reverted changes

darwin/client.py
Fl4m3Ph03n1x: Solved over Zoom talk.

darwin/client.py
Fl4m3Ph03n1x: Ticket created for this topic:
https://linear.app/v7labs/issue/V7-2547/harmonization-part-2

PR #293, opened 2021-12-09
Added docs

Fl4m3Ph03n1x:


darwin/config.py
```diff
@@ -54,12 +64,19 @@ def get(self, key: Union[str, List[str]], default: Optional[Any] = None) -> Any:
...
+
+        Parameters
+        ----------
+        key: Union[str, List[str]]
+            The key where the value is going to be stored.
+            It can be formatted as a simple string, or as a path/like/string to fetch nested values.
+        value: Any
+            The value to be stored.
+        save: bool
+            If True, persists the value in the FileSystem. Defaults to 'True'.
```
andreaazzini, line 78: I think backticks are preferred?
```suggestion
            If True, persists the value in the FileSystem. Defaults to `True`.
```

darwin/config.py
```diff
@@ -80,21 +97,83 @@ def _save(self) -> None:
...
+        Returns the Team object from the team with the given slug if an authentication with an API 
+        key was performed earlier.
+        
+        Parameters
+        ----------
+        team: Optional[str]
+            The Team's slug. If none is given, searches for the default team instead. Defaults to 
+            'None'.
+        raise_on_invalid_team: bool
+            If True, raises if no team is found, if False returns 'None' instead. Defaults to True.
```
andreaazzini, line 164:
```suggestion
            If True, raises if no team is found, if False returns 'None' instead. Defaults to `True`.
```

darwin/config.py
```diff
@@ -80,21 +97,83 @@ def _save(self) -> None:
...
     def get_team(self, team: Optional[str] = None, raise_on_invalid_team: bool = True) -> Optional[Team]:
+        """
+        Returns the Team object from the team with the given slug if an authentication with an API 
+        key was performed earlier.
+        
+        Parameters
+        ----------
+        team: Optional[str]
+            The Team's slug. If none is given, searches for the default team instead. Defaults to 
+            'None'.
```
andreaazzini, line 162:
```suggestion
            `None`.
```

andreaazzini (approved):
Looks great, and I don't want to block it, but I think backticks are to be preferred from `'`

PR #297, opened 2021-12-10
fixing code blocks

Fl4m3Ph03n1x:
We cannot display code blocks like in markdown.

The parser translates them to a sort of Italic:
https://darwin-py-sdk.v7labs.com/darwin.html#module-darwin.config

While it is clearly not code, it is passable. We could use double backtick and go for monospace.
I don't like it, but I could live with it if only it didn't create messes like this:
https://darwin-py-sdk.v7labs.com/darwin.importer.formats.html#module-darwin.importer.formats.labelbox

This PR tries to fix that.

### DIFF
--- a/darwin/importer/formats/labelbox.py
+++ b/darwin/importer/formats/labelbox.py
@@ -1,5 +1,4 @@
 import json
-from dataclasses import dataclass
 from functools import partial, reduce
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


andreaazzini (approved):

FrancescoSaverioZuppichini (commented):
+1

Issue #305, opened 2021-12-20
Missed team -> team_slug change in torch get_dataset()


ck37:
It looks like #287 changed the API for client.list_local_datasets() to [expect team_slug rather than team](https://github.com/v7labs/darwin-py/blame/master/darwin/client.py#L42).

darwin/torch/dataset.py defines get_dataset() and it has a call to list_local_dataset() that still needs to be updated:
https://github.com/v7labs/darwin-py/blob/master/darwin/torch/dataset.py#L61

If possible it would be good to add a test for get_dataset() to detect this minor bug automatically.

Also it would be good to update the docstring for dataset_type in that same file (adding "object-detection"): https://github.com/v7labs/darwin-py/blob/master/darwin/torch/dataset.py#L37

andreaazzini:
Hi @ck37, thank you for reporting the issue.
We are on it, and should be able to release a fix soon! 😃 

Best!

Andrea

PR #306, opened 2021-12-20
Minor fixes for #305

ck37:
Hello,

This fixes two minor issues in darwin/torch/dataset.py noted in #305. I unfortunately don't know how to implement an automated test for get_dataset() but that would be ideal to add.

Thanks,
Chris


ck37:
Superseded by #307

PR #307, opened 2021-12-21
[HOTFIX] list_local_datasets accepts team_slug parameter

### LINKED ISSUE #305: Missed team -> team_slug change in torch get_dataset()

andreaazzini:
Fixes #305.
Supersedes #306.


darwin/client.py
```diff
@@ -150,7 +150,7 @@ def get_remote_dataset(self, dataset_identifier: Union[str, DatasetIdentifier])
                 client=self,
             )
         if not matching_datasets:
-            raise NotFound(parsed_dataset_identifier)
+            raise NotFound(str(parsed_dataset_identifier))
```
andreaazzini, line 153: MyPy rightly complained about this line, as `NotFound` accepts a `str`-typed parameter, and `parsed_dataset_identifier` is of type `DatasetIdentifier`

Fl4m3Ph03n1x:
This error was in the torch part, right ?
This worries me, as we introduced a regression while passing all the tests. This is imo an indicator that we should find a way to test the torch section of the project.

ck37:
Right, there is nothing in https://github.com/v7labs/darwin-py/blob/master/tests/darwin/torch/dataset.py that tests get_dataset()

Fl4m3Ph03n1x (approved):

PR #312, opened 2021-12-22
[V7-2617] datatset convert to respect export version

simedw:


darwin/cli_functions.py
```diff
@@ -830,22 +832,23 @@ def dataset_convert(dataset_slug: str, format: str, output_dir: Optional[PathLik
...
+    client: Client = _load_client(team_slug=identifier.team_slug)
     exporter_module = import_module(f"darwin.exporter.formats.{format}")
     try:
         parser: ExportParser = getattr(exporter_module, "export")
     except AttributeError:
         _error(f"Unsupported import format, currently supported: {export_formats}")
 
     try:
-        dataset: RemoteDataset = client.get_remote_dataset(dataset_identifier=dataset_slug)
+        dataset: RemoteDataset = client.get_remote_dataset(dataset_identifier=identifier.dataset_slug)
```
andreaazzini, line 844: Might be out of context, but here the api is confusing. It looks like `get_remote_dataset` needs `dataset_identifier` (which one would assume to be of `DatasetIdentifier` type). Instead `dataset_identifier` is actually a dataset slug. Can we change the nomenclature here?

(Disclaimer: I'm aware we're potentially breaking the API, but consistency is in my opinion more important in v0.x.x)

An alternative could be accepting either a `dataset_identifier: DatasetIdentifier` or a `dataset_slug: str`.

darwin/cli_functions.py
```diff
@@ -830,22 +832,23 @@ def dataset_convert(dataset_slug: str, format: str, output_dir: Optional[PathLik
...
         if not dataset.local_path.exists():
             _error(
                 f"No annotations downloaded for dataset f{dataset}, first pull a release using "
-                f"'darwin dataset pull {dataset_slug}'"
+                f"'darwin dataset pull {identifier}'"
```
andreaazzini, line 848: Will mypy complain here?
```suggestion
                f"'darwin dataset pull {str(identifier)}'"
```

andreaazzini (changes requested):
Two comments: the first one is included in my change requests; the second one is: can we add a test before we merge this? @simedw either Pedro or myself can write the test if you don't have the time for it.

darwin/cli_functions.py
simedw: you shouldn't need that, and identifier implements `__str__`

Fl4m3Ph03n1x:
I dont quite understand why this change is needed:
- The function takes a dataset_slug. So that parameter is called that.
- Also I am not sure if the documentation added wouldn't break our doc generation process as it is not numpydoc compatible.
- We are aware that in some places we use a DatasetIdentifier and in other a slug and in others something different. We have not yet had a discussion for this, but we noted the inconsistency. The reason I mention this is that such a change would break the Public API and so a proper RFC and discussion should be had to evaluate implications.
- As is, this change will still break the API.

Fl4m3Ph03n1x (approved):

andreaazzini (approved):

andreaazzini (approved):

PR #317, opened 2021-12-28
added superannotate types to class name

Fl4m3Ph03n1x:
Background
--------

With  SuperAnnotate you can have a class name being used for multiple types. 
Example: you have a cat class, and you can create polygon , bounding_box or ellipse annotations with it

This PR adds the SuperAnnotate type to the end of the class name, this way we will be able to support this feature because for Darwin these will be different Annotations: `cat-polygon, cat-bounding_box and cat-ellipse`

### DIFF
--- a/darwin/importer/formats/superannotate.py
+++ b/darwin/importer/formats/superannotate.py
@@ -159,7 +159,7 @@ def _to_keypoint_annotation(point: Dict[str, Any], classes: List[Dict[str, Any]]
     class_id: int = cast(int, point.get("classId"))
 
     name = _find_class_name(class_id, classes)
-    return make_keypoint(name + "-point", x, y)
+    return make_keypoint(f"{name}-point", x, y)
 
 
 def _to_bbox_annotation(bbox: Dict[str, Any], classes: List[Dict[str, Any]]) -> Annotation:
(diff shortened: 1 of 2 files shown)


darwin/importer/formats/superannotate.py
```diff
@@ -159,7 +159,7 @@ def _to_keypoint_annotation(point: Dict[str, Any], classes: List[Dict[str, Any]]
     class_id: int = cast(int, point.get("classId"))
 
     name = _find_class_name(class_id, classes)
-    return make_keypoint(name, x, y)
+    return make_keypoint(name + "-point", x, y)
```
andreaazzini, line 162: Rather than concatenation, I'd go for interpolation in these cases, as we do in other places in the codebase
```suggestion
    return make_keypoint(f"{name}-point", x, y)
```

andreaazzini (changes requested):
minor preference of interpolation over concatenation, unless you have a strong opinion yourself 🙂

darwin/importer/formats/superannotate.py
Fl4m3Ph03n1x: done !

andreaazzini (approved):

PR #319, opened 2021-12-29
V7 2629 support attributes

Fl4m3Ph03n1x:
Background
---------------

This PR adds attribute import support for superannotate attributes.

How can  I test?
-------------

Before testing there is one important detail: **The attributes will only be imported if they already exist**. Because for this condition to be true, the AnnotationClass must already exist and have the attributes defined, this means that if you are importing a new class, the class import will work, but it won't have any attributes.

A print message was added to give visibility to this issue.

**test.json**
```
{
    "instances": [
        {
            "type": "bbox",
            "classId": 2,
            "probability": 100,
            "points": {
                "x1": 275.4,
                "x2": 570.3,
                "y1": 590.8,
                "y2": 1440
            },
            "groupId": 0,
            "pointLabels": {},
            "locked": false,
            "visible": true,
            "attributes": [
                {
                    "id": 2,
                    "groupId": 1
                },
                {
                    "id": 3,
                    "groupId": 2
                }
            ]
        }
    ],
    "tags": [],
    "metadata": {
        "version": "1.0.0",
        "name": "sonnie-hiles-PUBt7UPbJFY-unsplash.jpg",
        "status": "Completed"
    }
}
```

**classes.json**
```
[
    {
        "attribute_groups": [
            {
                "id": 1,
                "name": "Sex",
                "is_multiselect": false,
                "attributes": [
                    {
                        "id": 1,
                        "name": "Male"
                    },
                    {
                        "id": 2,
                        "name": "Female"
                    }
                ],
                "opened": true,
                "hasChanges": false
            },
            {
                "id": 2,
                "name": "Emotion",
                "is_multiselect": false,
                "attributes": [
                    {
                        "id": 3,
                        "name": "Smiling"
                    },
                    {
                        "id": 4,
                        "name": "Sadness"
                    },
                    {
                        "id": 5,
                        "name": "Crying"
                    }
                ],
                "opened": true,
                "hasChanges": false
            }
        ],
        "color": "#fd7007",
        "id": 2,
        "name": "Pedestrian",
        "opened": true
    }
]
```

**command**:
`darwin dataset import cars superannotate YOUR_FOLDER`

### DIFF
--- a/darwin/importer/formats/superannotate.py
+++ b/darwin/importer/formats/superannotate.py
@@ -162,7 +162,11 @@ def _to_keypoint_annotation(point: Dict[str, Any], classes: List[Dict[str, Any]]
 
     instance_class: Dict[str, Any] = _find_class(class_id, classes)
     name: str = str(instance_class.get("name"))
-    return make_keypoint(f"{name}-point", x, y)
+    attributes: Optional[SubAnnotation] = _get_attributes(point, instance_class)
+    subannotations: Optional[List[SubAnnotation]] = None
+    if attributes:
+        subannotations = [attributes]
+    return make_keypoint(f"{name}-point", x, y, subannotations)
 
 
 def _to_bbox_annotation(bbox: Dict[str, Any], classes: List[Dict[str, Any]]) -> Annotation:
@@ -192,7 +196,12 @@ def _to_ellipse_annotation(ellipse: Dict[str, Any], classes: List[Dict[str, Any]
 
     instance_class: Dict[str, Any] = _find_class(class_id, classes)
     name: str = str(instance_class.get("name"))
-    return make_ellipse(f"{name}-ellipse", ellipse_data)
+    attributes: Optional[SubAnnotation] = _get_attributes(ellipse, instance_class)
+    subannotations: Optional[List[SubAnnotation]] = None
+    if attributes:
+        subannotations = [attributes]
+
+    return make_ellipse(f"{name}-ellipse", ellipse_data, subannotations)
 
 
 def _to_cuboid_annotation(cuboid: Dict[str, Any], classes: List[Dict[str, Any]]) -> Annotation:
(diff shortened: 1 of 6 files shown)


andreaazzini (changes requested):

darwin/importer/formats/superannotate.py
```diff
@@ -224,21 +236,54 @@ def _to_polygon_annotation(polygon: Dict[str, Any], classes: List[Dict[str, Any]
...
+                )
+                attribute: Optional[Dict[str, Union[str, int]]] = next(
+                    (attribute for attribute in group_attributes if attribute.get("id") == attribute_id), None
+                )
+
+                if attribute is None:
+                    raise ValueError(f"No attribute data found for {info}.")
+
+                final_attribute = f"{str(group.get('name'))}-{str(attribute.get('name'))}"
+                all_attributes.append(final_attribute)
```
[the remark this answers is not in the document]
Fl4m3Ph03n1x: Resolved all issues except the `str` one. I use it so Pylance does not complain. That is the only reason.

darwin/importer/formats/superannotate.py
```diff
@@ -208,37 +229,83 @@ def _to_cuboid_annotation(cuboid: Dict[str, Any], classes: List[Dict[str, Any]])
...
+                group_attributes: List[AttributeGroup] = cast(List[AttributeGroup], group.get("attributes"))
+                attribute: Optional[AttributeGroup] = next(
+                    (attribute for attribute in group_attributes if attribute.get("id") == attribute_id), None
+                )
+
+                if attribute is None:
+                    raise ValueError(f"No attribute data found for {info}.")
+
+                final_attribute: str = f"{str(group.get('name'))}:{str(attribute.get('name'))}"
+                all_attributes.append(final_attribute)
```
andreaazzini, line 303: My linter would complain we're over-nesting here, so I'd propose this instead
```suggestion
            if info_group_id != group_id:
                continue
                
            group_attributes: List[AttributeGroup] = cast(List[AttributeGroup], group.get("attributes"))
            attribute: Optional[AttributeGroup] = next(
                (attribute for attribute in group_attributes if attribute.get("id") == attribute_id), None
            )

            if attribute is None:
                raise ValueError(f"No attribute data found for {info}.")

            final_attribute: str = f"{str(group.get('name'))}:{str(attribute.get('name'))}"
            all_attributes.append(final_attribute)
```

andreaazzini (changes requested):
One tiny thing

darwin/importer/formats/superannotate.py
Fl4m3Ph03n1x: Your linter is likely right. We should probably use the same linters as I am surprised mine didn't complain. 
Anyway, suggested committed !

andreaazzini (approved):
LGTM now!

PR #325, opened 2022-01-11
Add comments to items in datasets 

simedw:
What ?
-------

A client of ours added this functionality a while ago to their version of darwin-py and then they got mad when they updated to the newest version and realized their changes to the cloned project were gone ( go figure ! )

This PR adds the functionality of adding a comment to a file via CLI and SDK.

How to test it?
----------

Pick a dataset and a file and use the following command:

(all the args passed with `--` are OPTIONAL, meaning you only have to pass the dataset slug, filename and the text_comment):

```
dataset comment {DATASET_SLUG} {FILENAME} --text {TEXT_COMMENT} --x {X} --y {Y} --w {WIDTH} --h {HEIGHT}
```

You should then see a comment box with your comment on the image.

### DIFF
--- a/darwin/cli_functions.py
+++ b/darwin/cli_functions.py
@@ -883,6 +883,38 @@ def convert(format: str, files: List[PathLike], output_dir: Path) -> None:
     export_annotations(parser, files, output_dir)
 
 
+# TODO: Move this elsewhere
+def post_workflow_comment(client, workflow_id, text, x=1, y=1, w=1, h=1):
+    client._post(
+        f"workflows/{workflow_id}/workflow_comment_threads",
+        {"bounding_box": {"x": x, "y": y, "w": w, "h": h}, "workflow_comments": [{"body": text}]},
+    )
+
+
+# TODO: move this elsewhere
+def instantitate_item(client, item_id):
+    result = client._post(f"dataset_items/{item_id}/workflow")
+    return result["current_workflow_id"]
+
+
+def post_comment(dataset_name, filename, text, x=1, y=1, w=1, h=1):
+    client = Client.local()
+    try:
+        dataset = client.get_remote_dataset(dataset_identifier=dataset_name)
+    except NotFound:
+        _error(f"unable to find dataset: {dataset_name}")
+
+    items = dataset.fetch_remote_files(filters={"filenames": [filename]})
+    items = list(items)
+    if len(items) == 0:
+        print(f"No files matching '{filename}' found")
+    for item in items:
+        workflow_id = item.current_workflow_id
+        if not workflow_id:
+            workflow_id = instantitate_item(dataset.client, item.id)
+        post_workflow_comment(dataset.client, workflow_id, text, x, y, w, h)
+
+
 def help(parser: argparse.ArgumentParser, subparser: Optional[str] = None) -> None:
     """
     Prints the help text for the given command.
--- a/darwin/cli_functions.py
+++ b/darwin/cli_functions.py
@@ -4,7 +4,7 @@ import datetime
 import os
 import sys
 from pathlib import Path
-from typing import Any, Dict, Iterator, List, NoReturn, Optional, Union
+from typing import Any, Dict, Iterator, List, NoReturn, Optional, Union, cast
 
 import humanize
 from rich.console import Console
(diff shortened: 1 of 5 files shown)


darwin/cli_functions.py
```diff
@@ -883,6 +881,59 @@ def convert(format: str, files: List[PathLike], output_dir: Path) -> None:
...
+        for item in items:
+            maybe_workflow_id: Optional[int] = item.current_workflow_id
+
+            if maybe_workflow_id is None:
+                workflow_id: int = client.instantitate_item(item.id)
+            else:
+                workflow_id = maybe_workflow_id
+
+            client.post_workflow_comment(workflow_id, text, x, y, w, h)
+            console.print("[bold green]Comment added successfully!")
```
andreaazzini, line 934: Rather than printing a line per item, I'd rather:
- add a rich progress bar for this loop
- print a final line, to recap how many comments were added successfully, and how many weren't (is that even possible?)

darwin/cli_functions.py
```diff
@@ -883,6 +881,59 @@ def convert(format: str, files: List[PathLike], output_dir: Path) -> None:
...
+    else:
+        for item in items:
+            maybe_workflow_id: Optional[int] = item.current_workflow_id
+
+            if maybe_workflow_id is None:
+                workflow_id: int = client.instantitate_item(item.id)
+            else:
+                workflow_id = maybe_workflow_id
+
+            client.post_workflow_comment(workflow_id, text, x, y, w, h)
```
andreaazzini, line 933: Can this be unsuccessful? If so, we could either:
- collect successful and unsuccessful operations, then print a final message
- simply print an error message for unsuccessful posts

darwin/client.py
```diff
@@ -714,6 +714,73 @@ def reset_item(self, dataset_slug: str, team_slug: str, payload: Dict[str, Any])
...
+            The width of the comment box.
+        h: float, default: 1
+            The height of the comment box.
+
+        Returns
+        -------
+        int
+            The id of the created comment.
+        """
+        response: Dict[str, Any] = cast(
```
andreaazzini, line 743: Just to be 100% sure here, why do we need `cast` here?

darwin/options.py
```diff
@@ -209,6 +208,28 @@ def __init__(self):
...
+        parser_comment.add_argument(
+            "--comment-y-position", required=False, type=float, default=1, help="Y coordinate for comment"
+        )
+        parser_comment.add_argument(
+            "--comment-width-px", required=False, type=float, default=1, help="Comment width in pixels"
+        )
+        parser_comment.add_argument(
+            "--comment-height-px", required=False, type=float, default=1, help="Comment height in pixels"
+        )
+
```
andreaazzini, line 232: What do you think of simplifying these arg names? After all, we already use `comment` as a command:

```suggestion
        parser_comment.add_argument("--text", type=str, help="Comment string")
        parser_comment.add_argument(
            "-x", type=float, default=1, help="X coordinate for comment"
        )
        parser_comment.add_argument(
            "-y", type=float, default=1, help="Y coordinate for comment"
        )
        parser_comment.add_argument(
            "-w", "--width", type=float, default=1, help="Comment width in pixels"
        )
        parser_comment.add_argument(
            "-h", "--height", type=float, default=1, help="Comment height in pixels"
        )

```

I also get why we'd like to keep `default` values set to `1`, but we should actually check that doing stuff like this works and the backend validates it correctly
```
darwin dataset comment --text "Will be placed at (1, 1), width 1px, height 1px"
```

andreaazzini (changes requested):

darwin/cli_functions.py
Fl4m3Ph03n1x: We will ever only comment on 1 item. So a progress bar does not fir this use case imo.

darwin/cli_functions.py
Fl4m3Ph03n1x: Yes, the `post_workflow_comment` can be unsuccessful, the HTTP call can fail (in which case we will raise with the error code, or the result may not be what we expect (in which case I will raise with `ValueError`). But as I mentioned, we will ever only execute this once (even though it is a for loop).

darwin/client.py
Fl4m3Ph03n1x: Because `_post` returns a Union type.

darwin/options.py
Fl4m3Ph03n1x: I am OK with simplifying names, but I am not sure what you mean by "works". I would not have asked for a review if I had not tested it :P

Fl4m3Ph03n1x: Simplified args as requested.

darwin/cli_functions.py
andreaazzini: Fair enough about that `for` point, but if the request in unsuccessful, we could catch the exception here and print a nice error message

darwin/options.py
andreaazzini: > I would not have asked for a review if I had not tested it

Of course I know and trust that, but some things may just slip through, just wanted to be 100% sure

Fl4m3Ph03n1x: :+1:

darwin/cli_functions.py
Fl4m3Ph03n1x: fixed

darwin/cli_functions.py
Fl4m3Ph03n1x: fixed

andreaazzini (approved):
LGTM

PR #327, opened 2022-01-20
Added docs to datatypes.py

Fl4m3Ph03n1x:


darwin/datatypes.py
```diff
@@ -1,6 +1,17 @@
...
+    Any,
+    Callable,
+    Dict,
+    Iterator,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    Union,
+    cast,
```
andreaazzini, line 13: who is using `cast` here?

Fl4m3Ph03n1x: no one is. Removed !

andreaazzini (approved):

PR #328, opened 2022-01-21
Added docs to item_sorter.py

Fl4m3Ph03n1x:


darwin/item_sorter.py
```diff
@@ -2,12 +2,49 @@
 from typing import Union
 
 
-class SortDirection(Enum):
-    ASCENDING = "asc"
-    DESCENDING = "desc"
+class DocEnum(Enum):
```
andreaazzini, line 5: I'd move this to a different file, so it can be reused elsewhere

andreaazzini (changes requested):
small change request, otherwise lgtm

andreaazzini (approved):

PR #329, opened 2022-01-24
Added docs to item.py

Fl4m3Ph03n1x:
NOTE
-----

I defend that, much like with `item_sorter.py`, the `parse_dataset_item` function should be refactored to inside the `DatasetItem` class in order to maintain consistency. 

However because this would break the public API of this module, I didn't do it.
Waiting for a second opinion on the matter.


darwin/item.py
```diff
@@ -6,6 +6,44 @@
 
 @dataclass(frozen=True, eq=True)
 class DatasetItem:
+    """
+    DatasetItems represent files that can be images or videos which belong to a dataset.
```
andreaazzini, line 10:
```suggestion
    DatasetItem represent files that can be images or videos which belong to a dataset.
```

darwin/item.py
```diff
@@ -6,6 +6,44 @@
...
+    Said dataset can be initiated or not.
```
andreaazzini, line 11: Not clear here what "initiated" means

darwin/item.py
```diff
@@ -6,6 +6,44 @@
...
+
+    Attributes
+    ----------
+    id : int
+        The id of this ``DatasetItem``.
+    filename : str
+        The filename of this ``DatasetItem``.
+    status : str
+        The filename of this ``DatasetItem``. It can be:
```
andreaazzini, line 20:
```suggestion
        The status of this ``DatasetItem``. It can be: 
```

andreaazzini (changes requested):

darwin/item.py
Fl4m3Ph03n1x: done!

darwin/item.py
Fl4m3Ph03n1x: I got confused with the second line. Took it away!

darwin/item.py
Fl4m3Ph03n1x: fixed!

andreaazzini (approved):

darwin/item.py
```diff
@@ -19,10 +59,101 @@ class DatasetItem:
...
+            filename,
+            status,
+            archived,
+            file_size,
+            dataset_id,
+            "n/a",
+            sequence,
+            workflow_id,
+            path,
+        )
```
andreaazzini, line 134: I personally don't like this `if not` validation logic for each item in the dictionary. Feels off from best practices tbh. Would something like this still work?

```suggestion
    @classmethod
    def parse(cls, **raw: Dict[str, Any]) -> "DatasetItem":
        """
        Parses the given dictionary into a ``DatasetItem``. Raises if such is not possible.

        Parameters
        ----------
        raw : Dict[str, Any]
            The dictionary to parse.

        Returns
        -------
        DatasetItem
            A dataset item with the parsed information.

        Raises
        ------
        ValueError
            If any of the keys from the given dictionary do not have the correct format.
        """

        return DatasetItem(**kwargs)
```

darwin/item.py
```diff
@@ -19,10 +59,101 @@ class DatasetItem:
...
+
+
+@deprecation.deprecated(
+    deprecated_in="0.7.5",
+    removed_in="0.7.7",
```
andreaazzini, line 139: Mmh, I think we should removing it in the next minor or major version number, not patch IMHO
```suggestion
    removed_in="0.8.0",
```

darwin/utils.py
```diff
@@ -574,3 +579,27 @@ def is_unix_like_os() -> bool:
...
+    -------
+    str
+        A string in the format of Major.Minor.Bugfix indicating the SDK's current version.
+
+    Raises
+    ------
+    ValueError
+        If the version cannot be retrieved.
+    """
+    with open(Path(__file__).parent.parent / "darwin" / "__init__.py", "r") as f:
```
andreaazzini, line 598: We should import the version number, rather than opening the text file and parse it

andreaazzini (changes requested):

darwin/item.py
Fl4m3Ph03n1x: I have never used kargs.. I will look into it.
I am also not  a fan of this form of validation, but the only other option I know is JSonSchema, which I think is totally overkill for this.

darwin/item.py
Fl4m3Ph03n1x: Agree, that was a placeholder :D

darwin/utils.py
Fl4m3Ph03n1x: As far as I know , we can't do that. That is why I open the file and parse it.

darwin/item.py
Fl4m3Ph03n1x: Doing it this way would not:
 - perform validation on the Parse function, which is something we should do not only because we do it in other parse functions but also because we are converting a potentially unwanted dictionary into an object we understand
 - force `datase_slug` to be "n/a", which is the case with the original function

I am afraid using the unpacking operator is not going to work for this case.

setup.py
```diff
@@ -3,14 +3,15 @@
...
-with open("README.md", "rb") as f:
-    long_description = f.read().decode("utf-8")
-
-with open(Path(__file__).parent / "darwin" / "__init__.py", "r") as f:
+with open(Path(__file__).parent / "darwin" / "version" / "__init__.py", "r") as f:
     content = f.read()
     # from https://www.py4u.net/discuss/139845
     version = re.search(r'__version__\s*=\s*[\'"]([^\'"]*)[\'"]', content).group(1)
 
+# print(f"VERSION IS: {darwin.version.__version__}")
```
andreaazzini, line 11: Let's get rid of this

setup.py
```diff
@@ -45,5 +51,5 @@
     ],
     entry_points={"console_scripts": ["darwin=darwin.cli:main"]},
     classifiers=["Programming Language :: Python :: 3", "License :: OSI Approved :: MIT License"],
-    python_requires=">=3.6",
+    python_requires=">=3.",
```
andreaazzini, line 54: I guess you meant
```suggestion
    python_requires=">=3.7",
```

andreaazzini (changes requested):

andreaazzini (approved):

PR #333, opened 2022-01-27
Adding docs to release.py

Fl4m3Ph03n1x:
Boundaries Issue
----------------

I have detected a boundaries issue here:

```python
with requests.get(self.url, stream=True) as response:
            with open(path, "wb") as download_file:
                shutil.copyfileobj(response.raw, download_file)
```

Technically speaking, any requests to the outside world should be in `client.py`. The reason for this has been widely explore in other places so I wont repeat myself here.

I have changed `releases.py` to use client and created  a new method called `fetch_binary`.


darwin/dataset/release.py
```diff
@@ -83,22 +184,26 @@ def download_zip(self, path: Path) -> Path:
         Returns
         --------
         Path
-            Same Path as provided in the parameters.
-        
+            Same ```Path``` as provided in the parameters.
```
kamilc, line 187: Should it use just 2 backticks?

<pre>
Same ``Path`` as provided in the parameters.
</pre>

kamilc: There are more of these which I won't mark not to clutter the review.

kamilc (changes requested):

kamilc (approved):

PR #335, opened 2022-01-31
Adding docs to download_manager.py

Fl4m3Ph03n1x:
Notes
--------

I am trimming the API of this file by trying to deprecated functions that are not used in any other part of the project. 
One of the things I am also trying to do, is to remove arguments that are not used at all. In this case `api_url` is my target. 

Deprecated the `api_url` parameter.


andreaazzini (approved):

darwin/dataset/download_manager.py
```diff
@@ -209,3 +266,28 @@ def download_image(url: str, path: Path, api_key: str) -> None:
...
             raise Exception(f"Timeout url request ({url}) after {TIMEOUT} seconds.")
         time.sleep(1)
+
+
+def _download_image(url: str, path: Path, api_key: str) -> None:
+    if path.exists():
+        return
+    TIMEOUT: int = 60
+    start: float = time.time()
+    while True:
```
kamilc, line 276: Does this mean that it will attempt 50~60 requests if the response is always HTTP 500?

Fl4m3Ph03n1x: Also, we should not have `requests.get` here, this HTTP request needs to go to `client.py`.
Good catch!
Any ideas for the new function?

PR #336, opened 2022-01-31
Added docs to identifier.py

Fl4m3Ph03n1x:


darwin/dataset/identifier.py
```diff
@@ -3,13 +3,53 @@
 
 
 class DatasetIdentifier:
+    """
+    Formal representation of a dataset identifier for the SDK.
```
andreaazzini, line 7: Here, I would add what a dataset identifier _is_.

Something along the lines of:
> A dataset identifier is a string that uniquely identifies a dataset on Darwin. A dataset identifier is made of the following substrings: `<team-slug>/<dataset-slug>:<version>`. If `version` is missing, that it defaults to `latest`

andreaazzini (changes requested):

darwin/dataset/identifier.py
Fl4m3Ph03n1x: Love the idea !

andreaazzini (approved):

PR #342, opened 2022-02-10
Feature/preserve extra text on coco import

hacmorgan:
We would like to add a feature to the COCO importing logic, to allow us to add a field to the import COCO that will be preserved in the output COCO.

Specifically, we were planning to use the `extra.text` field of an annotation to store an ID.

I would have liked to submit this in a working state, but have not been able to get the field to persist when exporting (from the web interface), so am hoping for a nudge in the right direction.

### DIFF
--- a/darwin/importer/formats/coco.py
+++ b/darwin/importer/formats/coco.py
@@ -27,6 +27,7 @@ def parse_json(path: Path, data: Dict[str, Any]) -> Iterator[dt.AnnotationFile]:
         image_id = annotation["image_id"]
         annotation["category_id"]
         annotation["segmentation"]
+
         if image_id not in image_annotations:
             image_annotations[image_id] = []
         image_annotations[image_id].append(parse_annotation(annotation, category_lookup_table))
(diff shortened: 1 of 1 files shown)


darwin/importer/formats/coco.py
```diff
@@ -48,12 +48,18 @@ def parse_annotation(annotation: Dict[str, Any], category_lookup_table: Dict[str
         print("Warning, unsupported RLE, skipping")
         return None
 
+    try:
+        extra_text = annotation["extra"]["text"]
```
andreaazzini, line 52: Is this `extra` field part of the original COCO format? If so, could you please send over a piece of documentation that would prove that?

andreaazzini (commented):
This PR proposes using an `extra` field to retain textual annotations. Until we know whether or not that's part of the official COCO format, we prefer not merging this one in.

owencjones:
Closing this as stale for now.

@hacmorgan Thanks for your suggestion, if you can point us to a COCO format doc or schema that details what we can put in this field, happy to reopen and consider for merge.

If this is blocking your work, or would assist your work significantly, please feel free to reach out again, too.  Product feature suggestions, and PRs always welcome from users.

PR #350, opened 2022-02-16
V7 2246 darwin/exporter/formats/coco.py

Fl4m3Ph03n1x:
NOTE
------

I have deprecated most of the file, because most of it were simply private functions in reality. Not being used anywhere else. 
Reduction of the public API is something I usually aim for when possible.

### DIFF
--- a/darwin/exporter/formats/coco.py
+++ b/darwin/exporter/formats/coco.py
@@ -12,6 +12,10 @@ from darwin.exporter.formats.numpy_encoder import NumpyEncoder
 from darwin.utils import convert_polygons_to_sequences
 from darwin.version import __version__
 
+DEPRECATION_MESSAGE = """This function is going to be turned into private. This means that breaking 
+changes in its interface and implementation are to be expected. We encourage using ``export`` 
+instead of calling this low-level function directly."""
+
 
 def export(annotation_files: Iterator[dt.AnnotationFile], output_dir: Path) -> None:
     """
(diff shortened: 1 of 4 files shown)


darwin/exporter/formats/coco.py
```diff
@@ -3,33 +3,39 @@
...
     output_file_path = (output_dir / "output").with_suffix(".json")
     with open(output_file_path, "w") as f:
         json.dump(output, f, cls=NumpyEncoder, indent=1)
 
 
+@deprecation.deprecated(
+    deprecated_in="0.7.7",
+    removed_in="0.8.0",
+    current_version=__version__,
+    details="Use the ``export`` instead.",
```
andreaazzini, line 37: I'm not sure this `details` string is very well descriptive of the reason why we're deprecating these functions. The following is an attempt at making it a bit better, but lmk what you think
```suggestion
    details="This function is going to be turned into private. This means that breaking changes in its interface and implementation are to be expected. We encourage using ``export`` instead of calling this low-level function directly.",
```

andreaazzini (changes requested):

darwin/exporter/formats/coco.py
Fl4m3Ph03n1x: I like the idea!
I was following the convention the library suggests in the docs, but I think your description makes more sense!

darwin/exporter/formats/coco.py
```diff
@@ -34,7 +34,9 @@ def export(annotation_files: Iterator[dt.AnnotationFile], output_dir: Path) -> N
     deprecated_in="0.7.7",
     removed_in="0.8.0",
     current_version=__version__,
-    details="Use the ``export`` instead.",
+    details="""This function is going to be turned into private. This means that breaking changes 
+    in its interface and implementation are to be expected. We encourage using ``export`` instead 
+    of calling this low-level function directly.""",
```
andreaazzini, line 39: Sorry for being a bit nitpicky, but couldn't we define this string once, so we can reuse it in all these places?

andreaazzini (changes requested):

andreaazzini (approved):

PR #353, opened 2022-02-21
[V7-2915] PascalVoc convertion failed for video

simedw:


andreaazzini (changes requested):
`output_dir` is an argument of this function. Can that folder be possible created earlier than when we hit this `export_file` function?

Also, can we write a test that shows this PR is fixing the issue?

simedw:
> `output_dir` is an argument of this function. Can that folder be possible created earlier than when we hit this `export_file` function?
> 
> Also, can we write a test that shows this PR is fixing the issue?

@andreaazzini. We can't do that, since we are creating a folder inside the output folders for each video.
If you have `test.mp4` and output folders is `/tmp/output` we need to create `/tmp/output/test/` to store `000001.xml`, `000002.xml` and so onwards.

andreaazzini (approved):
@simedw ah, right, that makes sense!
@Fl4m3Ph03n1x thanks for adding the test!

Issue #357, opened 2022-02-23
fetch_remote_files not working for spaces (or maybe commas) in filenames


jbencook:
I'm pretty sure there's an issue with the `fetch_remote_files()` method on the `RemoteDataset` class (or more more likely in the actual API call) with either spaces or commas inside filenames.

To reproduce: upload two files to a dataset with filenames `Video Sep 09, 10 02 59 AM.mov` and `5897.mp4` respectively

`dataset.fetch_remote_files({"filenames": ["5897.mp4"]}))` returns results whereas `dataset.fetch_remote_files({"filenames": ["Video Sep 09, 10 02 59 AM.mov"]}))` does not.

Fl4m3Ph03n1x:
Hello @jbencook and thank you for reporting.

Darwin-py currently parses file names using `,` as a separator. Therefore we don't allow `, ` to be in filenames. 
We understand this information is not visible in a user friendly manner and will work on the documentation to make it clear.

Best regards, 
Wind Team, V7

Fl4m3Ph03n1x:
Hello @jbencook , after careful consideration and some internal discussions, we decided to change this unintuitive behaviour. 
As of the last release, v 0.7.10, we now allow files with commas. 

https://github.com/v7labs/darwin-py/releases/tag/v0.7.10

I hope you enjoy the new version as much as we enjoyed upgrading it!
Best regards, 
Wind Team, V7

jbencook:
That's great news -- thank you!

PR #359, opened 2022-02-24
fixing failure to fetch files with commas

Fl4m3Ph03n1x:
Addressing issue with: https://github.com/v7labs/darwin-py/issues/357

### DIFF
--- a/tests/darwin/dataset/remote_dataset_test.py
+++ b/tests/darwin/dataset/remote_dataset_test.py
@@ -393,8 +393,7 @@ def describe_fetch_remote_files():
             status=200,
         )
 
-        actual = remote_dataset.fetch_remote_files({"filenames": ["example,with, comma.mp4"]})
-        list(actual)
+        list(remote_dataset.fetch_remote_files({"filenames": ["example,with, comma.mp4"]}))
 
         request_body = json.loads(responses.calls[0].request.body)
 
--- a/tests/darwin/dataset/remote_dataset_test.py
+++ b/tests/darwin/dataset/remote_dataset_test.py
@@ -378,6 +378,28 @@ def describe_fetch_remote_files():
         assert item_1.id == 386074
         assert item_2.id == 386073
 
+    @responses.activate
+    def it_fetches_files_with_commas(
+        darwin_client: Client, dataset_name: str, dataset_slug: str, team_slug: str, files_content: dict
+    ):
+        remote_dataset = RemoteDataset(
+            client=darwin_client, team=team_slug, name=dataset_name, slug=dataset_slug, dataset_id=1
+        )
+        url = "http://localhost/api/datasets/1/items?page%5Bsize%5D=500"
+        responses.add(
+            responses.POST,
+            url,
+            json=files_content,
+            status=200,
+        )
+
+        actual = remote_dataset.fetch_remote_files({"filenames": ["Video Sep 09, 10 02 59 AM.mov"]})
+        list(actual)
+
+        request_body = json.loads(responses.calls[0].request.body)
+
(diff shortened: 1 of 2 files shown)


tests/darwin/dataset/remote_dataset_test.py
```diff
@@ -378,6 +378,28 @@ def it_works(darwin_client: Client, dataset_name: str, dataset_slug: str, team_s
...
+        url = "http://localhost/api/datasets/1/items?page%5Bsize%5D=500"
+        responses.add(
+            responses.POST,
+            url,
+            json=files_content,
+            status=200,
+        )
+
+        actual = remote_dataset.fetch_remote_files({"filenames": ["Video Sep 09, 10 02 59 AM.mov"]})
+        list(actual)
```
tomvars, line 397: What's the purpose of line 397?

Fl4m3Ph03n1x: To consume the generator so I can then use it to make the asserts.

andreaazzini: I don't necessarily love the suggestions in the following thread, but perhaps they look a bit nicer?
https://stackoverflow.com/questions/47456631/simpler-way-to-run-a-generator-function-without-caring-about-items

tests/darwin/dataset/remote_dataset_test.py
```diff
@@ -378,6 +378,28 @@ def it_works(darwin_client: Client, dataset_name: str, dataset_slug: str, team_s
...
+        url = "http://localhost/api/datasets/1/items?page%5Bsize%5D=500"
+        responses.add(
+            responses.POST,
+            url,
+            json=files_content,
+            status=200,
+        )
+
+        actual = remote_dataset.fetch_remote_files({"filenames": ["example,with, comma.mp4"]})
+        list(actual)
```
andreaazzini, line 397: and we don't need `actual`, so we should be fine with this anyway
```suggestion
        list(remote_dataset.fetch_remote_files({"filenames": ["example,with, comma.mp4"]}))
```

andreaazzini (approved):
Functionally speaking, it works well.

Tested manually
```python
from collections import deque
from darwin import Client
from pathlib import Path

c = Client.from_config(Path("/Users/andrea/.darwin/config.yaml"))
dataset = c.get_remote_dataset("v7-labs/test-comma")

# outputs `deque([])`
deque(dataset.fetch_remote_files({"filenames": "1,2.png"}))

# outputs `deque([DatasetItem(id=337857499, filename='1,2.png', status='processing', archived=False, filesize=None, dataset_id=401344, dataset_slug='n/a', seq=1, current_workflow_id=None, path='/')])`
deque(dataset.fetch_remote_files({"filenames": ["1,2.png"]}))
```

I'd say let's change that test line and merge!

andreaazzini:
@Fl4m3Ph03n1x when it's merged and released, please remember to message that user who raised the issue, and tell him the good news

tests/darwin/dataset/remote_dataset_test.py
Fl4m3Ph03n1x: I somehow fail to see how defining an additional function that uses an external library to exaust the generator will make it more readable/mantainable:

```
from collections import deque

def exhaust(generator):
    deque(generator, maxlen=0)
```

Other suggestions are OKish:

```def exhaust(generator): all(generator)``` 
(assuming the generator returns `True`)

But I am still not convinced. 

I could do the same:

```def exhaust(generator): list(generator)``` 

Which would work for all generators, irrespective of their return values. 

I am not sold on the idea, please comment here until tomorrow so I know if I should merge or not.

PR #360, opened 2022-02-24
Addes --yes flag to skip prompts when importing

simedw:
Allows for 
```sh
darwin dataset import --yes my-dataset darwin.json
```

And 
```python
 import_annotations(dataset, parser, files, append, class_prompt=False)
```

to skip the `secure_continue_request` prompts.

### DIFF
--- a/darwin/cli.py
+++ b/darwin/cli.py
@@ -112,7 +112,7 @@ def _run(args: Namespace, parser: ArgumentParser) -> None:
         elif args.action == "pull":
             f.pull_dataset(args.dataset, args.only_annotations, args.folders, args.video_frames)
         elif args.action == "import":
-            f.dataset_import(args.dataset, args.format, args.files, args.append)
+            f.dataset_import(args.dataset, args.format, args.files, args.append, args.yes)
         elif args.action == "convert":
             f.dataset_convert(args.dataset, args.format, args.output_dir)
         elif args.action == "set-file-status":
@@ -125,7 +125,13 @@ def _run(args: Namespace, parser: ArgumentParser) -> None:
             f.help(parser, "dataset")
         elif args.action == "comment":
             f.post_comment(
-                args.dataset, args.file, args.text, args.x, args.y, args.w, args.h,
+                args.dataset,
+                args.file,
+                args.text,
+                args.x,
+                args.y,
+                args.w,
+                args.h,
             )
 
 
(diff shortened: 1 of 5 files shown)


andreaazzini (changes requested):
It works, just tested it!

Blocking this because tests are missing, @simedw is this urgent to release?

Fl4m3Ph03n1x:
Wait, shouldn't the default behavior be: "We prompt the user?"

I just imported some annotations (the classes were deleted beforehand) and I was not asked for confirmation:

```
$ darwin dataset import cars darwin /home/pedro/Downloads/annotations/
Fetching remote class list...
Retrieving local annotations ...
Fetching remote file list...
3 annotation file(s) found.
3 classes needs to be created.
0 classes needs to be added to andreas-team/cars
About to create the following classes
        car_bbox, type: bounding_box
        car, type: polygon
        car_ellipse, type: ellipse
Working... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00
Working... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00
Working... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00
```

darwin/cli.py
```diff
@@ -112,7 +112,7 @@ def _run(args: Namespace, parser: ArgumentParser) -> None:
         elif args.action == "pull":
             f.pull_dataset(args.dataset, args.only_annotations, args.folders, args.video_frames)
         elif args.action == "import":
-            f.dataset_import(args.dataset, args.format, args.files, args.append)
+            f.dataset_import(args.dataset, args.format, args.files, args.append, args.yes)
```
andreaazzini, line 115: Blocked for another reason:
```suggestion
            f.dataset_import(args.dataset, args.format, args.files, args.append, not args.yes)
```

In the other functions, we're using `class_prompt` as a flag that is `True` when `--yes` is `False`!

Luckily caught that today, when using this branch to upload data with classes which did not exist! Yesterday everything went fine as classes existed in the team, but not in the dataset, and were automatically added

andreaazzini (changes requested):

andreaazzini (approved):

PR #361, opened 2022-02-24
Support yolo export for bounding boxes

Fl4m3Ph03n1x:
What?
-------

This is the first draft of the YOLO exporter. 
Right now I only support bounding boxes.

For more on the YOLO format please see:
https://towardsdatascience.com/image-data-labelling-and-annotation-everything-you-need-to-know-86ede6c684b1

Todo (after this PR)
------

- Support polygons
- Support Ellipses
- Support Complex polygons
- Support other shapes I am not remembering right now
- Test console error messages

I have taken the decision to **NOT** support `VideoAnnotation`s. I believe this to be reasonable, let me know if you think otherwise. 

What I am looking for if you review
-------------------------------

- Is the code easy to read? (are the functions too big?)
- What do you think of the new `datatypes` I added? I am trying to make feedback for users as friendly as possible. In my head, nothing beat a nice Rich message with good information.
- Am I missing any test cases here? (Failure to create folders or permissions denied on files comes to mind, but I am not sure this is a case I should consider covering to begin with).

Assumptions
-------------

This code is written understand the assumption that the `AnnotationFile`s I will be receiving have coherent, well formed data, since this data is being generated from within our app. Let me know if this assumption could be wrong.

Other things you may see relevant, please share. 
I will not move on with another PR until we are all happy and agree this PR is a solid base to built upon.

### DIFF
--- a/darwin/datatypes.py
+++ b/darwin/datatypes.py
@@ -284,6 +284,53 @@ class AnnotationFile:
         return construct_full_path(self.remote_path, self.filename)
 
 
+@dataclass(frozen=True, eq=True)
+class ConversionError:
+    """
+    Represents a failed conversion to YOLO format.
+
+    Attributes
+    ----------
+    reason : str
+        The reason on why the conversion failed.
+    annotation : Union[Annotation, VideoAnnotation]
+        The annotation that failed to be converted.
+    filename : Path
+        The path of the ``AnnotationFile`` holding the failed ``Annotation``.
+    """
+
+    reason: str
+    annotation: Union[Annotation, VideoAnnotation]
+    filename: Path
+
+
+@dataclass(frozen=True, eq=True)
+class YoloAnnotation:
+    """
+    Represents a YOLO annotation ready to be persisted as a bounding box.
+    The XY coordinates represent the top left corner of said bounding box.
+
+    Attributes
+    ----------
+    annotation_class : str
+        The name of the ``AnnotationClass``.
+    x : float
+        Left X coordinate of the bounding box.
+    y : float
+        Top Y coordinate of the bounding box.
+    width : float
+        Width of the bounding box.
+    height : float
+        Height of the bounding box.
+    """
+
+    annotation_class: str
+    x: float
+    y: float
+    width: float
+    height: float
+
+
 def make_bounding_box(
     class_name: str, x: float, y: float, w: float, h: float, subs: Optional[List[SubAnnotation]] = None
 ) -> Annotation:
@@ -383,7 +430,7 @@ def make_complex_polygon(
     subs: Optional[List[SubAnnotation]] = None,
 ) -> Annotation:
     """
-    Creates and returns a conplex polygon annotation. Complex polygons are those who have holes 
+    Creates and returns a conplex polygon annotation. Complex polygons are those who have holes
     and/or disform shapes.
 
     Parameters
@@ -391,8 +438,8 @@ def make_complex_polygon(
     class_name: str
         The name of the class for this ``Annotation``.
     point_paths: List[List[Point]]
-        A list of lists points that comprises the complex polygon. This is needed as a complex 
-        polygon can be effectively seen as a sum of multiple simple polygons. The list should have 
+        A list of lists points that comprises the complex polygon. This is needed as a complex
+        polygon can be effectively seen as a sum of multiple simple polygons. The list should have
         a format simillar to:
 
         .. code-block:: python
@@ -515,7 +562,7 @@ def make_ellipse(class_name: str, parameters: EllipseData, subs: Optional[List[S
     class_name: str
         The name of the class for this ``Annotation``.
     parameters: EllipseData
-        The data needed to build an Ellipse. This data must be a dictionary with a format simillar 
+        The data needed to build an Ellipse. This data must be a dictionary with a format simillar
         to:
 
         .. code-block:: javascript
@@ -532,10 +579,10 @@ def make_ellipse(class_name: str, parameters: EllipseData, subs: Optional[List[S
             }
 
         Where:
-        
+
         - ``angle: float`` is the orientation angle of the ellipse.
         - ``center: Point`` is the center point of the ellipse.
-        - ``radius: Point`` is the width and height of the elipse, where ``x`` represents the width 
+        - ``radius: Point`` is the width and height of the elipse, where ``x`` represents the width
         and ``y`` represents height.
     subs: Optional[List[SubAnnotation]], default: None
         List of ``SubAnnotation``s for this ``Annotation``. Defaults to ``None``.
@@ -543,7 +590,7 @@ def make_ellipse(class_name: str, parameters: EllipseData, subs: Optional[List[S
     Returns
     -------
     Annotation
-        An ellipse ``Annotation``. 
+        An ellipse ``Annotation``.
     """
     return Annotation(AnnotationClass(class_name, "ellipse"), parameters, subs or [])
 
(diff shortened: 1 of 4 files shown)


darwin/datatypes.py
```diff
@@ -284,6 +284,53 @@ def full_path(self) -> str:
         return construct_full_path(self.remote_path, self.filename)
 
 
+@dataclass(frozen=True, eq=True)
+class ConversionError:
+    """
+    Represents a failed conversion to YOLO format.
```
andreaazzini, line 290: While I like `ConversionError`, I feel like it would be useful not just for YOLO. Not suggesting that we should use it in all export modules in this PR though.

Devil's advocate: do we _need_ it at all? We could return `None` if the conversions fail, and instead of keeping conversion errors in memory, use a file logger to keep track of all the errors. 🤔

darwin/datatypes.py
```diff
@@ -284,6 +284,53 @@ def full_path(self) -> str:
...
+        The path of the ``AnnotationFile`` holding the failed ``Annotation``.
+    """
+
+    reason: str
+    annotation: Union[Annotation, VideoAnnotation]
+    filename: Path
+
+
+@dataclass(frozen=True, eq=True)
+class YoloAnnotation:
```
andreaazzini, line 308: This is a "type" that is only useful for the YOLO exporting logic, and as such I would keep it there.

darwin/exporter/formats/yolo.py
```diff
@@ -0,0 +1,90 @@
...
+    console = Console(theme=_console_theme())
+
+    output_dir.mkdir(parents=True, exist_ok=True)
+    for annotation_file in annotation_files:
+        output_file_path = (output_dir / annotation_file.filename).with_suffix(".txt")
+        errors, yolo_annotations = _convert_file(annotation_file)
+
+        with open(output_file_path, "w") as f:
+            for yolo in yolo_annotations:
+                f.write(f"{yolo.annotation_class} {yolo.x} {yolo.y} {yolo.width} {yolo.height}")
```
andreaazzini, line 37: I'm sure this is the right format, but I think we could include something about the format in this module's documentation

darwin/exporter/formats/yolo.py
```diff
@@ -0,0 +1,90 @@
...
+        The ``AnnotationFile``s to be exported.
+    output_dir : Path
+        The folder where the new coco file will be.
+    """
+    console = Console(theme=_console_theme())
+
+    output_dir.mkdir(parents=True, exist_ok=True)
+    for annotation_file in annotation_files:
+        output_file_path = (output_dir / annotation_file.filename).with_suffix(".txt")
+        errors, yolo_annotations = _convert_file(annotation_file)
```
andreaazzini, line 33: So if a file cannot be converted, we're not blocking the execution. This obviously has some useful properties, but I'm not sure it's consistent with the other exporters.

darwin/exporter/formats/yolo.py
```diff
@@ -0,0 +1,90 @@
...
+                f"Failed to convert: '{err.filename}'. Reason: {err.reason}. Annotation: {err.annotation}\n",
+                style="error",
+            )
+
+
+def _convert_file(file: AnnotationFile) -> Tuple[List[ConversionError], List[YoloAnnotation]]:
+    conversions: List[Union[ConversionError, YoloAnnotation]] = _to_yolo(file)
+    yolo_annotations: List[YoloAnnotation] = [x for x in conversions if isinstance(x, YoloAnnotation)]
+    errors: List[ConversionError] = [x for x in conversions if isinstance(x, ConversionError)]
+    return (errors, yolo_annotations)
```
andreaazzini, line 50:
```suggestion
    return errors, yolo_annotations
```

darwin/exporter/formats/yolo.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def _to_yolo(annotation_file: AnnotationFile) -> List[Union[ConversionError, YoloAnnotation]]:
+    create_yolo_annotation_with_file = partial(_create_yolo_annotation, annotation_file=annotation_file)
+    return _map_list(create_yolo_annotation_with_file, annotation_file.annotations)
```
andreaazzini, line 55: Looks a bit convoluted to me. Would something like this work instead?
```suggestion
    return [_create_yolo_annotation(annotation) for annotation in annotation_file.annotations]
```

darwin/exporter/formats/yolo.py
```diff
@@ -0,0 +1,90 @@
...
+    name: str = annotation.annotation_class.name
+    bbox: Dict[str, float] = annotation.data.get("bounding_box", {})
+    return YoloAnnotation(annotation_class=name, x=bbox["x"], y=bbox["y"], width=bbox["w"], height=bbox["h"])
+
+
+def _map_list(fun: Callable[[Any], Any], the_list: List[Any]) -> List[Any]:
+    return list(map(fun, the_list))
+
+
+def _console_theme() -> Theme:
```
andreaazzini, line 89: Rich related things don't belong to the exporter module to me, but the CLI's instead

andreaazzini (changes requested):

darwin/datatypes.py
Fl4m3Ph03n1x: > Devil's advocate: do we need it at all? We could return None if the conversions fail, and instead of keeping conversion errors in memory, use a file logger to keep track of all the errors. thinking

I thought about that (discarding error info) but went against it after talking to Rafal Z. It turns out users really like to know _why_ something fails. An error in conversion is something potentially useful for a user to see after an operation completes.

If you still want a logger, we can still use this data structure for such a thing. I personally like the idea.

> While I like ConversionError, I feel like it would be useful not just for YOLO. Not suggesting that we should use it in all export modules in this PR though.

This is the direction I would like to move. Testing waters here, lets see how this PR/functionality goes and then draw conclusions / improvements :D

darwin/datatypes.py
Fl4m3Ph03n1x: done

darwin/exporter/formats/yolo.py
Fl4m3Ph03n1x: Done

darwin/exporter/formats/yolo.py
Fl4m3Ph03n1x: Currently, there is no consistent behaviour. Some exports, like coco, have a full dataset inside, while other, have an export file for each image. I could with the "All or Nothing" approach here, but I think it is better for the customers to have something working, than nothing. At least this is the idea I have from Rafal Z and James.

darwin/exporter/formats/yolo.py
Fl4m3Ph03n1x: done

darwin/exporter/formats/yolo.py
Fl4m3Ph03n1x: So, here is the thing, you are using list comprehensions to do a `map`. Basically, list comprehensions in any language are for the most part , nested `flat_maps`. Using a map is simply more visible because it represents less complexity.

Having this in mind, I personally do not favor this specific change. What I have is the equivalent of:

```
[map(create_yolo_annotation_with_file, annotation_file.annotations)]
```
Which I have abstracted away into a function that I can reuse everywhere (same strategy i used for superannotate and labelbox). 

If I were to use Python's full functional power, I would probably end up with something like:

```
pipe(
  map(create_yolo_annotation_with_file, annotation_file.annotations), 
  list
)
```

Which is debatable. This, if you catch my _wave_, will likely remind you of Elixir (only a little). 
I understand this is Python. Not Elixir. 

But I also defend that when using abstractions from any science, I ought to make them clearly visible/descriptive. 
I cannot think of anything more descriptive than `_map_list`.

darwin/exporter/formats/yolo.py
Fl4m3Ph03n1x: I was actually considering having a file called `rich_settings.py` where we would have all the Rich settings. 
This would likely create/return a Singleton for the entire app to use. 

Open to debate.

andreaazzini:
Closing because they are too old. Happy to reopen later

PR #368, opened 2022-03-02
Detect if new class will clash with existing class of different type.

myFree:
This is likely to change soon ... and will then require re-updating and releasing darwin-py.
https://www.notion.so/v7labs/Allow-Same-Name-classes-of-different-Types-424578ba84a14c118101ccd3ed29366f

But in the meantime; could it be worth giving a friendly; informative helpful message to the end user?

```
Fetching remote class list...
Retrieving local annotations ...
Fetching remote file list...
1 annotation file(s) found.
`one` class of type `bounding_box` from imported annotations clashes with an existing class of the same name; but of type `tag`
`Surface damage` class of type `keypoint` from imported annotations clashes with an existing class of the same name; but of type `bounding_box`
Class names must be unique; even across types (see above error); can not proceed with import.
```

### DIFF
--- a/darwin/importer/importer.py
+++ b/darwin/importer/importer.py
@@ -96,6 +96,7 @@ def _resolve_annotation_classes(
 ) -> Tuple[Set[dt.AnnotationClass], Set[dt.AnnotationClass]]:
     local_classes_not_in_dataset: Set[dt.AnnotationClass] = set()
     local_classes_not_in_team: Set[dt.AnnotationClass] = set()
+    clashing_name_warnings: Set[str] = set()
 
     for local_cls in local_annotation_classes:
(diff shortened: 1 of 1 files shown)


Fl4m3Ph03n1x:
Is there a way to test this ? 
Under what specific scenario would this code come into play?

myFree:
I tested in manually.

This is the case we had the 2 hour triage with the customer on; when trying to create a class of the same name as some other already existing class type.

Fl4m3Ph03n1x:
One day we will have unit tests for that module. Until that day comes, your word will have to do.
Approved on my side.

Fl4m3Ph03n1x (approved):

PR #370, opened 2022-03-08
Import annotations in parallel 

simedw:


darwin/importer/importer.py
```diff
@@ -165,7 +166,7 @@ def import_annotations(
     ValueError
         If file_paths is not a list.
     """
-
+    print(format)
```
simedw, line 169:
```suggestion
```

darwin/importer/importer.py
```diff
@@ -264,8 +265,30 @@ def import_annotations(
...
+            remote_classes,
+            dataset.client,
+            attributes,
+            append,
+        ),
+        len(filenames),
+        True,
+    )
+    if len(result) != len(filenames):
+        print(f"error, only {len(result)} files succeeded (expected {len(filenames)})")
```
simedw, line 285: Do we want to raise here?

darwin/importer/importer.py
```diff
@@ -80,10 +81,10 @@ def build_attribute_lookup(dataset: "RemoteDataset") -> Dict[str, Any]:
 def get_remote_files(dataset: "RemoteDataset", filenames: List[str]) -> Dict[str, int]:
     """Fetches remote files from the datasets, in chunks of 100 filesnames at a time"""
     remote_files = {}
-    for i in range(0, len(filenames), 100):
-        chunk = filenames[i : i + 100]
+    for i in track(range(0, len(filenames), 1000)):
+        chunk = filenames[i : i + 1000]
```
simedw, line 85:
```suggestion
    for i in track(range(0, len(filenames), 250)):
        chunk = filenames[i : i + 250]
```

darwin/importer/importer.py
```diff
@@ -80,10 +81,10 @@ def build_attribute_lookup(dataset: "RemoteDataset") -> Dict[str, Any]:
 def get_remote_files(dataset: "RemoteDataset", filenames: List[str]) -> Dict[str, int]:
     """Fetches remote files from the datasets, in chunks of 100 filesnames at a time"""
     remote_files = {}
-    for i in range(0, len(filenames), 100):
-        chunk = filenames[i : i + 100]
+    for i in track(range(0, len(filenames), 250)):
+        chunk = filenames[i : i + 250]
```
andreaazzini, line 85: What do you guys think of having `chunk_size` as an argument of the function (defaulting to 250)

darwin/importer/importer.py
```diff
@@ -264,8 +264,30 @@ def import_annotations(
         remote_classes = build_main_annotations_lookup_table(team_classes)
 
     # Need to re parse the files since we didn't save the annotations in memory
-    for local_path in set(local_file.path for local_file in local_files):
+    from darwin.dataset.utils import exhaust_generator
```
andreaazzini, line 267: Any reason why we don't import this function on top of this file?

darwin/importer/importer.py
andreaazzini: I would say so. We could maybe catch the exception in `cli_functions` to print a nicer error/warning message. I think we shouldn't `print` in this module at all, even though we are doing it multiple times in this function.

andreaazzini (changes requested):
The importer works well now.

I'd change a few small things in the code, and potentially solve the inconsistencies in the way we use rich to show progress (as in, sometimes we show ETA, sometimes we don't, etc)

andreaazzini:
@simedw what should we do here? Are you reviewing my code?

owencjones:
Closing as stale

PR #373, opened 2022-03-21
Reduce API surface. Improved docs.

Fl4m3Ph03n1x:


darwin/importer/importer.py
```diff
@@ -125,12 +159,14 @@ def _resolve_annotation_classes(
...
     return local_classes_not_in_dataset, local_classes_not_in_team, clashing_name_warnings
 
+
 def _class_name_clash_error(
     local_cls_name: str,
     local_annotation_type: str,
     existing_type: str,
 ) -> str:
-    return f'`{local_cls_name}` class of type `{local_annotation_type}` from imported annotations clashes with an existing class of the same name; but of type `{existing_type}`'
+    return f"`{local_cls_name}` class of type `{local_annotation_type}` from imported annotations clashes with an existing class of the same name; but of type `{existing_type}`"
```
tomvars, line 168: This line is too long

andreaazzini (changes requested):
We are deprecating functions that are going to be turned into private ones, but we don't introduce their private counterparts here. Is that intended?

Fl4m3Ph03n1x:
Yes, for the time being, it is intended.

andreaazzini (approved):
Approving, since when we remove those functions it will be obvious we'll have to add the private functions

PR #378, opened 2022-03-21
Added docs

Fl4m3Ph03n1x:


andreaazzini (approved):

darwin/importer/formats/dataloop.py
```diff
@@ -42,3 +57,5 @@ def _parse_annotation(annotation: Dict[str, Any]) -> Optional[dt.Annotation]:
         x1, y1 = coords[0]["x"], coords[0]["y"]
         x2, y2 = coords[1]["x"], coords[1]["y"]
         return dt.make_bounding_box(annotation_label, x1, y1, x2 - x1, y2 - y1)
+
+    return None
```
tomvars, line 61: This is not necessary, if you get to the end of a python function without returning anything it returns None. Unless its being written there to be more explicit.

tomvars (commented):
Looks good! left 1 comment

darwin/importer/formats/dataloop.py
Fl4m3Ph03n1x: It is there for 2 reasons:
1. to be more explicit
2. to shut up pylance reporting an error for not explicitly returning it

PR #403, opened 2022-04-26
Added docs to utils.py

Fl4m3Ph03n1x:


darwin/torch/utils.py
```diff
@@ -16,13 +16,19 @@ def convert_segmentation_to_mask(segmentations: List[Segment], height: int, widt
...
-        segmentations: list of float values -> [x1, y1, x2, y2, ..., xn, yn]
-        height: image's height
-        width: image's width
-
-    Output:
-        torch.tensor
+    Parameters
+    ----------
+    segmentations : List[Segment]
+        List of float values -> ``[x1, y1, x2, y2, ..., xn, yn]``.
```
kamilc, line 22: If it's a `List[Segment]]` should an example instead be: `[[x1,1 y11, x12, y12], ..., [xn1, yn1, xn2, yn2]]` ?

darwin/torch/utils.py
```diff
@@ -36,13 +42,37 @@ def convert_segmentation_to_mask(segmentations: List[Segment], height: int, widt
...
+    Returns
+    -------
+    float
+        The area of the polygon.
     """
     return 0.5 * np.abs(np.dot(x, np.roll(y, 1)) - np.dot(y, np.roll(x, 1)))
 
 
-def collate_fn(batch) -> Tuple:
+def collate_fn(batch: Iterable[Any]) -> Tuple:
```
kamilc, line 62: You can switch `Any` for `Tuple` here. The idea of the `collate_fn` is to turn a list of tuples into a tuple of lists.

kamilc (changes requested):
I've found two minor issues

darwin/torch/utils.py
Fl4m3Ph03n1x: fixed !

darwin/torch/utils.py
Fl4m3Ph03n1x: fixed!

kamilc (approved):
👍

PR #405, opened 2022-04-27
fixing documentation errors

Fl4m3Ph03n1x:
Some files Sphinx errors and warnings, which resulted in broken docs. 
This PR fixes all warnings and errors.

### DIFF
--- a/darwin/importer/formats/coco.py
+++ b/darwin/importer/formats/coco.py
@@ -9,9 +9,12 @@ import darwin.datatypes as dt
 from darwin.path_utils import deconstruct_full_path
 from darwin.version import __version__
 
-DEPRECATION_MESSAGE = """This function is going to be turned into private. This means that breaking 
+DEPRECATION_MESSAGE = """
+
+This function is going to be turned into private. This means that breaking 
 changes in its interface and implementation are to be expected. We encourage using ``parse_annotation`` 
-instead of calling this low-level function directly."""
+instead of calling this low-level function directly.
+"""
 
 
 def parse_path(path: Path) -> Optional[List[dt.AnnotationFile]]:
@@ -151,10 +154,10 @@ def parse_annotation(annotation: Dict[str, Any], category_lookup_table: Dict[str
 )
 def decode_binary_rle(data: str) -> List[int]:
     """
-    decodes binary rle to integer list rle
+    Decodes binary rle to integer list rle.
     """
     m = len(data)
-    cnts = [0] * m
+    counts = [0] * m
     h = 0
     p = 0
     while p < m:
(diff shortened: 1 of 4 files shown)


darwin/importer/formats/coco.py
```diff
@@ -9,9 +9,12 @@
 from darwin.path_utils import deconstruct_full_path
 from darwin.version import __version__
 
-DEPRECATION_MESSAGE = """This function is going to be turned into private. This means that breaking 
+DEPRECATION_MESSAGE = """
+
```
andreaazzini, line 13: Is this double newline intended?

darwin/importer/formats/superannotate.py
```diff
@@ -55,21 +56,22 @@ def parse_path(path: Path) -> Optional[AnnotationFile]:
...
+
     Currently we support the following annotations:
 
         - point ``Vector``: https://doc.superannotate.com/docs/vector-json#point
         - ellipse ``Vector``: https://doc.superannotate.com/docs/vector-json#ellipse
         - cuboid ``Vector``: https://doc.superannotate.com/docs/vector-json#cuboid
-        - bbox ``Vector`` (not rotated): https://doc.superannotate.com/docs/vector-json#bounding-box-and-rotated-bounding-box  
-        - polygon and polyline ``Vector``s: https://doc.superannotate.com/docs/vector-json#polyline-and-polygon
+        - bbox ``Vector`` (not rotated): https://doc.superannotate.com/docs/vector-json#bounding-box-and-rotated-bounding-box
+        - polygon and polyline ``Vector``\\s: https://doc.superannotate.com/docs/vector-json#polyline-and-polygon
```
andreaazzini, line 66: out of curiosity, what was the problem here?

darwin/importer/formats/coco.py
Fl4m3Ph03n1x: Yes. It removes a warning from sphinx.

Fl4m3Ph03n1x:
To be clear here, the main issue is that the CLI for sphinx docs will **always** pass as green, even with errors and warnings. Looks like the only time the pipeline will break is if a critical error that prevents all documentation from being generated happens. 

This is not ideal imo, it means errors will be added undetected. I need to find a solution for this eventually.

andreaazzini:
> I need to find a solution for this eventually.
Can we do this now, or else ticket it down?

Fl4m3Ph03n1x:
Thus far, it is not possible to remove some of the warning we have in sphinx, due to a bug that makes compatibility with numpydoc not possible:

https://github.com/sphinx-doc/sphinx/issues/7817

I am looking into a workaround, as this negatively impacts our documentation:
doc

Fl4m3Ph03n1x:
Found a workaround that fixes this issue. Also documented it in our internal guide:
https://www.notion.so/v7labs/Python-Code-Style-351a7831c2074e78bf4e21ec13777542

Applying workaround everywhere now.

kamilc (approved):
👏

PR #409, opened 2022-05-27
Added "short" option for dataset remote sub command

frostasm:
Added ability to print only dataset names to stdout. This feature is useful for processing program execution results in bash scripts.


owencjones:
Closing as this has been able to become stale, by no fault of the author.  We will be keeping a better eye in future.

PR #410, opened 2022-05-27
Fix issue with cyrillic text in csv report

frostasm:
Command `darwin dataset report ${dataset}` generate wrong csv data when `first_name` or `last_name` contains cyrillic text.

image

After current fix
image


owencjones:
Closing this because it's become stale, but thank you for the contribution.  We'll endeavour to get to community contributions faster here on.

Issue #412, opened 2022-06-02
Uploading a COCO format json with "attributes" defined in it for each annotation doesn't appear in the platform. 


francozacco:
Hi everyone, I'm submitting this issue because I downloaded a `coco.json` from the platform which had a series of attributes for each annotation defined. So under the "extra" key of each annotation, I have
```json
"extra": {"attributes": ["some-attribute", "some-other-attribute"]}
```
but then after downloading the COCO formatted json I wanted to re-upload the same file to another dataset I created with the same images and the attributes defined previously don't appear in the platform anymore for this new dataset.
All the classes and segmentation masks seem to be uploaded correctly.

andreaazzini:
Hi @francozacco, thank you for opening this issue.

I will make sure this is ticketed, so we can work on a fix ASAP.
In the meantime, let us know if you are keen on fixing this issue on your personal fork of darwin-py, and I can sync with you to can open a PR and add in your contribution!

Cheers,
Andrea

francozacco:
Hi @andreaazzini, yes I'll be working on a solution on my personal fork, would love to open a PR as soon as I have a fix to add my contribution.

Cheers,
Franco.

francozacco:
Hello @andreaazzini I started this PR [https://github.com/v7labs/darwin-py/pull/413](https://github.com/v7labs/darwin-py/pull/413) to fix this, let me know what you think. Thank you.

Cheers,
Franco.

francozacco:
@andreaazzini  any news about this? The solution as I mentioned above can be found [here](https://github.com/v7labs/darwin-py/pull/413).
Cheers,
Franco.

owencjones:
As the PR, before the repo had a dedicated maintainer, fell behind master and is now closed, I can only close this with my apologies.

I am implementing a better system for highlighting github bugs to the internal team.

Owen V7

PR #413, opened 2022-06-03
COCO attributes are parsed as SubAnnotations

francozacco:
This Pull Request refers to the following issue: [https://github.com/v7labs/darwin-py/issues/412](https://github.com/v7labs/darwin-py/issues/412)

The idea is that the COCO importer appends the attributes defined for each annotation. The attributes should be defined inside the `"extra"` key i.e.
```python
"annotations": [
    {
         "id": 1,
         "image_id": 1,
         "category_id": 18,
         "segmentation": [[ ... ]],
         "area": 11624.0,
         "bbox": [ ... ],
         "iscrowd": 0,
         "extra": {"attributes": ["some-attr", "some-other-attr"]}
    },
...]
```
The importer now should be able to parse the attributes and send them to the platform correctly.

### DIFF
--- a/darwin/importer/formats/coco.py
+++ b/darwin/importer/formats/coco.py
@@ -102,12 +102,13 @@ def parse_annotation(annotation: Dict[str, Any], category_lookup_table: Dict[str
         print("Warning, unsupported RLE, skipping")
         return None
 
+    subs = _build_sub_annotations(annotation)
     if len(segmentation) == 0 and len(annotation["bbox"]) == 4:
         x, y, w, h = map(int, annotation["bbox"])
-        return dt.make_bounding_box(category["name"], x, y, w, h)
+        return dt.make_bounding_box(category["name"], x, y, w, h, subs=subs)
     elif len(segmentation) == 0 and len(annotation["bbox"]) == 1 and len(annotation["bbox"][0]) == 4:
         x, y, w, h = map(int, annotation["bbox"][0])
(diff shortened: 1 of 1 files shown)


gpldecha:
Hi, 

What is the process on getting this merged @andreaazzini  ? We have raised this with our customer success representatives at V7. Also where is the best place to make feature requests ?

Kind regards,

Guillaume

darwin/importer/formats/coco.py
```diff
@@ -141,11 +142,29 @@ def parse_annotation(annotation: Dict[str, Any], category_lookup_table: Dict[str
...
+    Returns:
+        List[SubAnnotation]: SubAnnotation list or None if no extra info is
+        available.
+    """
+    extra_info = annotation.get("extra")
+    attributes = extra_info.get("attributes")
+    sub_ann = None
+    if extra_info and attributes:
+        sub_ann = [dt.make_attributes(attributes)]
+    return sub_ann
```
simedw, line 165: I have seen coco files missing the `extra` field, so we need check that it exists before looking for attributes. 

```suggestion
    extra_info = annotation.get("extra")
    if not extra_info:
        return None
    attributes = extra_info.get("attributes")
    if not attributes:
        return None
    return [dt.make_attributes(attributes)]
```

owencjones:
Hi all,

This predates my joining the company, but unfortunately fell significnatly behind master, so I'm closing - please reopen if it's still persistent as an issue.

PR #415, opened 2022-06-14
Support v2 backend

simedw:
`RemoteDataset` becomes and abstract class with an implementation for v1 and one for v2. 
Same for `UploadHandler`.

Also all 2.0 endpoints are in the `backend_v2.py` file which is accessed via `client.api_v2`, this will make it easier to introduce test and make `client` a bit thinner.


darwin/dataset/upload_manager.py
```diff
@@ -395,6 +452,99 @@ def callback(monitor):
...
+
+            upload_response.raise_for_status()
+        except Exception as e:
+            raise UploadRequestError(file_path=file_path, stage=UploadStage.UPLOAD_TO_S3, error=e)
+
+        try:
+            self.client.api_v2.confirm_upload(dataset_slug, upload_id, team_slug=team_slug)
+        except Exception as e:
+            raise UploadRequestError(file_path=file_path, stage=UploadStage.CONFIRM_UPLOAD_COMPLETE, error=e)
+
```
andreaazzini, line 546: As far as I can see, all that changed here is that we point to `api_v2`. If that's the case, looks good. If not, the code is quite hard to parse, so would be great to be pointed to the parts of the code that substantially differ from v1.

andreaazzini (commented):
Some comments before I even try and use it, but finally managed to make a full first pass

darwin/dataset/upload_manager.py
```diff
@@ -134,6 +157,12 @@ def _type_check(self, args) -> None:
         self.data["filename"] = args.get("filename") or self.local_path.name
         self.data["path"] = args.get("path") or "/"
 
+    def serialize(self):
+        return {"files": [{"file_name": self.data["filename"], "slot_name": "0"}], "name": self.data["filename"]}
+
```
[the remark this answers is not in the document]
simedw: I believe so, yes

.github/workflows/tests.yml
```diff
@@ -5,6 +5,7 @@ on: [push]
 jobs:
   ubuntu:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
```
andreaazzini, line 8: do we still want to keep this? Could be even lower I guess

darwin/client.py
```diff
@@ -86,15 +89,26 @@ def list_remote_datasets(self, team_slug: Optional[str] = None) -> Iterator[Remo
...
+            else:
+                yield RemoteDatasetV1(
+                    name=dataset["name"],
+                    slug=dataset["slug"],
+                    team=team_slug or self.default_team,
+                    dataset_id=dataset["id"],
+                    item_count=dataset["num_images"] + dataset["num_videos"],
+                    progress=dataset["progress"],
+                    client=self,
+                )
```
andreaazzini, line 111: Minor, but since we are almost repeating this resolution logic for remote dataset class, we could think of creating a `resolve_remote_dataset` function.

tests/darwin/client_test.py
```diff
@@ -7,7 +7,7 @@
 
 from darwin.client import Client
 from darwin.config import Config
-from darwin.dataset.remote_dataset import RemoteDataset
+from darwin.dataset.remote_dataset_v1 import RemoteDatasetV1
```
andreaazzini, line 10: We should write the same tests for V2...

.github/workflows/tests.yml
simedw: It's just a safety guard from spending too many github action credits. I think I want to keep it.

darwin/client.py
simedw: since it's limited to only be inside client.py I'm not too keen on trying it to DRY it. I think that they will diverge more as time goes on.

andreaazzini (approved):

PR #417, opened 2022-06-24
Fix dataclasses dependency for Python>=3.7

campbell-remedy:
According to https://pypi.org/project/dataclasses/ in Python>=3.7 the built-in module should be used instead.
This was causing issues for us on Python3.8 where the darwin dataclasses dependency was causing issues by using the PyPi module not the built-in one which is incompatible.


simedw:
Thanks for adding this @campbell-remedy 🎉

PR #420, opened 2022-07-10
[V7-3306] 2.0 Comment support

simedw:
Refactors out post_comments from cli_functions to remotedataset,
    Adds a new implementation for v2 datasets.

### DIFF
--- a/darwin/dataset/remote_dataset.py
+++ b/darwin/dataset/remote_dataset.py
@@ -41,6 +41,7 @@ from darwin.exporter.formats.darwin import build_image_annotation
 from darwin.item import DatasetItem
 from darwin.item_sorter import ItemSorter
 from darwin.utils import parse_darwin_json, split_video_annotation, urljoin
+from numpy import void
 from rich.console import Console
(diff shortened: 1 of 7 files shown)


darwin/dataset/remote_dataset.py
```diff
@@ -758,6 +759,12 @@ def workview_url_for_item(self, item: DatasetItem) -> str:
             The url.
         """
 
+    @abstractmethod
+    def post_comment(self, item_id: str, text: str, x: int, y: int, w: int, h: int) -> void:
```
andreaazzini, line 763:
```suggestion
    def post_comment(self, item_id: str, text: str, x: int, y: int, w: int, h: int) -> None:
```

andreaazzini (commented):
Everything looks good, but:
- importing `void` from numpy for typing is not right
- tests are missing

darwin/dataset/remote_dataset.py
simedw: Good point!

PR #422, opened 2022-08-10
[SWAT-394][internal] fix JSONDecodeError exception when non 200 response doesn't contain json payload

AndriiKlymchuk:
This PR removes response.json() call because if response content is not in json format it rises exception which leads to losing of required debug information.


andreaazzini (commented):
LGTM! 👍 

Can we write a couple of unit tests for `has_json_response` and `get_response_debug_text`?

andreaazzini (approved):
Perfect! Approved

PR #425, opened 2022-08-12
[Bug] Dataset Creation fix

simedw:
The num_video field is no longer supplied when creating a dataset, made it optional.

### DIFF
--- a/darwin/client.py
+++ b/darwin/client.py
@@ -100,7 +100,7 @@ class Client:
                     slug=dataset["slug"],
                     team=team_slug or self.default_team,
                     dataset_id=dataset["id"],
-                    item_count=dataset.get("item_count", dataset.get("num_images", 0) + dataset.get("num_videos", 0)),
+                    item_count=dataset.get("num_items", dataset.get("num_images", 0) + dataset.get("num_videos", 0)),
                     progress=dataset["progress"],
                     client=self,
                 )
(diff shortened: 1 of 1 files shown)


darwin/client.py
```diff
@@ -170,7 +175,7 @@ def get_remote_dataset(self, dataset_identifier: Union[str, DatasetIdentifier])
                     slug=dataset["slug"],
                     team=parsed_dataset_identifier.team_slug,
                     dataset_id=dataset["id"],
-                    item_count=dataset.get("item_count", dataset["num_images"] + dataset["num_videos"]),
+                    item_count=dataset.get("num_items", dataset["num_images"] + dataset["num_videos"]),
```
andreaazzini, line 178: Why not using `.get()` here?

darwin/client.py
```diff
@@ -211,7 +216,7 @@ def create_dataset(self, name: str, team_slug: Optional[str] = None) -> RemoteDa
                 team=team_slug or self.default_team,
                 slug=dataset["slug"],
                 dataset_id=dataset["id"],
-                item_count=dataset.get("item_count", dataset["num_images"] + dataset["num_videos"]),
+                item_count=dataset.get("num_items", dataset["num_images"] + dataset["num_videos"]),
```
andreaazzini, line 219: and here?

darwin/client.py
simedw: it's v1

andreaazzini (approved):

PR #426, opened 2022-08-16
[ANN-133][internal] Export to YOLO format

andreaazzini:
image

[Source](https://towardsdatascience.com/image-data-labelling-and-annotation-everything-you-need-to-know-86ede6c684b1)

### DIFF
--- a/darwin/exporter/formats/yolo.py
+++ b/darwin/exporter/formats/yolo.py
@@ -59,11 +59,14 @@ def _build_txt(annotation_file: dt.AnnotationFile, class_index: ClassIndex) -> s
         else:
             continue
 
+        if annotation.data is None:
+            continue
+
         i = class_index[annotation.annotation_class.name]
-        x = round(data.get("x"))
-        y = round(data.get("y"))
-        w = round(data.get("w"))
-        h = round(data.get("h"))
+        x = round(data["x"])
+        y = round(data["y"])
+        w = round(data["w"])
+        h = round(data["h"])
 
         yolo_lines.append(f"{i} {x} {y} {w} {h}")
     return "\n".join(yolo_lines)
--- a/darwin/exporter/formats/yolo.py
+++ b/darwin/exporter/formats/yolo.py
@@ -1,0 +1,75 @@
+from pathlib import Path
+from typing import Dict, Iterable
+
+import darwin.datatypes as dt
+
+ClassIndex = Dict[str, int]
+
+
+def export(annotation_files: Iterable[dt.AnnotationFile], output_dir: Path) -> None:
+    """
+    Exports the given ``AnnotationFile``\\s into the YOLO format inside of the given
+    ``output_dir``.
+
+    Parameters
+    ----------
+    annotation_files : Iterable[dt.AnnotationFile]
+        The ``AnnotationFile``\\s to be exported.
+    output_dir : Path
+        The folder where the new pascalvoc files will be.
+    """
+
+    annotation_files = list(annotation_files)
+
+    class_index = _build_class_index(annotation_files)
+
+    for annotation_file in annotation_files:
+        _export_file(annotation_file, class_index, output_dir)
+
+    _save_class_index(class_index, output_dir)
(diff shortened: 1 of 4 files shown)


darwin/exporter/formats/yolo.py
```diff
@@ -0,0 +1,75 @@
...
+        if annotation_type == "bounding_box":
+            data = annotation.data
+        elif annotation_type in ["polygon", "complex_polygon"]:
+            data = annotation.data
+            data = data.get("bounding_box")
+        else:
+            continue
+
+        i = class_index[annotation.annotation_class.name]
+        x = round(data.get("x"))
```
tomvars, line 63: round will error if when 'x' is missing from data. Would change this to:
```
x = data.get('x')
y = data.get('y')
w = data.get('w')
h = data.get('h')
if None not in [x, y, w, h]:
    yolo_lines.append(f"{i} {round(x)} {round(y)} {round(w)} {round(h)}")
```

darwin/exporter/formats/yolo.py
```diff
@@ -0,0 +1,75 @@
...
+    for annotation in annotation_file.annotations:
+        annotation_type = annotation.annotation_class.annotation_type
+        if annotation_type == "bounding_box":
+            data = annotation.data
+        elif annotation_type in ["polygon", "complex_polygon"]:
+            data = annotation.data
+            data = data.get("bounding_box")
+        else:
+            continue
+
```
tomvars, line 61: data.get("x") will fail if data is None, could do a `if data is None: continue` here

darwin/exporter/formats/yolo.py
simedw: we should be able to expect x,y,w,h from a bounding box, shouldn't we? 

Otherwise it does make sense to fail loudly (not silently just ignore the annotation)

andreaazzini: Yeah, we should indeed always expect those values (to the point I should avoid using `.get` here)

darwin/exporter/formats/yolo.py
```diff
@@ -0,0 +1,78 @@
...
+        f.write(txt)
+
+
+def _build_class_index(annotation_files: Iterable[dt.AnnotationFile]) -> ClassIndex:
+    index: ClassIndex = {}
+    for annotation_file in annotation_files:
+        for annotation in annotation_file.annotations:
+            class_name = annotation.annotation_class.name
+            if class_name not in index:
+                index[class_name] = len(index)
```
simedw, line 46: the annotation_files list is not guaranteed to be in a certain order, so everytime we run this, we might get a new mapping between class_name and count - which isn't ideal...

```
classes = set()
 for annotation_file in annotation_files:
            classes.add(annotation.annotation_class.name)

return {k: v for (v,k) in enumerate(classes)}
```
wdyt?

darwin/exporter/formats/yolo.py
```diff
@@ -0,0 +1,78 @@
...
+        h = round(data["h"])
+
+        yolo_lines.append(f"{i} {x} {y} {w} {h}")
+    return "\n".join(yolo_lines)
+
+
+def _save_class_index(class_index: ClassIndex, output_dir: Path) -> None:
+    with open(output_dir / "darknet.labels", "w") as f:
+        for class_name in class_index:
+            f.write(f"{class_name}\n")
```
simedw, line 78: if we are not using the value in the dict why, is class_index a dict to begin with? 
Just a list would be good enough?

```
for class_name in sorted(class_index):
   f.write(f"{class_name}\n")
```

andreaazzini: We need the dictionary because the annotation files `.txt` expect an integer index, not the class name. So we first build the index, then build all annotations, and finally write the index to a file.

simedw: In that case I think my point in https://github.com/v7labs/darwin-py/pull/426#discussion_r947326825 holds true

Issue #427, opened 2022-08-16
There is no support for passing the number of maximum workers for the upload.


ahmedosama9777:
Hello,

While using the `pull` method in `RemoteDataset`, I noticed that it can be useful to pass the parameter max_workers as an integer to the `upload` method of `UploadHandler`, rather than just have it set by defauly to `None`, as agreed in [issue #187](https://github.com/v7labs/darwin-py/pull/187).

Specifying the number of `max_workers`, can be useful, for some people, they might want files to be uploaded sequentially and thus have only 1 worker, other maybe for bandwidth limitatoins, and there might be other cases as well. 

All in all, I think it can be a good enhancement.

ahmedosama9777:
Issue fixed and merged into master!

PR #428, opened 2022-08-16
Pass number of max workers for upload

ahmedosama9777:
Hello,

This is for [issue #427 ](https://github.com/v7labs/darwin-py/issues/427).

Now, the value of `max_workers`, can be passed from `RemoteDataset` objects directly into `UploadHandler`. Additionally, there are some validation that is done on `max_workers` in `UploadHandler`.

### DIFF
--- a/darwin/dataset/remote_dataset_v2.py
+++ b/darwin/dataset/remote_dataset_v2.py
@@ -111,7 +111,7 @@ class RemoteDatasetV2(RemoteDataset):
         *,
         blocking: bool = True,
         multi_threaded: bool = True,
-        max_workers: int = None,
+        max_workers: Optional[int] = None,
         fps: int = 0,
         as_frames: bool = False,
         files_to_exclude: Optional[List[PathLike]] = None,
--- a/darwin/dataset/remote_dataset_v2.py
+++ b/darwin/dataset/remote_dataset_v2.py
@@ -111,6 +111,7 @@ class RemoteDatasetV2(RemoteDataset):
         *,
         blocking: bool = True,
         multi_threaded: bool = True,
+        max_workers: int = None,
         fps: int = 0,
         as_frames: bool = False,
         files_to_exclude: Optional[List[PathLike]] = None,
@@ -131,6 +132,8 @@ class RemoteDatasetV2(RemoteDataset):
         multi_threaded : bool, default: True
             Uses multiprocessing to upload the dataset in parallel.
             If blocking is False this has no effect.
+        max_workers : int, default: None
+            Maximum number of workers to use for parallel upload.
         fps : int, default: 0
             When the uploading file is a video, specify its framerate.
         as_frames: bool, default: False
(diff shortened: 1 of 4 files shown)


darwin/dataset/upload_manager.py
```diff
@@ -338,6 +338,14 @@ def callback(file_name, file_total_bytes, file_bytes_sent):
                     file_complete.add(file_name)
                     progress_callback(self.pending_count, 1)
 
+        if max_workers:
+            if max_workers < 1:
+                raise ValueError("max_workers must be greater than 0")
+            elif max_workers > concurrent.futures.ThreadPoolExecutor()._max_workers:
+                raise ValueError(
```
ahmedosama9777, line 345: I'd suggest that here, instead of raising a `ValueError` exception, we can just set the `max_workers` to be the max workers possible on the machine.

darwin/dataset/remote_dataset_v2.py
```diff
@@ -111,6 +111,7 @@ def push(
         *,
         blocking: bool = True,
         multi_threaded: bool = True,
+        max_workers: int = None,
```
rslota, line 114:
```suggestion
        max_workers: Optional[int] = None,
```

darwin/dataset/upload_manager.py
rslota: I would actually rather to raise here, rather then implicitly "fix" things behind the scene, so IMO we can keep the code we have here.

rslota (commented):
Thanks for the PR! I have small comment about typing, but otherwise it looks good!

darwin/dataset/upload_manager.py
ahmedosama9777: Alright cool!

darwin/dataset/remote_dataset_v2.py
ahmedosama9777: All done.
@rslota hope you can resolve this one.

rslota (approved):

ahmedosama9777:
Cool!

rslota (changes requested):
@ahmedosama9777 I'm very sorry, but we can't merge this just yet. I missed one thing in my original review. The argument `max_workers` is added here for class `RemoteDatasetV2` which is a child of `RemoteDataset`. This means, the changed function is now different the the matching abstract method and implementation in the other sub-class - `RemoteDatasetV1`. Would you mind updating those as well?

ahmedosama9777:
@rslota All done!

rslota:
@ahmedosama9777 I think you missed the abstract method in `RemoteDataset` :)

ahmedosama9777:
sorry for being silly! Done

rslota (approved):

ahmedosama9777:
@rslota all done now?

rslota:
@ahmedosama9777 Yes, all looks good. Merged :)

PR #432, opened 2022-08-29
[SWAT-428][external] Add naive implementation for V2 Darwin imports

rslota:


brentgracey (approved):
Am able to import a v2 export in this branch.

rslota:
> Am able to import a v2 export in this branch.

Thanks for testing! 
I myself tested both image and videos on both V1 -> V1 and V2 -> V2 cases.

PR #434, opened 2022-08-30
[SWAT-431][external] Fixed commenting items not working

Fl4m3Ph03n1x:
Posting comments on items was broken for V1. This PR fixes that, while also adding some missing docs. 

Ticket: https://linear.app/v7labs/issue/SWAT-431/bug-comments-arent-being-added-via-cli

Tested in Staging for V1 and V2.

### DIFF
--- a/darwin/cli_functions.py
+++ b/darwin/cli_functions.py
@@ -995,7 +995,7 @@ def post_comment(
     item: DatasetItem = items.pop()
 
     try:
-        dataset.post_comment(str(item.id), text, x, y, w, h)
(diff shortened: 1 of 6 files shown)


rslota:
@Fl4m3Ph03n1x Can you describe what the problem was in the PR description? There are plenty of changes here and it's not obvious to me what was the issue.

rslota:
How is it possible that we were fetching different item while using its ID?

Fl4m3Ph03n1x:
@rslota Will do!

rslota (approved):

PR #436, opened 2022-08-30
Support for Darwin JSON 2.0

rslota:
This PR and branch was initially based on `support-darwin-json-v2` by @simedw .
Since then it's been rebased on master and contains few major fixes that make the conversion working in all tested by me cases. This still requires quite a lot of testing though.


darwin/utils.py
```diff
@@ -312,119 +290,108 @@ def parse_darwin_json(path: Path, count: Optional[int]) -> Optional[dt.Annotatio
...
-        An AnnotationFile with the information from the parsed JSON file.
-    """
-
-    annotations: List[dt.VideoAnnotation] = list(filter(None, map(parse_darwin_video_annotation, data["annotations"])))
-    annotation_classes: Set[dt.AnnotationClass] = set([annotation.annotation_class for annotation in annotations])
-
-    if "width" not in data["image"] or "height" not in data["image"]:
-        raise OutdatedDarwinJSONFormat("Missing width/height in video, please re-export")
+    slot = dt.Slot(
+        name="default",
```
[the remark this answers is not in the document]
rslota: Well, not really. Everywhere we use "0" as default (I know you prefer "default", but that was a decision made based on discussion with Alberto half a year ago). 
That being said, I'm not sure that matters. This value here is meaningless, there is no way it can be used in any way. Maybe it should be `None` instead to make sure it doesn't get used?

darwin/backend_v2.py
```diff
@@ -70,8 +70,10 @@ def fetch_items(
          Dict[str, Any]
             A response dictionary with the file information.
         """
+        if isinstance(cursor, dict):
+            cursor = list(cursor.items())
 
-        cursor["dataset_ids"] = dataset_id
+        cursor.append(("dataset_ids[]", dataset_id))
```
tomvars, line 76: something's wrong here

darwin/cli_functions.py
```diff
@@ -966,7 +988,7 @@ def convert(format: str, files: List[PathLike], output_dir: Path) -> None:
     except AttributeError:
         _error(f"Unsupported export format, currently supported: {export_formats}")
 
-    export_annotations(parser, files, output_dir)
+    export_annotations(parser, files, output_dir, split_sequences=(format != "darwin_1.0"))
```
tomvars, line 991: If we're changing the format name I would probably favour a different naming convention such as `darwin_v1`, The use of underscores and full stops interchangeably in these string names is likely to cause issues.

tomvars: Also, should this change affect the supported formats list? Does this also have effects on busybee code used for conversion etc?

darwin/dataset/remote_dataset.py
```diff
@@ -228,7 +230,7 @@ def pull(
         if release is None:
             release = self.get_release()
 
-        if release.format != "json":
+        if release.format != "json" and release.format != "darwin_json_2":
```
tomvars, line 233: Again, if we're going for consistency they should both be `darwin_json_v1` and `darwin_json_v2` or `darwin_v1` and `darwin_v2` but not a mix

darwin/datatypes.py
```diff
@@ -604,7 +701,155 @@ def make_cuboid(
     Annotation
         A cuboid ``Annotation``.
     """
-    return Annotation(AnnotationClass(class_name, "cuboid"), cuboid, subs or [], slot_names=slot_names)
+    return Annotation(AnnotationClass(class_name, "cuboid"), cuboid, subs or [], slot_names=slot_names or [])
+
+
+def make_table(
```
tomvars, line 707: I appreciate this is nice to have but maybe this could be in a separate PR? Doesn't feel necessary for the migration.

darwin/utils.py
```diff
@@ -312,129 +290,131 @@ def parse_darwin_json(path: Path, count: Optional[int]) -> Optional[dt.Annotatio
...
+            image_height=None,
+            image_url=None,
+            image_thumbnail_url=None,
+            workview_url=item_source.get("workview_url", None),
+            seq=0,
+            frame_urls=None,
+            remote_path=item["path"],
+        )
+    else:
+        slot = slots[0]
```
tomvars, line 328: this seems like a strange hard-coding. Are we ignoring other slots? how do we parse the annotations of slots other than the first?

darwin/utils.py
```diff
@@ -312,129 +290,131 @@ def parse_darwin_json(path: Path, count: Optional[int]) -> Optional[dt.Annotatio
...
+        name=None,
+        type="image",
+        source_files=[{"url": data["image"].get("url"), "file_name": _get_local_filename(data["image"])}],
+        thubmnail_url=data["image"].get("thubmnail_url"),
+        width=data["image"].get("width"),
+        height=data["image"].get("height"),
+    )
 
-    return dt.AnnotationFile(
+    annotation_file = dt.AnnotationFile(
```
tomvars, line 378: This is not an action point for this PR but using named arguments would make these AnnotationFile instantiations much easier to read.

darwin/utils.py
rslota: This has nothing to do with annotations here, this is just about metadata like file names etc. This is mostly backwards compatibility with V1 format, where we still take values from root of `AnnotationFile`. All V2 code is looking at `slots` directly and pull data from there.

darwin/datatypes.py
rslota: It is required. We allow those types to be exported, and since our fallback to Darwin JSON 1.0 works by converting from 2.0 via darwin-py, darwin-py has to support **all** annotation types, otherwise we'll get regression on export.

darwin/utils.py
rslota: True, but yes, I don't want to put more time into this PR right now.

darwin/dataset/remote_dataset.py
rslota: We can't change old format name. This is exposed in our APIs, this wouldn't be backwards compatible change. In V2 APIs the old format will be eventually dropped entirely, but we cannot rename it right now. 
I agree that consistency would be cool, I just don't see an option to do it.

darwin/cli_functions.py
rslota: We never had conversion into darwin, as it doesn't make any sense. This a new format, so no, it doesn't affect busybee. Also, I don't want this on a list of supported formats tbh, I don't think that it's a good idea to have legacy stuff like that so openly available. This is a fallback that is supposed to be used:
* by us in busybee to support legacy format in V2 dataset **temporarily** 
* by customers that create exports via CLI if they really, really need legacy format, also **temporarily**

rslota: As for name - please post this on the channel on Slack `tmp-darwin-json-2`. The name that we have here was discussed in quite wide audience, I don't feel like overriding this decision without discussing with others that initially approved this. 
I don't really buy argument about full stops being an issue, I've never seen that yet myself. And full stop here isn't used interchangeably with underscore - those mean different things. Thus, let's please discuss that with wider audience.

darwin/backend_v2.py
rslota: Can you elaborate?

darwin/cli_functions.py
```diff
@@ -378,8 +389,19 @@ def pull_dataset(
...
+        dataset.pull(
+            release=release,
+            only_annotations=only_annotations,
+            use_folders=folders,
+            video_frames=video_frames,
+            force_slots=force_slots,
+        )
         print_new_version_info(client)
+        if release.format == "darwin_json_2":
+            print(f"NOTE: Your dataset has been exported using new Darwin JSON 2.0 format.")
```
simedw, line 401: Look how `print_new_version_info` is implemented, it prints to stderr so that we can safely pipe the output we care about

darwin/dataset/download_manager.py
```diff
@@ -165,33 +176,85 @@ def download_image_from_annotation(
...
             _download_image(frame_url, path, api_key)
     else:
-        image_url = annotation["image"]["url"]
-        image_path = parent_path / sanitize_filename(annotation["image"]["filename"])
-        _download_image(image_url, image_path, api_key)
+        if len(slot.source_files) > 0:
+            image_url = slot.source_files[0]["url"]
+            image_path = parent_path / sanitize_filename(slot.filename or annotation.filename)
+            _download_image(image_url, image_path, api_key)
+            _update_local_path(annotation, image_url, image_path)
```
simedw, line 239: why don't we loop over source files? the function only does the first slot, but if it consists of mutliple source files shouldn't they all be downloaded, `_update_local_path` seems to be able to handle it since it matches on `image_url` which will be unique per source_file.

darwin/datatypes.py
```diff
@@ -301,7 +394,7 @@ def make_bounding_box(
         AnnotationClass(class_name, "bounding_box"),
         {"x": round(x, 3), "y": round(y, 3), "w": round(w, 3), "h": round(h, 3)},
         subs or [],
-        slot_names=slot_names or [],
+        slot_names=slot_names or [] or [],
```
simedw, line 397:
```suggestion
        slot_names=slot_names or [],
```

simedw: there are a few occurrences in this file, unless I'm missing something

darwin/exporter/exporter.py
```diff
@@ -37,7 +37,12 @@ def darwin_to_dt_gen(file_paths: List[PathLike]) -> Iterator[AnnotationFile]:
             count += 1
 
 
-def export_annotations(exporter: ExportParser, file_paths: List[PathLike], output_directory: PathLike) -> None:
+def export_annotations(
+    exporter: ExportParser,
+    file_paths: List[PathLike],
+    output_directory: PathLike,
+    split_sequences: bool = True,
```
simedw, line 44: no function doc for `split_sequences`

darwin/exporter/formats/darwin_1_0.py
```diff
@@ -0,0 +1,134 @@
+import json
+from pathlib import Path
+from typing import Any, Dict, Iterable, List, Union
+
+import darwin.datatypes as dt
+from darwin.exporter.formats.numpy_encoder import NumpyEncoder
+from darwin.version import __version__
```
simedw, line 7: why are we importing `__version__` here?

darwin/exporter/formats/darwin_1_0.py
```diff
@@ -0,0 +1,134 @@
...
+    return {"full_name": author.name, "email": author.email}
+
+
+def _build_sub_annotation(sub: dt.SubAnnotation) -> Dict[str, Any]:
+    if sub.annotation_type == "instance_id":
+        return {sub.annotation_type: {"value": sub.data}}
+    elif sub.annotation_type == "attributes":
+        return {sub.annotation_type: sub.data}
+    elif sub.annotation_type == "text":
+        return {sub.annotation_type: {"text": sub.data}}
```
simedw, line 82: we are missing `directional vectors`, `measures` and `inference`

darwin/exporter/formats/darwin_1_0.py
```diff
@@ -0,0 +1,134 @@
...
+    else:
+        return {**base_json, "slot_names": annotation.slot_names}
+
+
+def _build_legacy_annotation_data(annotation_class: dt.AnnotationClass, data: Dict[str, Any]) -> Dict[str, Any]:
+    if annotation_class.annotation_type == "complex_polygon":
+        data["path"] = data["paths"][0]
+        data["additional_paths"] = data["paths"][1:]
+        del data["paths"]
+        return {annotation_class.annotation_internal_type or "polygon": data}
```
simedw, line 132:
```suggestion
        return {"complex_polygon": data}
```
don't you want this? you already know that it's a complex polygon

darwin/utils.py
```diff
@@ -312,129 +290,131 @@ def parse_darwin_json(path: Path, count: Optional[int]) -> Optional[dt.Annotatio
...
+            frame_urls=slot.frame_urls,
+            remote_path=item["path"],
+        )
 
-    Returns
-    -------
-    dt.AnnotationFile
-        An AnnotationFile with the information from the parsed JSON file.
-    """
+    annotation_file.slots.extend(slots)
```
simedw, line 346: it really feels like slots would be better as an optional argument to the constructor of AnnotationFile, defaulting to `[]`

simedw: but it's not super important

darwin/cli_functions.py
rslota: Changed!

darwin/dataset/download_manager.py
rslota: Are we fine with the fact that we'll get more files then items in target directory? We do loop over `source_files` when `--force_slots` option is used, because we also get a nice directory structure that groups source files per slot. Here we put everything in one spot, so it's gonna be a bit of a mess. 

But your call - if you confirm we want that either way - I'll do the change.

darwin/datatypes.py
rslota: Fixed, bad grep/replace.

darwin/exporter/exporter.py
rslota: Added

darwin/exporter/formats/darwin_1_0.py
rslota: I have no clue, probably IDE acting out. Removed.

darwin/exporter/formats/darwin_1_0.py
rslota: I see no mention of any of them anywhere in darwin-py. I've added generic clause, because I think importer does that too already. I hope this handles those cases - could you take a look during your functional pass? I'll do the same if I find out how to create those.

darwin/exporter/formats/darwin_1_0.py
rslota: fair anough 👍

darwin/utils.py
rslota: It already is actually, we don't need that line - removed.

simedw (approved):

PR #440, opened 2022-09-09
[SWAT-444][External] Added missing path to item

Fl4m3Ph03n1x:
Fixes issue on:
https://linear.app/v7labs/issue/SWAT-444/[bug]-files-are-stuck-in-uploading-when-uploading-them-to-a-folder-via


rslota (approved):

rslota:
@Fl4m3Ph03n1x Do we have any existing tests for upload manager that we can extend with a check for this field?

Fl4m3Ph03n1x:
> @Fl4m3Ph03n1x Do we have any existing tests for upload manager that we can extend with a check for this field?

Will check

dsienkiewicz (approved):

PR #443, opened 2022-09-12
[SWAT-450][External] Skipping import of file if no annotations are present

Fl4m3Ph03n1x:
- Fixes crash when importing an annotations file that has no annotations
- Revamped the print messages to use Rich instead of `print` (consistency changes).

Ticket: https://linear.app/v7labs/issue/SWAT-450/regression-error-when-importing-empty-annotation-files

### DIFF
--- a/darwin/importer/importer.py
+++ b/darwin/importer/importer.py
@@ -180,6 +180,7 @@ def import_annotations(
     delete_for_empty : bool, default: False
         If ``True`` will use empty annotation files to delete all annotations from the remote file.
         If ``False``, empty annotation files will simply be skipped.
+        Only works for V2 datasets.
 
     Raises
     -------
@@ -284,10 +285,14 @@ def import_annotations(
     else:
         remote_classes = build_main_annotations_lookup_table(team_classes)
 
-    console.print(
-        "Importing annotations...\nEmpty annotations will be skipped, if you want to delete annotations rerun with '--delete-for-empty' ",
-        style="info",
-    )
+    if dataset.version == 1:
+        console.print("Importing annotations...\nEmpty annotations will be skipped.", style="info")
+    else:
+        console.print(
+            "Importing annotations...\nEmpty annotations will be skipped, if you want to delete annotations rerun with '--delete-for-empty' ",
+            style="info",
+        )
+
     # Need to re parse the files since we didn't save the annotations in memory
     for local_path in set(local_file.path for local_file in local_files):
 
@@ -304,7 +309,7 @@ def import_annotations(
         parsed_files = [parsed_file for parsed_file in parsed_files if parsed_file.full_path not in missing_files]
 
         for parsed_file in track(parsed_files):
-            if parsed_file.annotations or delete_for_empty:
+            if parsed_file.annotations or (delete_for_empty and dataset.version == 2):
                 image_id = remote_files[parsed_file.full_path]
                 _import_annotations(
                     dataset.client, image_id, remote_classes, attributes, parsed_file.annotations, dataset, append
--- a/darwin/importer/importer.py
+++ b/darwin/importer/importer.py
@@ -1,4 +1,3 @@
-from ast import Or
 from pathlib import Path
 from typing import (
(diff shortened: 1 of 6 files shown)


dsienkiewicz (approved):

darwin/importer/importer.py
```diff
@@ -175,19 +177,26 @@ def import_annotations(
         If ``True`` appends the given annotations to the datasets. If ``False`` will override them.
     class_prompt : bool
         If ``False`` classes will be created and added to the datasets without requiring a user's prompt.
+    delete_for_empty : bool, default: False
+        If ``True`` will use empty annotation files to delete all annotations from the remote file.
+        If ``False``, empty annotation files will simply be skipped.
+        Only works for V2 datasets.
+        Takes precedence over the ``append`` flag.
```
rslota, line 184: Let's please not do that. It's a very bad idea to allow a flag to override another one. Easy to loose data due to simple copy-paste mistake. Those flags should never be used together and darwin-py should IMO print error if someone tries.

darwin/importer/importer.py
```diff
@@ -378,4 +418,13 @@ def _import_annotations(
     if append:
         payload["overwrite"] = "false"
 
+    if delete_for_empty and not annotations:
```
rslota, line 421: We should add `payload["overwrite"] = "true"` in all other cases where we don't set `false`. We seem to have a bug on BE side that doesn't properly default to `true`, which makes this impossible to test this PR unless we default to `true` on darwin-py side.

darwin/importer/importer.py
```diff
@@ -287,11 +313,24 @@ def import_annotations(
...
+                    image_id,
+                    remote_classes,
+                    attributes,
+                    parsed_file.annotations,
+                    dataset,
+                    append,
+                    delete_for_empty,
+                )
+            else:
+                console.print(
```
rslota, line 331: This is gonna look really odd, we're already in `track` progress bar, so it's gonna show a empty progress bar and then this error. IMO we should look for this condition before we start the progress bar and display this warn instead.

rslota (changes requested):

darwin/importer/importer.py
Fl4m3Ph03n1x: Fixed. Now I print an error and stop the process.

darwin/importer/importer.py
Fl4m3Ph03n1x: Done !

darwin/importer/importer.py
Fl4m3Ph03n1x: Ooooki. Also fixed. Progress bar is no longer destroyed. 
I also added a nice warning that lets the user know if he is using `--delete-for-empty` with a V1 dataset. Sometimes our users can be confused, so a warning should save us some swat ticket. 

Let me know !

rslota (changes requested):
I have 2 more minor comments, otherwise this seems to be working as expected 👍

darwin/importer/importer.py
```diff
@@ -377,5 +440,15 @@ def _import_annotations(
     payload: Dict[str, Any] = {"annotations": serialized_annotations}
     if append:
         payload["overwrite"] = "false"
+    elif delete_for_empty and not annotations:
```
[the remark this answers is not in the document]
Fl4m3Ph03n1x: True, but I wanted to make it explicit. Like "no matter what you do here, if this 2 things happen, overwrite must be true".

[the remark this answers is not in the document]
rslota: To me personally, this is just confusing. In this function, as this point, neither `delete_for_empty` nor annotations being empty matter. This function is and should be as simple as "if append, then false, otherwise always true". 
Otherwise one could think we have some logic to correctly handle the case of `delete_for_empty and not annotations` while we don't - we expect those to be handled earlier by this function execution of lack there of.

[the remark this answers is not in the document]
Fl4m3Ph03n1x: Done!

rslota (approved):

PR #444, opened 2022-09-15
[IO-311][external] Add support to setting 'complete' status on items

rslota:
This is a new feature that adds ability to set items as "complete" for both V1 and V2 system.
The implementation for V2 is pretty simple and doesn't have anything weird about it.
The implementation for V1 is pretty horrible:
* We store more workflow data now, which is un-parsed and looks like it needs at least 2 more classes to store it "nicely" 
* We have pretty complex logic to circumvent the fact that API works per-workflow template, while items may be in different ones. We bail on any validation error by saying that item needs to be assigned. This feels similar to frontend does and is related to this virtual workflow called `default_auto_complete`. I don't fully understand how this works, so I would appreciate confirmation from @begedin that this logic is sound. The error is mostly reworded error that I saw on our UI.

That being said, I don't want to put more effort into this on V1, I don't think it's worth it. Let's please discuss issues with the logic, or maybe some simple improvements, but I don't want to refactor V1 specific code here.

### DIFF
--- a/darwin/client.py
+++ b/darwin/client.py
@@ -3,7 +3,7 @@ import os
 import time
 from logging import Logger
 from pathlib import Path
-from typing import Any, Dict, Iterator, List, Optional, Union, cast
+from typing import Any, Dict, Iterator, List, Optional, Tuple, Union, cast
 from urllib import parse
 
 import requests
@@ -25,6 +25,7 @@ from darwin.exceptions import (
     Unauthorized,
     ValidationError,
 )
+from darwin.item import DatasetItem
 from darwin.utils import (
     get_response_content,
     has_json_content_type,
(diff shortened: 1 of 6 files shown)


darwin/client.py
```diff
@@ -797,7 +819,7 @@ def post_workflow_comment(
 
         return comment_id
 
-    def instantiate_item(self, item_id: int) -> int:
+    def instantiate_item(self, item_id: int, include_metadata: bool = False) -> Union[int, Tuple[int, DatasetItem]]:
```
rslota, line 822: This looks weird, but I didn't want to make backwards incompatible change here, not sure if this is used in SDK mode. @simedw maybe you can say if that's sensible approach, or maybe that's not part of public part of SDK and I can just refactor this to always return DatasetItem.

simedw: `instantiate_item` is not really needed in v2 as an item should always be in workflow.

rslota: This is for V1 only.

rslota: I mean, this change. This PR addresses both versions.

simedw (approved):

PR #449, opened 2022-09-28
Fix the dataset create bug where num_videos could be missing in RemoteDatasetV1

biendltb:
Hello,

It was a fix for a bug with dataset create in RemoteDatasetV2 [here](https://github.com/v7labs/darwin-py/commit/7bebe08a4e416ddfb0aee70d3de2e9e5176d887e). However, the issue still persists with RemoteDatasetV1.

This fix is about to solve it.


simedw (approved):
Thank you for adding the fix @biendltb !

PR #454, opened 2022-10-05
Video annotation split branch

ChristofferEdlund:
Pulling video data with *--video-frames* argument only transform the videos into correct format, but not the annotations. Generating a miss-match between the two. This PR addresses this by also transforming the annotations into a format that takes the parsed image frames into account.


owencjones:
@ChristofferEdlund is this still a current issue?  If so, can you ping me in linear or slack, and I'll see it on to the backlog, seeing as you've done much of the work some time ago.

Nathanjp91 (approved):

JBWilkie:
I think this makes sense, but I'm worried that it might be considered breaking. Also, ` split_video_annotations()` can be invoked at any point after pulling a dataset. I'll close this PR but I've recorded this change for consideration as part of 0.9.1 which will include some breaking changes to the way darwin-py names files!

PR #455, opened 2022-10-09
[ANN-241][external] import nifti as polygon

tomvars:
Importing niftis as polygons.


darwin/importer/formats/nifti.py
```diff
@@ -0,0 +1,288 @@
...
+    if path.suffix != ".json":
+        return None
+    with open(path, "r") as f:
+        data = json.load(f)
+        validate(data, schema=nifti_import_schema)
+    nifti_annotations = data.get("data")
+    if nifti_annotations is None:
+        return None
+    annotation_files = []
+    for nifti_annotation in nifti_annotations:
```
rslota, line 53: Are those annotations all for different items? Each `annotation_file` is imported separately and running clear of annotations by default for that item. So having more then one `annotation_file` per Darwin Item would lead to some unexpected behaviors.

darwin/importer/formats/nifti.py
```diff
@@ -0,0 +1,288 @@
...
+                continue
+            video_annotations.append(video_annotation)
+    annotation_classes = set(
+        [dt.AnnotationClass(class_name, "polygon", "polygon") for class_name in class_map.values()]
+    )
+    return dt.AnnotationFile(
+        path=json_path,
+        filename=str(filename),
+        annotation_classes=annotation_classes,
+        annotations=video_annotations,
```
rslota, line 117: `slots` should also be added, as fields like `filename` should be deprecated very soon. All V2 logic uses or should use data from `dt.Slot` not from root of `AnnotationFile`. This code doesn't crash only because right now use-case using it (import) doesn't need this metadata, but I think this is not a reason to create broken `AnnotationFile`. I know `slot_name` is not known here, and for that we still create `Slot` with `name: None`.

rslota: You can take a look at how Darwin JSON 1.0 is read and decoded and do the same.

darwin/importer/formats/nifti.py
```diff
@@ -0,0 +1,288 @@
...
+            slice_mask = img[:, :, i].astype(np.uint8)
+            for class_name, class_idxs in processed_class_map.items():
+                frame_annotations = {}
+                if class_name == "background":
+                    continue
+                class_mask = np.isin(slice_mask, class_idxs).astype(np.uint8).copy()
+                polygon = mask_to_polygon(mask=class_mask, class_name=class_name)
+                if polygon is None:
+                    continue
+                frame_annotations[i] = polygon
```
rslota, line 93: I'm a bit confused - we expect single frame sequence, and yet there's a variable here as index. Why is that?

darwin/importer/importer.py
```diff
@@ -128,7 +128,7 @@ def get_remote_files(dataset: "RemoteDataset", filenames: List[str], chunk_size:
         for remote_file in dataset.fetch_remote_files(
             {"types": "image,playback_video,video_frame", "filenames": chunk}
         ):
-            remote_files[remote_file.full_path] = remote_file.id
+            remote_files[remote_file.full_path.strip("/")] = remote_file.id
```
rslota, line 131: Isn't this a breaking change? This code is working on `master` and it feels like there is no way that both versions are fine.

darwin/importer/formats/nifti.py
tomvars: Hi Rafal, "image" mode is perhaps confusingly named. A volume has been uploaded but in "image" mode the annotations are per class and per frame. Explained in these two Looms https://www.loom.com/share/30833a63ef2c49bc8b8f8f0eadecb9fa and https://www.loom.com/share/b25929896c7e40c9836e8c48322ea5bc

darwin/importer/formats/nifti.py
```diff
@@ -0,0 +1,291 @@
...
+    Returns
+    -------
+    Optional[List[dt.AnnotationFile]]
+        Returns ``None`` if the given file is not in ``json`` format, or ``List[dt.AnnotationFile]``
+        otherwise.
+    """
+    if not isinstance(path, Path):
+        path = Path(path)
+    if path.suffix != ".json":
+        return None
```
andreaazzini, line 48: Should we warn users here?

darwin/importer/formats/nifti.py
```diff
@@ -0,0 +1,291 @@
...
+        keyframes={f_id: True for f_id in all_frame_ids},
+        segments=segments,
+        interpolated=False,
+        slot_names=[],
+    )
+    return video_annotation
+
+
+def mask_to_polygon(mask: np.ndarray, class_name: str) -> Optional[dt.Annotation]:
+    _labels, external_paths, _internal_paths = find_contours(mask)
```
andreaazzini, line 154: Is this the same code we have in centermask and clicker?

darwin/importer/formats/nifti.py
```diff
@@ -0,0 +1,288 @@
...
+
+try:
+    import cc3d
+    import nibabel as nib
+except ImportError:
+    import_fail_string = """
+    You must install `darwin-py` with `pip install darwin-py[medical]`
+    in order to import with using nifti format
+    """
+    print(import_fail_string)
```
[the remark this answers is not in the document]
andreaazzini: Ignore, this was addressed already

darwin/importer/formats/nifti.py
tomvars: Hey Rafal! I'm not sure how to add the slots here. It works without any of this metadata, and the `slot_name` is not known here. Wouldn't know what to fill in for any of these fields (except maybe width and height, but not sure how that feeds in to any frontend process):

```
dt.Slot(
            name=data["slot_name"],
            type=data["type"],
            width=data.get("width"),
            height=data.get("height"),
            source_files=data.get("source_files", []),
            thumbnail_url=data.get("thumbnail_url"),
            frame_count=data.get("frame_count"),
            frame_urls=data.get("frame_urls"),
            fps=data.get("fps"),
        ),
  ```
  Should I define an empty Slot object and pass it?

darwin/importer/formats/nifti.py
tomvars: Similar, maybe not exactly the same. Why?

andreaazzini: It might be wise to turn this into a more generic helper that we can use both here and in the model integrations, but it's ok to keep this out of scope

darwin/importer/formats/nifti.py
tomvars: this is each for a separate item 👍  so should be ok

darwin/importer/formats/nifti.py
rslota: Ok, I think I get it now. This is not an image either way ;)

rslota (approved):
I don't fully understand all conversions here, I just reviewed things that are closer to our internal format and that looks good 👍

andreaazzini (approved):

PR #458, opened 2022-10-14
[IO-361][external] Fix FPS setting for V2 uploads

rslota:


Fl4m3Ph03n1x (approved):

darwin/dataset/upload_manager.py
```diff
@@ -158,11 +158,13 @@ def _type_check(self, args) -> None:
         self.data["path"] = args.get("path") or "/"
 
     def serialize(self):
+        print(self.data)
         return {"files": [{"file_name": self.data["filename"], "slot_name": "0"}], "name": self.data["filename"]}
 
     def serialize_v2(self):
+        print(self.data)
```
dsienkiewicz, line 165: Should we remove it?

dsienkiewicz (approved):

darwin/dataset/upload_manager.py
rslota: Yup, 💯 . Thanks for noticing ;)

PR #459, opened 2022-10-14
[IO-341][external] Add support for importing tags from COCO

rslota:


Fl4m3Ph03n1x (approved):
Since there is no corresponding test file for importing coco files, I will approve because creating a test file for this now would be rather out of scope and potentially overwhelming.

rslota:
> Since there is no corresponding test file for importing coco files, I will approve because creating a test file for this now would be rather out of scope and potentially overwhelming.

Yes, I did check that as well, if it did exist I would update it, but creating from scratch is a bit too much for me :(

Issue #462, opened 2022-10-17
Type error in convert_polygons_to_sequences() (Conversion to COCO)


SchernHe:
Hey, 

I am using `darwin-py==0.7.25` and it seems like there is a bug in `convert_polygons_to_sequences()` when calling: 

```
  # Create release
  dataset.export(name=release_name, annotation_class_ids=class_ids)
  
  # Download release
  release: Release = dataset.get_release(release_name)
  dataset.pull(release=release, multi_threaded=False)

  # Convert to COCO
  dataset_path = os.path.join(DARWIN_PATH, PROJECT, v7_dataset_name, "releases", release_name)
  annotation_paths = glob(f"{dataset_path}/**/*.json", recursive=True)
  parser = get_exporter("coco")
  export_annotations(parser, annotation_paths, dataset_dir)
```

**Traceback:**
```
Traceback (most recent call last):
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/operators/python.py", line 171, in execute
    return_value = self.execute_callable()
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/operators/python.py", line 189, in execute_callable
    return self.python_callable(*self.op_args, **self.op_kwargs)
  File "/home/airflow/auvisus-ml-airflow/auvisus_airflow/handler/darwin/download.py", line 63, in download_from_darwin
    export_annotations(parser, annotation_paths, dataset_dir)
  File "/home/airflow/.local/lib/python3.8/site-packages/darwin/exporter/exporter.py", line 53, in export_annotations
    exporter(darwin_to_dt_gen(file_paths), Path(output_directory))
  File "/home/airflow/.local/lib/python3.8/site-packages/darwin/exporter/formats/coco.py", line 35, in export
    output = _build_json(list(annotation_files))
  File "/home/airflow/.local/lib/python3.8/site-packages/darwin/exporter/formats/coco.py", line 324, in _build_json
    "annotations": list(_build_annotations(annotation_files, categories)),
  File "/home/airflow/.local/lib/python3.8/site-packages/darwin/exporter/formats/coco.py", line 402, in _build_annotations
    annotation_data = _build_annotation(annotation_file, annotation_id, annotation, categories)
  File "/home/airflow/.local/lib/python3.8/site-packages/darwin/exporter/formats/coco.py", line 436, in _build_annotation
    sequences = convert_polygons_to_sequences(annotation.data["paths"])
[...]

PR #464, opened 2022-10-18
[SWAT-526][external]: Bakcport _parse_darwin_annotation from 0.8.0

balysv:
Target branch is `0.7-patches` that is to be used for hotfixes until `0.8.0` is out

Closes https://github.com/v7labs/darwin-py/issues/462


darwin/utils.py
```diff
@@ -555,6 +580,20 @@ def _parse_darwin_annotation(annotation: Dict[str, Any]) -> Optional[dt.Annotati
         main_annotation.subs.append(dt.make_attributes(annotation["attributes"]))
     if "text" in annotation:
         main_annotation.subs.append(dt.make_text(annotation["text"]["text"]))
+    if "inference" in annotation:
+        main_annotation.subs.append(dt.make_opaque_sub("inference", annotation["inference"]))
+    if "directional_vector" in annotation:
```
rslota, line 585: All those new things won't work, as the dependency functions like `make_opaque_sub` don't exist in this version.

darwin/utils.py
```diff
@@ -544,6 +559,16 @@ def _parse_darwin_annotation(annotation: Dict[str, Any]) -> Optional[dt.Annotati
         main_annotation = dt.make_cuboid(name, annotation["cuboid"], slot_names=slot_names)
     elif "skeleton" in annotation:
         main_annotation = dt.make_skeleton(name, annotation["skeleton"]["nodes"], slot_names=slot_names)
+    elif "table" in annotation:
```
rslota, line 562: None of those types have the dependency functions present in this tree.

rslota (changes requested):
Let's only merge changes to polygons, as other changes have deeper dependencies.

rslota (approved):

PR #467, opened 2022-10-19
Throw error when unknown export version is used

rslota:


darwin/dataset/remote_dataset_v2.py
```diff
@@ -379,13 +380,15 @@ def export(
             When used for V2 dataset, allows to force generation of either Darwin JSON 1.0 (Legacy) or newer 2.0.
             Omit this option to get your team's default.
         """
-
-        if version == "2.0":
+        str_version = str(version)
```
Fl4m3Ph03n1x, line 383: We dont need to convert the values to a string. The following should work:

```
def _get_format(version):
...    if version == "1.0":
...       return "json"
...    elif version == "2.0":
...       return "darwin_json_2"
...    elif version == None:
...       return None
...    else:
...       raise ArgumentError(f"Unknown version '{version}'.")
... 
>>> 
>>> print(_get_format(None))
None

>>> print(_get_format("1.0"))
json

>>> print(_get_format("2.0"))
darwin_json_2

>>> print(_get_format("3.0"))
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 9, in _get_format
ctypes.ArgumentError: Unknown version '3.0'.
```

May I also suggest, that instead of `ArgumentError` we create a proper exception for this, since it will be both more specific and cause less confusion on cascading errors:

```
class UnknownVersion(Exception):
    """Used when dataset version is not recognized."""

    def __init__(self, version: str):
        """
        Parameters
        ----------
        version: str
            The version that is not recognized.
        """
        super().__init__()
        self.version = version

    def __str__(self):
        return f"Unknown version: '{self.version}'"
```

Fl4m3Ph03n1x (changes requested):

darwin/dataset/remote_dataset_v2.py
rslota: No, this won't work, because people can pass `1.0`, i.e. float to this function, and it's gonna look pretty odd if we say that `1.0` is not supported. It's gonna confuse users a lot.

rslota: I can add the special exception, but I'm not 100% sold on this. This is literally bad argument 🤷

rslota: I've added the exception, but unless I misunderstood your point, I'm leaving string conversion.

rslota:
@Fl4m3Ph03n1x I actually updated the exception in: https://github.com/v7labs/darwin-py/pull/468 . I'm closing this PR. Please review the other one if you still have concerns.

PR #468, opened 2022-10-19
Io 422 fix dataset item archiverestore

rslota:


dsienkiewicz (approved):

darwin/dataset/remote_dataset_v2.py
```diff
@@ -379,13 +368,15 @@ def export(
             When used for V2 dataset, allows to force generation of either Darwin JSON 1.0 (Legacy) or newer 2.0.
             Omit this option to get your team's default.
         """
-
-        if version == "2.0":
+        str_version = str(version)
```
Fl4m3Ph03n1x, line 371: If version can be a float, then the typespec is incorrect.

Fl4m3Ph03n1x (changes requested):

darwin/dataset/remote_dataset_v2.py
rslota: I can't be, typespec is correct, but this doesn't stop user for passing differently typed values.

rslota: Pretty much everyone that has tested this code passed float instead of string. People rarely look at types in python, especially less technical ones.

rslota: Thus, my string conversion is to reduce friction and not make customers wonder what's wrong if they pass `2.0` and exception says, that it has to be `"2.0"` instead. I bet many won't notice the difference and will open up a ticket to us instead that we have a bug.

Fl4m3Ph03n1x: Typespec is: 
```
  version : Optional[str], default: None, enum: ["1.0", "2.0"]
```

If the user passes a float, he is breaking the contract and we are under no obligation whatsoever to provide a valid result. 
If you are going to start bullet proof code this way, might was well just remove the typecpecs altogether and avoid the maintenance cost they come with.

I want to make this very clear: "Users not caring about the contracts our functions employ when using them, is a user problem". Not something we have to fix/code against.

People have the docs. The typespcs. If they still choose to use this incorrectly, its on them. Pointing the user the documentation is usually the preferred approach Not bulletproofing code against their lack of diligence. 

If you truly want this code to accept floats, then updating the typespec is in order.

Fl4m3Ph03n1x: After a slack conversation where I voiced my concerns, the decision was made. 
I will now approve this, but I want to make it clear that typespecs are now inconsistent. We cannot trust the guarantees they bring if we use the affected function.

Fl4m3Ph03n1x (approved):
Approved as per requested by team lead.

darwin/dataset/remote_dataset_v2.py
rslota: >  I want to make it clear that typespecs are now inconsistent. We cannot trust the guarantees they bring if we use the affected function.

I'm sorry but that's simply not true. The fact that we allow some unsupported types to work has nothing to do with actual typespec and their function. We have unbelievable amount of functions that do that even by mistake instead of crashing or via implicit casts. Types in python are information what input is correct. I don't want to extend this one to `float` becuase in some edge cases it may not work correctly, but I want "stupid" mistakes not to be annoying.

Since we're definitely not gonna agree on this, I would like to ask for 3rd opinion here 🙏 : @tomvars @andreaazzini

PR #472, opened 2022-10-20
[ANN 245][external] export annotations as nifti files

tomvars:
Nifti export PR with latest changes using slot level metadata


andreaazzini (approved):
This looks good to me, so pre-approving. It'd be ideal if this could be reviewed by someone in IO.

rslota (approved):

PR #473, opened 2022-10-20
[IO-427][external] Only export completed items by default

rslota:


darwin/backend_v2.py
```diff
@@ -148,15 +148,17 @@ def export_dataset(
...
             "format": format,
             "include_authorship": include_authorship,
             "include_export_token": include_token,
             "name": name,
-            "annotation_filters": {}
+            "annotation_filters": {},
         }
         if annotation_class_ids:
             payload["annotation_filters"] = {"annotation_class_ids": annotation_class_ids}
+        if filters is not None:
```
balysv, line 159: should `annotation_filters` behave in the same way?

balysv (approved):

darwin/backend_v2.py
rslota: Well, they do already. But even if not, that wouldn't be a problem, there are no magical, implicit `annotation_filters` like in case of regular `filters`.

PR #476, opened 2022-10-26
Allow running document GH action on demand

balysv:


rslota:
@balysv I'm sorry, I missed this PR somehow. I merged almost identical changes already, so I think we can just close this.

PR #479, opened 2022-10-31
[ANN-402][external] adding extract views flag in darwin py

tomvars:
Adds the option to add `--extract_views` at upload to get volumetric medical data split into its orthogonal planes.


darwin/dataset/upload_manager.py
```diff
@@ -162,9 +162,19 @@ def serialize(self):
 
     def serialize_v2(self):
```
owencjones, line 163: It's not a change you've made, but this function `serialize_v2` doesn't actually seem to serialize/serialise data, just makes it ready for serialisation?

tests/darwin/cli_functions_test.py
```diff
@@ -86,6 +87,7 @@ def with_verbose_flag(
                     None,
                     None,
                     False,
+                    False,
```
owencjones, line 90: More a general comment for this line, but I'd consider annotating (not to confuse things here - not the imagery type!) these properties, e.g.

```
upload_data(
    argumentname=argumentvalue
    [repeated as needed]
)
```

Just for test readability.

Not necessarily for now.  Also looks like a test is needed at some point for the new extract flag you've added, but that these tests don't have them for others so not as part of this task.

owencjones (approved):
Looks good, and good to see some of the code.

I've added some comments, and only one (the typo) needs definite attention.  The others are purely stylistic, and techdebt/backlogable.

darwin/dataset/upload_manager.py
```diff
@@ -162,9 +162,19 @@ def serialize(self):
...
         return {
-            "slots": [{"file_name": self.data["filename"], "slot_name": "0", "fps": self.data["fps"]}],
+            "slots": [
+                {
+                    "file_name": self.data["filename"],
+                    "slot_name": "0",
+                    "fps": self.data["fps"],
+                    "extract_views": self.data["extract_views"],
```
rslota, line 170: This change makes `extract_views` mandatory for SDK users, which it shouldn't. Please see how I refactored this here: https://github.com/v7labs/darwin-py/pull/486 and let's make this in similar way. I'm hoping my PRs will get merged soon, only waiting for QA now. Which yours will be waiting for too actually.

rslota (changes requested):

darwin/dataset/upload_manager.py
tomvars: Could you update this ? change has been made

PR #481, opened 2022-11-07
[ANN-433][external] changed yolo format to conform to Darknet standard

tomvars:
Changed from `<class_id> <top_left_x_in_pixels> <top_left_y_in_pixels> <width_in_pixels> <height_in_pixels>` to `<class_id> <top_left_x_normalized> <top_left_y_normalized> <width_normalized> <height_normalized>` to conform with the Darknet YOLO standard as in https://pjreddie.com/darknet/yolo/


darwin/exporter/formats/yolo.py
```diff
@@ -60,12 +60,22 @@ def _build_txt(annotation_file: dt.AnnotationFile, class_index: ClassIndex) -> s
...
 
         i = class_index[annotation.annotation_class.name]
-        x = round(data["x"])
-        y = round(data["y"])
-        w = round(data["w"])
-        h = round(data["h"])
+        # x, y should be the center of the box
+        # x, y, w, h are normalized to the image size
+        x = data["x"] + (data["w"] / 2)
+        y = data["y"] + (data["h"] / 2)
```
andreaazzini, line 70:
```suggestion
        x = data["x"] + data["w"] / 2
        y = data["y"] + data["h"] / 2
```

setup.py
```diff
@@ -36,8 +36,8 @@
         "pydantic",
     ],
     extras_require={
-        "test": ["responses", "pytest", "pytest-describe", "sklearn"],
-        "ml": ["sklearn", "torch", "torchvision"],
+        "test": ["responses", "pytest", "pytest-describe", "scikit-learn"],
+        "ml": ["scikit-learn", "torch", "torchvision"],
```
andreaazzini, line 40: Is this necessary?

andreaazzini (changes requested):

setup.py
tomvars: The import that was there is wrong see - https://pypi.org/project/sklearn/

tomvars: Leads to failed tests if run at certain times of day.

andreaazzini (approved):

PR #485, opened 2022-11-15
[IO-481][internal] Use persistent HTTP connections

rslota:


darwin/client.py
```diff
@@ -1084,7 +1087,7 @@ def _post_raw(
         if payload is None:
             payload = {}
 
-        response: Response = requests.post(
+        response: Response = self.session.post(
```
dsienkiewicz, line 1090: Should we also add the same behaviour for `_delete` ?

dsienkiewicz (approved):

PR #490, opened 2022-11-17
[IO-461][external] Convert RG16 images into grayscale

rslota:
See: https://linear.app/v7labs/issue/IO-461/request-darwin-dataset-pull-command-with-video-frames-option-downloads

### DIFF
--- a/darwin/dataset/download_manager.py
+++ b/darwin/dataset/download_manager.py
@@ -7,10 +7,15 @@ import json
 import time
 import urllib
 from pathlib import Path
-from typing import Any, Callable, Iterable, List, Tuple
+from typing import Any, Callable, Iterable, List, Optional, Tuple
 
 import deprecation
+import numpy as np
 import requests
+from PIL import Image
+from rich.console import Console
+
+import darwin.datatypes as dt
 from darwin.dataset.utils import sanitize_filename
 from darwin.datatypes import AnnotationFile
 from darwin.utils import (
@@ -20,7 +25,6 @@ from darwin.utils import (
     parse_darwin_json,
 )
 from darwin.version import __version__
-from rich.console import Console
 
 
 @deprecation.deprecated(
@@ -275,7 +279,7 @@ def _download_all_slots_from_json_annotation(annotation, api_key, parent_path, v
             video_path.mkdir(exist_ok=True, parents=True)
             for i, frame_url in enumerate(slot.frame_urls or []):
                 path = video_path / f"{i:07d}.png"
-                generator.append(functools.partial(_download_image, frame_url, path, api_key))
+                generator.append(functools.partial(_download_image, frame_url, path, api_key, slot))
         else:
             for upload in slot.source_files:
                 file_path = slot_path / sanitize_filename(upload["file_name"])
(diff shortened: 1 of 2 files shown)


darwin/dataset/download_manager.py
```diff
@@ -464,14 +473,33 @@ def _fetch_multiple_files(path: Path, response: requests.Response) -> None:
...
-def _write_file(path: Path, response: requests.Response) -> None:
+def _write_file(path: Path, response: requests.Response, transform_file_function=None) -> None:
     with open(str(path), "wb") as file:
         for chunk in response:
             file.write(chunk)
+    if transform_file_function is not None:
+        transform_file_function(path)
+
+
+def _rg16_to_grayscale(path):
```
tomvars, line 491: feel like this helper function should live in a `utils.py` style file not here

tomvars (changes requested):
Looks good to me, haven't tested it but let me know if you need me to.

darwin/dataset/download_manager.py
rslota: Why though? It's only used here. If we ever need it somewhere else, we could move it and make it more public then.

rslota:
> Looks good to me, haven't tested it but let me know if you need me to.

I've tested this with DICOM file I found in busybee tests which is RG16 and one of mine that results in RGB. I tested this for both V1 and V2 with both Darwin JSON 1.0 and Darwin JSON 2.0.

darwin/dataset/download_manager.py
tomvars: ok, agreed its unlikely to be used elsewhere 👍

tomvars (approved):

darwin/dataset/download_manager.py
```diff
@@ -464,14 +473,33 @@ def _fetch_multiple_files(path: Path, response: requests.Response) -> None:
...
+    image_2d_rgb = np.asarray(image)
+
+    image_2d_r = np.uint16(image_2d_rgb[:, :, 0]) << 8
+    image_2d_g = np.uint16(image_2d_rgb[:, :, 1])
+
+    image_2d_gray = np.bitwise_or(image_2d_r, image_2d_g)
+    image_2d_gray = image_2d_gray / (1 << 16) * 255
+
+    new_image = Image.fromarray(np.uint8(image_2d_gray), mode="L")
+    new_image.save(path)
```
owencjones, line 505: This is doing a decent job of the RGB -> Grayscale conversion based on focussing on R and G is better than the mean luminescence of R,G, and B, but if you find the outputs are lacking, the coeffiecients below are useful for grayscale transformation:

If so, here they are:
```
grey = 0.2989 * R + 0.5870 * G + 0.1140 * B
```

rslota: This is not RGB input though. This is our custom format that encodes 16 grayscale image on first 2 bytes of RGB tuple. "B" channel is all zeros in that case. Those files are generated by our backend here: https://github.com/v7labs/busybee/blob/master/dicom-extractor/dicom-extractor.py#L72

owencjones: Ah right, didn't realise this was a DICOM input.   Yeah, in that case my earlier comment is not at all applicable!

PR #491, opened 2022-11-21
[SWAT-605] [External] Darwin rename file format .JPG to .jpg when exporting

dsienkiewicz:
Take `original_filename` as filename, if not available, fallback to `filename`


owencjones (commented):
Might've been tempted to do it as an inline if, but honestly this is probably more readable!

PR #492, opened 2022-11-22
[IO-453][external] Add optional payload compression feature

brain-geek:


Fl4m3Ph03n1x (changes requested):
Have a couple of minor questions, but other than that looks OK

darwin/client.py
```diff
@@ -1087,9 +1090,18 @@ def _post_raw(
         if payload is None:
             payload = {}
 
-        response: Response = self.session.post(
-            urljoin(self.url, endpoint), json=payload, headers=self._get_headers(team_slug)
-        )
+        compression_level = int(self.config.get("global/payload_compression_level", "0"))
```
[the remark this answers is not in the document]
brain-geek: Generally in darwin.py we're assuming that config files are correct, and we have a validation on setting the value via CLI.

darwin/exceptions.py
```diff
@@ -34,6 +34,12 @@ class InvalidTeam(Exception):
     """
 
 
+class InvalidCompressionLevel(Exception):
+    """
```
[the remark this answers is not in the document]
brain-geek: Done!

darwin/cli_functions.py
```diff
@@ -150,6 +150,20 @@ def set_team(team_slug: str) -> None:
     config = _config()
     config.set_default_team(team_slug)
 
+def set_compression_level(compression_level: int) -> None:
+    """
+    Change the compression level of text/json contents sent to backend and persist the change on the configuration file.
```
rslota, line 155: `backend` is not the best word considering this is public message. How about "Darwin APIs"?

darwin/client.py
```diff
@@ -999,7 +1001,8 @@ def default_base_url() -> str:
         return os.getenv("DARWIN_BASE_URL", "https://darwin.v7labs.com")
 
     def _get_headers(self, team_slug: Optional[str] = None) -> Dict[str, str]:
-        headers: Dict[str, str] = {"Content-Type": "application/json"}
+        headers: Dict[str, str] = {"Content-Type": "application/json", "X-Darwin-Payload-Compression-Version": "1"}
```
rslota, line 1004: Should we set this even if later we don't actually compress? It feels this should be set based on the same condition.

darwin/client.py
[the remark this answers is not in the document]
owencjones: Worthwhile consideration for the future though - config validator

darwin/options.py
```diff
@@ -22,6 +22,15 @@ def __init__(self):
         # AUTHENTICATE
         subparsers.add_parser("authenticate", help="Authenticate the user. ")
 
+        # SET COMPRESSION LEVEL
+        parser_compression = subparsers.add_parser("compression", help="Set compression level.")
+        parser_compression.add_argument(
+            "compression_level",
+            type=int,
+            choices=range(0, 10),
+            help="Compression level to use on uploaded data. 0 is no compression, 9 is the best.",
```
owencjones, line 31: Seems like unless this is either set a default setting, or is set required, then the user could set `--compression` without `--compression-level` and no level would be present, which would then cause a reference error to an on darwin/cli.py:L72 ?

To me, seems like it needs either:
```python
required=True
```
or
```python
default=5 #or whatever it should be!
```

owencjones: PS. Generally prefer default settings over required arguments, as this is the argparse standard, but I don't yet know how well that suits this application.

darwin/cli.py
```diff
@@ -67,6 +67,9 @@ def _run(args: Namespace, parser: ArgumentParser) -> None:
             f.current_team()
         else:
             f.list_teams()
+    # Set compression level
+    elif args.command == "compression":
+        f.set_compression_level(args.compression_level)
```
owencjones, line 72: See comment about argparse.

owencjones (changes requested):

darwin/options.py
brain-geek:
```
> darwin compression
usage: darwin compression [-h] {0,1,2,3,4,5,6,7,8,9}
darwin compression: error: the following arguments are required: compression_level
> darwin compression 3
> darwin compression 10
usage: darwin compression [-h] {0,1,2,3,4,5,6,7,8,9}
darwin compression: error: argument compression_level: invalid choice: 10 (choose from 0, 1, 2, 3, 4, 5, 6, 7, 8, 9)
```

Could you explain how to run it without the parameter? As this is positional argument, it can't be set to `required=True` as I understand it, and it does not make sense to set `default` value as this argument is the whole point of the command.

owencjones: No, you're right, it works without, because it's positional.  My mistake.

owencjones (approved):

rslota (approved):

Fl4m3Ph03n1x (approved):

PR #493, opened 2022-11-23
[IO-492][external] Make tag_categories optional on coco

rslota:


darwin/importer/formats/coco.py
```diff
@@ -68,7 +68,8 @@ def parse_json(path: Path, data: Dict[str, Any]) -> Iterator[dt.AnnotationFile]:
     annotations = data["annotations"]
     image_lookup_table = {image["id"]: image for image in data["images"]}
     category_lookup_table = {category["id"]: category for category in data["categories"]}
-    tag_category_lookup_table = {category["id"]: category for category in data["tag_categories"]}
+    tag_categories = data.get("tag_categories") or []
```
owencjones, line 71:
```suggestion
    tag_categories = data.get("tag_categories", [])
```

rslota: I actually intentionally didn't do that, as I prefer to cover also a case where someone by mistake passes `None` as value of `tag_categories`. This is, if I'm not mistake, the only case where those approaches differ.

rslota: wdyt? I don't feel strongly about this, can change.

owencjones (approved):

dsienkiewicz (approved):

PR #497, opened 2022-11-28
IO-496: Request: Add support for importing polygons in Dataloop format

owencjones:
Dataloop importer was not handling the 'segment' type in dataloop formats.  Have added support for it with simple polygons - it doesn't appear to include complex polygons - and have written a test suite for the whole importer.

Also took the opportunity to:

* Add a `.python-version` set at 3.9 for `pyenv` users to use in development
* Add a `setup.cfg` with `flake8` settings
* Add some testing, linting, and formatting tools to the `setup.pt` that developers can install with.


.python-version
```diff
@@ -0,0 +1 @@
+3.9
```
owencjones, line 1: Using 3.9 as 3.10 and 3.11 both have issues compiling the `[ml]` options in pip install.  Opening tickets to address this.

darwin/importer/formats/dataloop.py
```diff
@@ -28,7 +28,11 @@ def parse_path(path: Path) -> Optional[dt.AnnotationFile]:
         annotations: List[dt.Annotation] = list(filter(None, map(_parse_annotation, data["annotations"])))
         annotation_classes: Set[dt.AnnotationClass] = set([annotation.annotation_class for annotation in annotations])
         return dt.AnnotationFile(
-            path, _remove_leading_slash(data["filename"]), annotation_classes, annotations, remote_path="/"
```
owencjones, line 31: Just formatting change, done automatically by Black

setup.py
```diff
@@ -6,7 +6,7 @@
 with open(Path(__file__).parent / "darwin" / "version" / "__init__.py", "r") as f:
     content = f.read()
     # from https://www.py4u.net/discuss/139845
-    version = re.search(r'__version__\s*=\s*[\'"]([^\'"]*)[\'"]', content).group(1)
+    version = re.search(r'__version__\s*=\s*[\'"]([^\'"]*)[\'"]', content).group(1)  # type: ignore
```
owencjones, line 9: Prevents mypy complaining about this - rather than refactor type, when we have to replace `setup.py` in the coming months anyway.

tests/darwin/importer/formats/import_dataloop_test.py
```diff
@@ -0,0 +1,143 @@
...
+    def setUp(self) -> None:
+        _fd = open(
+            realpath(
+                join(dirname(__file__), "..", "..", "data", "dataloop.example.json")
+            )
+        )
+        self.DATALOOP_MOCK_DATA = _fd.read()
+        _fd.close()
+
+    def assertApproximatelyEqualNumber(
```
owencjones, line 27: Seems odd, but prevents issues with floating point numbers being off by `.0000000000001` etc.

tests/darwin/importer/formats/import_dataloop_test.py
```diff
@@ -0,0 +1,138 @@
...
+        annotation = _parse_annotation(self.parsed_json["annotations"][1])  # 1 is a class type
+        self.assertEqual(annotation, None)
+
+    def test_handles_segment_type(self):
+        from darwin.importer.formats.dataloop import _parse_annotation as pa
+
+        with patch("darwin.importer.formats.dataloop.dt.make_polygon") as make_polygon_mock:
+            pa(self.parsed_json["annotations"][2])  # 2 is a segment type
+
+            if "kwargs" in make_polygon_mock.call_args:
```
owencjones, line 109: `MagicMock` and `Mock` before version 3.8 don't support testing `kwargs`, so testing this would be difficult, but this feature switch will mean we still test against 3.8 and 3.9 in both Windows and Ubuntu, which I think is plenty for a simple dict comprehension.

setup.cfg
```diff
@@ -0,0 +1,2 @@
+[flake8]
+max-line-length = 160
```
rslota, line 2: end-of-line 🙏

dsienkiewicz (approved):

rslota (approved):

PR #498, opened 2022-11-28
[SWAT-628][internal] Fix `get_dataset` for darwin JSON v2 exports.

whilefalse:
Linear ticket: https://linear.app/v7labs/issue/SWAT-628/get-dataset-returns-keyerror-image

`get_dataset` was failing when being called on a Darwin JSON v2 dataset.

This is because the `get_dataset` function ends up looking at the raw JSON, and the format has changed in Darwin JSON v2.

Changed to use the `parse_darwin_json` function (which handles v1 and v2, and outputs proper python classes). There was some code further down the pipeline that also read from the raw annotation JSON. These have also had to be changed to read from the new generic class formats.

Also added some tests where different types of datasets weren't being tested, and also test the, all with Darwin JSON v2.


darwin/torch/dataset.py
```diff
@@ -310,12 +310,13 @@ def get_target(self, index: int) -> Dict[str, Any]:
 
         annotations = []
         for annotation in target["annotations"]:
-            if "polygon" not in annotation and "complex_polygon" not in annotation:
+            annotation_type: str = annotation.annotation_class.annotation_type
+            path_key = "paths" if annotation_type == "complex_polygon" else "path"
+            if path_key not in annotation.data:
                 print(f"Warning: missing polygon in annotation {self.annotations_path[index]}")
```
owencjones, line 316: It's not your code, but it would help our bug diagnosis if this used a logger.  If it's too much for a SWAT ticket, hit me back and I'll add it to my backlog.

owencjones (approved):
Looks fine to me, although does this affect compatibility with V1?

darwin/dataset/local_dataset.py
```diff
@@ -248,20 +248,37 @@ def parse_json(self, index: int) -> Dict[str, Any]:
...
             "image_id": index,
             "image_path": str(self.images_path[index]),
-            "height": data["image"]["height"],
-            "width": data["image"]["width"],
+            "height": parsed.image_height,
+            "width": parsed.image_width,
             "annotations": annotations,
         }
 
+    def annotation_type_supported(self, annotation) -> List[str]:
```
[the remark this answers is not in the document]
whilefalse: Yeh good spot, thanks

darwin/torch/dataset.py
whilefalse: I'd rather change as little as possible if that's ok? Especially since I'm not really a python expert.

whilefalse:
> Looks fine to me, although does this affect compatibility with V1?

I hope not! The `get_dataset` tests still passed. I've added a few more to cover some cases that I think were missing, and also made sure those tests run on Darwin JSON v2 versions of the same files.

PR #500, opened 2022-12-02
Change to Py versions

owencjones:
PRs are failing because github python available versions changed


tomvars (approved):
looks good! hope tests passing = all functions working as expected.

PR #501, opened 2022-12-02
IO-507_fixing_imports_with_spaces

owencjones:


darwin/utils.py
```diff
@@ -210,13 +280,15 @@ def find_files(
     for f in files:
         path = Path(f)
         if path.is_dir():
-            found_files.extend([f for f in path.glob(pattern) if is_extension_allowed("".join(f.suffixes))])
-        elif is_extension_allowed("".join(path.suffixes)):
+            found_files.extend([f for f in path.glob(pattern) if is_extension_allowed(str(path))])
```
owencjones, line 283: These changes made only by formatter

darwin/utils.py
```diff
@@ -210,13 +280,15 @@ def find_files(
...
+        elif is_extension_allowed_by_filename(str(path)):
             found_files.append(path)
         else:
             raise UnsupportedFileType(path)
 
-    return [f for f in found_files if f not in map(Path, files_to_exclude)]
+    files_to_exclude_full_paths = [str(Path(f)) for f in files_to_exclude]
```
owencjones, line 289: Turned into two lines for readability, because applying the `str` as
```python
[f for f in found file if str(f) not in map(str, map(Path, files_to_exclude))]
```
is not too readable, and `import this` will tell us it should be.

Comparing by string, because comparing `PosixPath`, `WindowsPath`, or generally `PurePath` objects is quite unreliable.  Iteratively comparing `path.parents` and then `path.name` works, and casting to `str` provides `...parents/name` in a single string - which is more efficient to compare.

Comparison of parents and then name is `O(n)=n*(p+1)` - where p is number of parents.  Comparison of string output is `O(n)=1` - at least on the surface - but under the hood, we should at least assume that `pathlib` is relatively optimised.

tests/darwin/utils/find_files_test.py
```diff
@@ -0,0 +1,136 @@
...
+        mock_path.glob.return_value = ["1"]
+
+        find_files(["1"], files_to_exclude=[], recursive=False)
+
+        mock_path.return_value.glob.assert_called_once_with("*")
+
+
+class TestIsExtensionAllowedByFilenameFunctions(FindFileTestCase):
+    @dataclass
+    class Dependencies:
```
owencjones, line 89: This dependency injection approach used because `pytest` already has `is_allowed_by_filename` in memory from the file header, and this guarantees a new memory pointer for each test - `pytest` runs 4 in parallel as default.

tests/darwin/utils_test.py
```diff
@@ -1,4 +1,5 @@
 from unittest.mock import MagicMock, patch
```
owencjones, line 1: Formatted made changes to a file I saved in advertently.

darwin/utils.py
tomvars: Just to give visibility this line - `"".join(path.suffixes)`  is there to handle the fact that some extensions have two dots in them e.g `.nii.gz`. I'm not sure this new logic works, as `is_extension_allowed` expects only the extension passed to it not the entire path.

tomvars: Ah I see the new `is_extension_allowed_by_filename` but, maybe that should be used instead of `is_extension_allowed`

owencjones: Yeah, you're absolutely right, not sure how I missed that!

Well, I am - I _unit_ tested it, so tested the functions individually!

rslota (approved):

PR #503, opened 2022-12-07
[SWAT-647][internal] Fix bug with "push -p".

whilefalse:
# Problem

`darwin dataset push -p` was broken with v2 datasets, since v0.8.3 (it worked fine in 0.8.2).

# Solution

It looked like a simple typo in a parameter name, which meant the path wasn't getting passed through when doing the upload.


owencjones (approved):
Looks good to me, looks like a typo being removed.

Tag me in the ticket when you move to Ready for QA and I'll test it.

PR #505, opened 2022-12-14
Altered codebase to use orjson

owencjones:
Replaced usage of Python `json` standard lib, with `orjson`, a C-compiled, speed-optimised lib.

Also tested ujson, but orjson had native support for Numpy, and so worked better

# Changelog
* Added orjson to dependencies
* Replaced all uses of `json` with `orjson`


owencjones (commented):
Note that many files are also reformatted by my formatter, this happens automatically on save, so changed to lines that contain no json commands are incidental

darwin/dataset/download_manager.py
```diff
@@ -3,14 +3,14 @@
 """
```
owencjones, line 3: Many changes that appear on this diff are style changes made by black, the code formatter, not by me.  They are of note, but not relevant to the ticket in play.

darwin/dataset/download_manager.py
```diff
@@ -316,15 +366,16 @@ def _update_local_path(annotation: AnnotationFile, url, local_path):
 
     # we modify raw json, as internal representation does't store all the data
     with annotation.path.open() as file:
-        raw_annotation = json.load(file)
+        raw_annotation = json.loads(file.read())
```
owencjones, line 369: `orjson` is significantly faster, but doesn't expose a `load` method to operate on `TextWrapperIO` functions.

I did plan to write a wrapper for this to expose the same interface, but decided against it for this quick-win, low hanging fruit ticket.

darwin/dataset/download_manager.py
```diff
@@ -316,15 +366,16 @@ def _update_local_path(annotation: AnnotationFile, url, local_path):
...
 
         for slot in raw_annotation["item"]["slots"]:
             for source_file in slot["source_files"]:
                 if source_file["url"] == url:
                     source_file["local_path"] = str(local_path)
 
     with annotation.path.open(mode="w") as file:
-        json.dump(raw_annotation, file, indent=4)
+        op = json.dumps(raw_annotation, json.OPT_INDENT_2)
```
owencjones: We can't recreate 4-space indented JSON with `orjson` - it offers 2 spaces, or no spaces.  For the performance gain, this seems fair.

darwin/dataset/download_manager.py
brain-geek: Wouldn't it be better to have separate PR with formatting, to have a clearer PR?

.gitignore
```diff
@@ -161,4 +161,8 @@ cython_debug/
 # conda environment
 environment.yml
 
-.DS_Store
\ No newline at end of file
+.DS_Store
+!.vscode/settings.json
```
owencjones, line 165: Allows commiting vscode files needed without git complaint

darwin/dataset/download_manager.py
owencjones: Addressed now, the settings of the formatter have been changed at @tomvars suggestion, as most were just changing line length anyway.

Might still be one or two, but the main, they're just directly relevent to the code change.

brain-geek (approved):

.gitignore
owencjones: In reaction to a request from Tom.  The `settings.json` file has been committed for ages, and contains a standard for formatting the files on save etc.  It's already there on `master` - this just stops git from complaining that it's in a commit.

Currently it's a tracked file that is _also_ in the `.gitignore`

owencjones: RE: the `recommendations.json` file allows us to make suggestions for useful plugins for working with the codebase, but honestly, I don't feel strongly about whether we have it or not.

rslota (approved):

PR #508, opened 2022-12-19
Validation CLI

Nathanjp91:
Exposes a new CLI function

`darwin validate location`
where location is a single file or a folder to recursively search for *.json files

glob style pattern based search
`--pattern PATTERN`

will print to the console each file checked and pass/fail + list of errors found on failed files
optional flags to turn off console printing for successful files
`--silent`
and a flag to specify an output file to save the result
`--output OUTPUT`

Changes to code base:
jsonschema had to be updated to support the draft202012Validator used for darwin v2 schema which was introduced in a later version than what the repo was based on. This unfortunately changes a range of tests and needed to be updated to keep in line with the Exception changes


darwin/cli_functions.py
```diff
@@ -783,6 +789,10 @@ def dataset_import(
         _error(str(e))
     except UnrecognizableFileEncoding as e:
         _error(str(e))
+    except UnknownAnnotationFileSchema as e:
```
owencjones, line 792: 👍🏻

darwin/cli_functions.py
```diff
@@ -927,6 +937,66 @@ def delete_files(dataset_slug: str, files: List[str], skip_user_confirmation: bo
         _error(f"An error has occurred, please try again later.")
```
owencjones, line 937: Not your code, but f-string with no placeholders, could remove.

darwin/cli_functions.py
```diff
@@ -927,6 +937,66 @@ def delete_files(dataset_slug: str, files: List[str], skip_user_confirmation: bo
...
 
 
+def validate_schemas(
+    files: Optional[List[str]] = None,
+    folder: Optional[str] = None,
+    pattern: Optional[str] = None,
+    silent: bool = False,
+    output: Optional[Path] = None,
+):
```
owencjones, line 946:
```suggestion
) -> None:
```

owencjones: Makes `mypy` happier overall.

darwin/cli_functions.py
```diff
@@ -927,6 +937,66 @@ def delete_files(dataset_slug: str, files: List[str], skip_user_confirmation: bo
...
+        folder location to check for *.json files, by default None
+    pattern : Optional[str], optional
+        glob style pattern matching, by default None
+    silent : bool, optional
+        flag to set silent console printing, only showing errors, by default False
+    output : Optional[Path], optional
+        filename for saving to output, by default None
+    """
+    if not files and not folder and not pattern:
+        # TODO insert logging here once we introduce that
```
owencjones, line 964: In the meantime, this is a CLI command, so move L974 above this and
```suggestion
        console.log('--files or --folder flag requires a pattern')
```
to give the user firm output on their issue?

https://rich.readthedocs.io/en/stable/console.html#attributes - Rich has a lot of stuff for this, and it can be handy.

owencjones: I'm also pretty sure that the `argparse` parsing actually makes this condition impossible, but handling it is good to avoid someone breaking it in future.

darwin/cli_functions.py
```diff
@@ -927,6 +937,66 @@ def delete_files(dataset_slug: str, files: List[str], skip_user_confirmation: bo
...
+
+    console.print(f"Validating schemas for {len(to_validate)} files")
+
+    for file in to_validate:
+        try:
+            errors = validate_file_against_schema(file)
+        except MissingSchema as e:
+            errors = [e]
+        all_errors[str(file)] = [str(error) for error in errors]
+        if len(errors) == 0:
```
owencjones, line 988: Same as putting
```python 
if not errors:
```

darwin/exceptions.py
```diff
@@ -127,6 +134,84 @@ class RequestEntitySizeExceeded(Exception):
...
+    """
+    Used to indicate error while validation JSON annotation files.
+    """
+
+    def __init__(self, parent_error: jscValidationError, file_path: Path):
+        """
+        Parameters
+        ----------
+        parent_error: ValidationError
+            Error reported by `jsonschema`.
```
owencjones, line 166: For Sphinx docs, we need to use double backticks.

owencjones (approved):
Looks good, no required changes, except the double backticks for the Sphinx docs, the rest are just me being picky 😉

rslota (changes requested):
I'm not a fun of the `--file` / `--folder` flags tbh. It's not consistent with our other commands in CLI which detect automatically if user provided a file or a folder.
Ideally users could just do the same as they do with import, but just changing `import` to `validate` in the shell and get validation, without looking at other things they need to provide for commend to work. It feels like making user decide between file and folder is just worse UX, as it's not like the argument is ambiguous. As long it's not a pattern, we can detect if it's a folder or file. Pattern could stay more complex case, for those wanting to use it. 
What do you think @Nathanjp91 @owencjones ?

Let's also amend the PR title to include ticket code and `[extenal]` tag please 🙏 

Also, the validation doesn't feel verbose enough. For example, I made a typo in one of the fields in my 100k line JSON and I got:
Screenshot 2023-01-09 at 20 52 43
As user, I would have no clue which name this is referring to. OpenAPI is capable printing very good location of the errors and we really should include that, otherwise it's pretty hard to know which of 20k annotations contains the bug.

rslota (approved):
LGTM!

PR #510, opened 2022-12-20
IO-575: Issue with directory imports

owencjones:
Code on master was not reading directories, only files.  This fixes that, and adds unit tests to avoid it in future.


rslota (approved):

rslota:
@owencjones Let's create a ticket for this and add it to the PR title please 🙏

PR #511, opened 2023-01-09
[V7-3952] Support multi array coco segmentation import

simedw:
- support multi array segments in coco for imports
- fix typo


owencjones (approved):
LGTM.  Bit worried that it didn't break any unit tests, so I'll add a ticket to write them.

JBWilkie:
Closing in favour of #989 which accomplishes the same goal

PR #512, opened 2023-01-12
[IO-557][internal] Move to use Poetry for packaging and publishing

owencjones:
Move development management on `darwin-py` to use [Poetry](https://python-poetry.org/)

Changelog:

* Removed `setup.py`
* Removed `setup.cfg`
* Added `pyproject.toml`, and wrote proper setup for it
* Autogenerated `poetry.lock` - exact version information and hashes for every package we use
* Added `docs/README.md` for developers.
* Took opportunity to add an issue template


.github/ISSUE_TEMPLATE
```diff
@@ -0,0 +1,40 @@
+name: Bug Report
```
owencjones, line 1: Not ticket related, just useful in terms of getting better issues in

PR #517, opened 2023-01-18
Fix bug that raises KeyError exception for complex_polygon

hwaxxer:
If an annotation has `paths` and not `path` (eg `complex_polygon`), the `SemanticSegmentationDataset` will raise a `KeyError` exception. This fix iterates through all paths and appends each sequence to the annotations.


rslota (approved):
Thanks @hwaxxer for the PR! This change seems simple and effective, so let's merge it.

PR #518, opened 2023-01-19
Fix transform bugs for SemanticMask by checking for "mask" or "masks".

hwaxxer:
The current transforms do not work for `SemanticMasks` which has `mask` and not `masks`. This is a minimal change that fixes the issue for the flip transforms.


Nathanjp91 (approved):
LGTM

PR #519, opened 2023-01-19
[IO-625][external] Fix missing fields when converting from v2 to darwin v1 format

brain-geek:
This changes two things in v1 conversion:
 - dataset name is now present
 - complex polygons don't actually need to be split into `additional_paths`


rslota (approved):
LGTM!

Nathanjp91 (approved):
LGTM.

PR #521, opened 2023-01-25
[IO-543][internal] Move Imports To Multiprocessing

owencjones:
# Make importer use multiprocessing library

Importer activities all now use (optional) multiprocessing using [MPIRE](https://github.com/Slimmer-AI/mpire), a `multiprocessing`-interface compatible multiprocessing drop-in that also offers better control over threads, and better ability to hand information to threads without the pickling obstacle.

Both SDK and CLI can now be used in this way.

## Files changed:

```
.github/workflows/tests.yml
darwin/cli.py
darwin/cli_functions.py
darwin/importer/importer.py
darwin/options.py
poetry.lock
pyproject.toml
```

Corrected all pre-existing type bugs and linter issues in all files touched, too


darwin/importer/importer.py
```diff
@@ -420,8 +502,8 @@ def _get_skeleton_name(skeleton: dt.AnnotationClass) -> str:
 
 
 def _handle_subs(
-    annotation: dt.Annotation, data: Dict[str, Any], annotation_class_id: str, attributes: Dict[str, Any]
-) -> Dict[str, Any]:
+    annotation: dt.Annotation, data: Dict[str, Unknown], annotation_class_id: str, attributes: Dict[str, Unknown]
+) -> Dict[str, Unknown]:
```
tomvars, line 506: Is this related to multiprocessing? Feels like it should be in a separate PR

darwin/importer/importer.py
```diff
@@ -442,7 +524,7 @@ def _handle_subs(
     return data
 
 
-def _handle_complex_polygon(annotation: dt.Annotation, data: Dict[str, Any]) -> Dict[str, Any]:
+def _handle_complex_polygon(annotation: dt.Annotation, data: Dict[str, Unknown]) -> Dict[str, Unknown]:
```
tomvars, line 527: Same as previous comment

tests/darwin/importer/importer_test.py
```diff
@@ -0,0 +1,205 @@
...
+
+        mock_importer = MagicMock()
+        mock_importer.side_effect = ["1", "2"]
+
+        result = find_and_parse(mock_importer, [Path("example_dir")], self.mock_console, False)
+
+        mock_wp.assert_not_called()
+        mock_importer.assert_called()
+
+        self.assertEqual(result, ["1", "2"])
```
tomvars, line 135: Just double checking that the result of `find_and_parse` is guaranteed to be in the same order always?

darwin/cli_functions.py
```diff
@@ -698,7 +698,7 @@ def file_upload_callback(file_name, file_total_bytes, file_bytes_sent):
         already_existing_items = []
         other_skipped_items = []
         for item in upload_manager.blocked_items:
-            if item.reason.upper() == "ALREADY_EXISTS":
+            if (item.reason is not None) and (item.reason.upper() == "ALREADY_EXISTS"):
```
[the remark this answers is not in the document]
owencjones: No, this was static type analyser fixes - I'm not religious about them, but they do reveal issues sometimes, and did to an extent this time, so I'm basically running `mypy` against files I change, in an effort to slowly improve things without a slew of tech debt tickets.

I always make it the final commit so that if anyone wants to see the focus of the work, they can diff the commits instead.

darwin/importer/importer.py
```diff
@@ -1,17 +1,28 @@
...
     Tuple,
     Union,
 )
 
+from darwin.datatypes import AnnotationFile
+from darwin.item import DatasetItem
+
+Unknown = Any  # type: ignore
+
+from mpire import WorkerPool, tqdm
```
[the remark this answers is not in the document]
owencjones: Yeah, we're not concluded for this ourselves, but it was a useful system for managing unpickable jobs.  Nathan and I discussed it, and the main reason I stuck with it is because the refactor from MPIRE to MP is almost as simple as switching the imports - the interface is deliberately similar.

There are other better established projects, but this has great metrics for this type of task, and overcomes the pickling issues that MP can bring, so that was the logic.  The alternative was to use something like Dask or joblib which involve very different paradigms - so that's the idea behind using it.

We get good progress bars and error interception by comparison to joblib, but yeah - aware that MPIRE might need swapping, and in that case, we'd likely use mp itself.  To do that, we have to refactor some chunks of code that we foresee being replaced relatively soon, so avoided it for now.

As JSON files contain more and more masks in RLE, we'll see file sizes shoot up anyway, and only streaming the JSON would help us.

darwin/importer/importer.py
owencjones: It's housekeeping, like the other one - I'm inclined to agree, but the nature of tasks needed on darwin-py so far has meant that the approach of fixing types a file at a time works well.  I can reverse it for readability - but the likelihood is that the ticket created to do it will not see light of day for a while.

tests/darwin/importer/importer_test.py
owencjones: With `map` yes, it will be - using `imap` or `starmap` it wouldn't necessarily be.

darwin/importer/importer.py
```diff
@@ -71,42 +88,84 @@ def build_main_annotations_lookup_table(annotation_classes: List[Dict[str, Any]]
...
+        except KeyboardInterrupt:
+            maybe_console("Keyboard interrupt. Stopping.")
+            return None
+        except Exception as e:
+            maybe_console(f"Error: {e}")
+            return None
+
+    else:
+        if use_multi_cpu and not MPIRE_AVAILABLE:
+            maybe_console("Multiprocessing module not available. Falling back to single CPU")
```
tomvars, line 142: I would add a string here like, "Using single CPU for upload. Run `pip install mpire` to benefit from faster uploads!"

darwin/cli_functions.py
```diff
@@ -957,7 +959,7 @@ def delete_files(dataset_slug: str, files: List[str], skip_user_confirmation: bo
 
     except NotFound as e:
         _error(f"No dataset with name '{e.name}'")
-    except:
+    except Exception:
```
Nathanjp91, line 962: Worthwhile to pass the exception to print so that customer knows more about what's going on?

Nathanjp91 (approved):
Can't see any issues

darwin/cli_functions.py
owencjones: It's in unrelated code, and was changed to satisfy Mypy so, for now would prefer not to?

PR #522, opened 2023-01-30
Allow passing a Client to get_dataset()

hwaxxer:
This makes it possible to use this function without setting an environment variable or creating a config file. Currently, calling the function without the environment variable or config file results in an obscure message followed by "Error: Authenticate first" https://github.com/v7labs/darwin-py/blob/1abb2595da0555aef88b589e04d7d2eee2edcf3e/darwin/cli_functions.py#L1242.


owencjones (approved):
This seems like a good DI-style approach that allows both use cases to work (SDK and CLI).  Thanks to @hwaxxer for input on this.

PR #523, opened 2023-01-31
[SWAT-734][Internal] Fixing conversion to darwin_json_1.0

Fl4m3Ph03n1x:
Conversions from darwin_2.0 to darwin_1.0 are broken right now. 
This PR aims to fix that with a temporary quick fix that reverts the json encoder used.

Ticket: https://linear.app/v7labs/issue/SWAT-734/dataset-conversion-failed-for-ffss
Context:  https://vseven.slack.com/archives/C03LKKX7D0U/p1675160311013969

### DIFF
--- a/darwin/exporter/formats/darwin_1_0.py
+++ b/darwin/exporter/formats/darwin_1_0.py
@@ -1,10 +1,8 @@
+import json
 from pathlib import Path
-from typing import Any, Dict, Iterable, List, Union
-
-import orjson as json
+from typing import Any, Dict, Iterable, Union
 
 import darwin.datatypes as dt
-from darwin.exporter.formats.numpy_encoder import NumpyEncoder
 
 
 def export(annotation_files: Iterable[dt.AnnotationFile], output_dir: Path) -> None:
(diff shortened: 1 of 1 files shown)


owencjones (approved):
LGTM

darwin/exporter/formats/darwin_1_0.py
```diff
@@ -1,22 +1,23 @@
...
         _export_file(annotation_file, id, output_dir)
 
 
-def _export_file(annotation_file: dt.AnnotationFile, id: int, output_dir: Path):
+def _export_file(annotation_file: dt.AnnotationFile, id: int, output_dir: Path) -> None:
     output: Dict[str, Any] = _build_json(annotation_file)
     output_file_path: Path = (output_dir / annotation_file.filename).with_suffix(".json")
     with open(output_file_path, "w") as f:
-        op = json.dumps(output, option=json.OPT_INDENT_2 | json.OPT_SERIALIZE_NUMPY).decode("utf-8")
+        op = json.dumps(output, option=json.OPT_INDENT_2 | json.OPT_SERIALIZE_NUMPY | json.OPT_NON_STR_KEYS).decode(
```
rslota, line 18: @Fl4m3Ph03n1x Do you have example of such annotation that has non string keys? We never do that, so allowing non string keys feels like a hack, rather then solution. Maybe I'm wrong here, that's why I want to see an example. 
I don't want to block this PR if it unblock clients, but I want us to be aware of tech debt that we're creating (cc @owencjones ).

owencjones: Sorry @rslota didn't see this: have sent you an example

PR #529, opened 2023-02-06
Io 665 housekeeping for io 622

owencjones:
Housekeeping only for another ticket


brain-geek (approved):

darwin/cli_functions.py
```diff
@@ -764,6 +771,8 @@ def dataset_import(
     append: bool,
     class_prompt: bool = True,
     delete_for_empty: bool = False,
+    import_annotators: bool = False,
```
brain-geek, line 774: This is technically from other ticket

brain-geek (approved):

PR #531, opened 2023-02-07
[SWAT-753][external]: Fix 'darwin convert' to mask formats path issue

balysv:
Files that have duplicate names in different folders would not correctly convert as the directory path was not taken into account when computing the destination file name. This change ensures that we include the full darwin file path instead of just the file name.


owencjones (approved):
Good find, LGTM

PR #534, opened 2023-02-10
including release in dataset

Nathanjp91:
Added release to the dataset object. Using get_remote_dataset with a string containing a release (ie 'v7-labs/random-image:testexport1')will set the release and download that rather than latest.

### DIFF
--- a/tests/darwin/dataset/remote_dataset_test.py
+++ b/tests/darwin/dataset/remote_dataset_test.py
@@ -306,6 +306,24 @@ def files_content() -> Dict[str, Any]:
     }
 
 
+def test_dataset_creation_from_id():
+    def it_should_set_id_correctly_from_id(darwin_client: Client):
+        dataset_id = "team_slug/dataset_name:test_release"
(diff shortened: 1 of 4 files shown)


owencjones (commented):
Any way we can have a unit test that prevents this confusion of terminology happening in the same part of the code again?

Nathanjp91:
> Any way we can have a unit test that prevents this confusion of terminology happening in the same part of the code again?

Added some tests to make sure the slugs/id were set correctly in the database object but given most of the functionality is around pulling data from the server, it would be very difficult for full coverage.

tests/darwin/dataset/remote_dataset_test.py
```diff
@@ -306,6 +306,24 @@ def files_content() -> Dict[str, Any]:
     }
 
 
+def test_dataset_creation_from_id():
```
owencjones, line 309: Implementation tests only really, but I appreciate the reasoning.

owencjones (approved):
LGTM

PR #535, opened 2023-02-10
[IO-622][external] Import annotator metadata

owencjones:
# Changelog
Adds `--import-annotators` to prompt CLI to import annotators in metadata contains information about them
Adds `--import-reviewers` to prompt CLI to import reviewers in metadata contains information about them
Adds equivalent optional parameters in import functions for SDK users


rslota (approved):
Code looks like it does what it's suppose to do 👍 Let's QA and :shipit: !

rslota:
@owencjones Let's add a PR description that is more in depth, that tells more about what flags were added and how they function, so that changelog on release is more meaningful for users 🙏

owencjones:
@rslota That's now done.

PR #537, opened 2023-02-13
Io 585 incorrect type annotation

Nathanjp91:
Changes List -> Sequence for function calls for correct mypy type inference


owencjones (approved):
LGTM

PR #538, opened 2023-02-14
[IO-605][internal] Unpin NumPy version

owencjones:
Currently numpy version is pinned below or equal to `1.23.0` to prevent a breaking change in Numpy breaking upolygon.  This addresses that in relation to a fix now in PR to upolygon.

# Changelog
Updated numpy


Nathanjp91 (approved):
LGTM

PR #540, opened 2023-02-20
[IO-749][external] Added python support install on 3.7

owencjones:
# Description

Changes that allow `darwin-py` to install with PIP on python 3.7.x versions.

# Motivation

Whilst installation on Python 3.7.x and below can be problematic, we have some users who do still use this, and this change will allow them to install more easily in pipelines.

# Changelog

Changed required Python versions from `<=3.8, <3.11` to `<=3.7, <=3.11`.  Allowing install on both 3.7 and 3.11 in untested versions.
Added pipeline step that attempts to install the pip package on 3.7 and 3.11 for both windows and ubuntu
Added constraints to prevent incompatible versions installing.
Relocked poetry versions

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.11"]
+        python-version: ["3.7"]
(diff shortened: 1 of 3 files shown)


Nathanjp91 (approved):
LGTM

PR #541, opened 2023-02-20
[IO-657] convert polygons to semantic mask

Nathanjp91:
### Problem
Using the SemanticSegmentationDataset object to generate masks for image annotations discards any overlapping sections and createss a new category_id 255 to assign them to. 

### Solution
This fix sets it so that overlapping polygon regions are now prioritised in the numerical order of the class annotations. The order of the polygon in the annotations file (determined on export by the z-level in the web platform) is what is used to determine what should happen with overlapping regions. 

It should also handle edge cases where class annotations are nested between each other, like for example
(Building_1 -> Billboard -> Building_2), where part of building 2 is in front of a billboard, which is itself in front of building 1. Both buildings share the same category id and could therefore swallow the billboard annotation. 

### Changelog 
Overlapping regions in SemanticSegmentationDataset now treat overlapping regions as part of the highest order polygon in that location and do not create a new category_id of 255 for these regions.


darwin/torch/utils.py
```diff
@@ -4,12 +4,44 @@
...
+        overlapping areas will be replaced with default_overlap value, by default False
+    overlap : int, optional
+        value to replace the overlapping segments with, used in conjuction with remove_overlap, by default 255
+
+    Returns
+    -------
+    torch.Tensor
+        Flattened mask of category id's
+    """
+    if masks.shape[0] != cats.shape[0]:
```
owencjones, line 37: Feels maybe more like a `ValueError`?  If it were and `AssertionError`, you'd expect it to come from an `assert`, and you could get that with just
```assert masks.shape[0] != cats.shape[0]```

Also feels a bit like it could expand slightly to be be more fault-tolerant, given that this is SDK use code, to test that these are `torch.Tensor` on the way in (type hinting will redline that, but the function signature won't prevent you calling it with a different object at runtime)

owencjones (approved):
Looks good, I'm not sure you needed to offer the overwrite overlap option, but it does no harm, and then see other comments.

One thing you do need to do is beef up the description and title to the new standard that deployers have been mentioning, and inline with the feedback Steve gave us all on Friday.  So do that and consider comments.

darwin/torch/utils.py
Nathanjp91: My thinking here is that yeah I would have normally used an assertion, but can send in a custom message in case this ever hits a customer call. Re torch types, this function is mostly internal to the classes that we know generate torch types. I'm unclear if it would find use in the wider SDK but I'll add the checking in case it does.

almazan:
I haven't tested it but the changes look good to me. Adding an extra class with index 255 for overlapping regions can be confusing and not applicable to many implementations (torchvision semantic segmentations models, the ones we used in the past, do use index 255 for pixels that should be ignored during training). Keeping the order of annotations instead of discarding overlapping pixels makes sense as an alternative solution to this issue.

PR #542, opened 2023-02-21
[IO-753][external] Importing classes with trailing spaces fault tolerance

owencjones:
# Introduction

Made Darwin JSON annotations importer tolerant of trailing spaces introduced elsewhere in code.

# Motivation

MLE found that a dataset with trailing spaces on class names was being problematic on import.  This improves both ability to import slightly imperfect sets and to provide better error feedback when not able to tolerate a fault.

# Changelog

Added try-except blocks, handlers, and list of errors and success state to `_import_annotations`
Added iterator for `import_annotations` function which checks for success and iterates over errors that occur, without blocking imports


darwin/importer/importer.py
```diff
@@ -555,44 +560,66 @@ def _import_annotations(
...
-        annotation_class_id = remote_classes[annotation_type][annotation_class.name]
-
-        if isinstance(annotation, dt.VideoAnnotation):
-            data = annotation.get_data(
-                only_keyframes=True,
-                post_processing=lambda annotation, data: _handle_subs(
-                    annotation, _handle_complex_polygon(annotation, data), annotation_class_id, attributes
-                ),
+
+        try:
```
Nathanjp91, line 572: I know this paradigm is kind of pythonic with EAFP but I'm not a huge fan of embedded try/except statements. Not blocking by any means, but I prefer 
```
if key in remote_classes[annotation_type]
...
elif key.strip() in remote_classes[annotation_type]
```
Thoughts?

Nathanjp91 (approved):
LGTM

Issue #526, opened 2023-01-31
Why is the background class removed in LocalDataset?


hwaxxer:
I haven't debugged this fully but I believe this breaks your torch datasets. For example the `SemanticSegmentationDataset` will not contain the `__background__` class and so when the mask is constructed, the first class (not `__background__`) of the dataset will have `category_id==0`, which is the same as background.

Line in question:
https://github.com/v7labs/darwin-py/blob/master/darwin/dataset/local_dataset.py#L91

Nathanjp91:
@hwaxxer I believe this is resolved with the latest release and [PR 543](https://github.com/v7labs/darwin-py/pull/543). Closing this issue, if the problem persists let us know.

PR #543, opened 2023-02-22
[IO-727] Changes to background class for SemanticSegmentationDataset

Nathanjp91:
A fix to overcome the underlying issue of background class in a  semantic segmentation datasets being mixed in with the first user generated class. Background classes were not present in the SemanticSegmentationDataset but was intendid behaviour. Background class now gets added to the class if not present, and class functions that call on class features like compute_weights have been updated to reflect the underlying background class, which is treated differently because it's not a user defined annotation. 

### Changelog
Fix to background classes bug relating to [this bug report](https://github.com/v7labs/darwin-py/issues/526) and leading to an issue losing the first class annotation when using the SemanticSegmentationDataset object

### DIFF
--- a/darwin/torch/dataset.py
+++ b/darwin/torch/dataset.py
@@ -471,7 +471,7 @@ class SemanticSegmentationDataset(LocalDataset):
                 sequences[:] = [s for s in sequences if len(s) >= 6]
                 if not sequences:
                     continue
-                # offset the index by 1 so that masks don't get mixed with background classes
+
                 annotations.append(
                     {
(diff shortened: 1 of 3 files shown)


owencjones (approved):
Looks fine, assuming it lines up with what you heard from Jon?

owencjones (approved):
Seems fine, simple approach, simple way to achieve it, and unit test in place to ensure no recurrence. 👍🏻

almazan:
Looks good to me. By prepending `__background__` to the list of classes we make sure it gets assigned the class id 0, and the user's classes start from id 1.

PR #544, opened 2023-02-23
[IO-742][external] Correct convert of dataset with duplicate names

owencjones:
**Explanation**

Datasets with multiple files (in different directories) of the same name, e.g. `1.png` were causing overwrites of the annotation files _only on conversion to version 1 JSON_, as we used a transform of the name.

Ex.

`1.png` becomes `1.json`.

Because we store all annotation files in the same directory, this means that they overwrite each other, and you only get the last one in a conversion.

**Motivation**

Customer Reported issues and having to work around this.

**Solution**

Because V2 JSON datasets already manage the renaming of files with `_2`, `_3` etc., we were able to take the local filename instead of the original filename, and use this to transform.

Also improved typing and error handling of the converter whilst working on the files.


darwin/datatypes.py
```diff
@@ -34,7 +34,11 @@
 ErrorHandler = Callable[[int, str], None]
 
 ItemId = Union[str, int]
-JSONFreeForm = Dict[str, Any]  # type: ignore
+
+# Types that assist in handling JSON payloads
+JSONFreeForm = Dict[str, UnknownType]
```
owencjones, line 39: I simplified these whilst working in the area as they were causing issues with type hints anyway, and didn't need to be as complex.

darwin/exceptions.py
```diff
@@ -1,11 +1,85 @@
...
-from typing import List
+from typing import List, Optional
 
 from jsonschema.exceptions import ValidationError as jscValidationError
 
-from darwin.datatypes import AnnotationFileVersion
+from darwin.datatypes import AnnotationFileVersion, KeyValuePairDict, UnknownType
+
+
+class DarwinException(Exception):
```
owencjones, line 11: Creates a root exception, from which we can branch all of our exceptions.

Includes ability to encapsulate a parent exception or a list of exceptions from a multi task job

darwin/exporter/formats/darwin_1_0.py
```diff
@@ -1,34 +1,71 @@
...
+            _export_file(annotation_file, id, output_dir)
+        except Exception as e:
+            errors.append(e)
+
+    if errors:
+        raise ExportException.from_multiple_exceptions(errors)
+
+
+def _export_file(annotation_file: AnnotationFile, _: int, output_dir: Path) -> None:
+
```
owencjones, line 37: As well as the basic fix, I also made the function catch and encapsulate errors that do occur, which improves user experience if things fail

Nathanjp91 (approved):
LGTM. I like the useage of new darwin exception class. We should move all our exceptions to that

PaulWhiteQA:
QA Pass - Files with the same name that are exported using JSON 2.0 now have the correct suffix

PR #545, opened 2023-02-23
[MOD-603][external] Add annotation file item_id and annotation id.

whilefalse:
These fields are present in the darwin JSON v2 export, but they are not exposed by the relevant python types.

We need these for some upcoming work which reads darwin-py exports.


Nathanjp91 (approved):
LGTM

PR #549, opened 2023-02-28
[ANN-760][external] nifti import to multi slot item

tomvars:
Currently when a user imports a nifti file using darwin-py it will upload the annotation to the zeroth slot. This PR introduces the ability to specify which slot you are uploading to.


darwin/exporter/formats/nifti.py
```diff
@@ -185,7 +190,18 @@ def process_metadata(metadata):
                 volume_dims = volume_dims[1:]
         else:
             volume_dims = None
-    return volume_dims, pixdim, affine
+    return volume_dims, pixdim, affine, original_affine
+
+
+def process_affine(affine):
+    if isinstance(affine, str):
+        affine = np.squeeze(np.array([eval(l) for l in affine.split("\n")]))
```
owencjones, line 198: Can we limit the available locals/globals dict to `eval` to prevent a injection vectors?

owencjones (changes requested):
Am I right in saying that broadly the issue here was asynchronicity?  Hence the addition of slot names for "reuniting" imported anns afterwards?

If so, nice one, good catch.  Might be good to break down the Nifti silo a bit too.

I think mostly comments, but the eval needs careful consideration, as someone putting a wayward value in a JSON might get it to do something unexpected.

darwin/exporter/formats/nifti.py
```diff
@@ -2,8 +2,19 @@
...
+try:
+    import nibabel as nib
+    from nibabel.orientations import io_orientation, axcodes2ornt, ornt_transform
+except ImportError:
+    import_fail_string = """
+    You must install darwin-py with pip install darwin-py\[medical]
+    in order to export using nifti format
+    """
+    console.print(import_fail_string)
+    exit()
```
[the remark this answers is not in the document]
tomvars: yes agreed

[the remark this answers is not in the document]
tomvars: problem is that I use `nib` for the typing in the functions contained within this file. Which are read when loading the module

darwin/exporter/formats/nifti.py
```diff
@@ -176,7 +174,7 @@ def process_metadata(metadata):
     original_affine = process_affine(metadata.get("original_affine"))
     # If the original affine is in the medical payload of metadata then use it
     if isinstance(pixdim, str):
-        pixdim = eval(pixdim)
+        pixdim = ast.literal_eval(pixdim)
```
owencjones, line 177: 👍🏻

owencjones (approved):

darwin/importer/importer.py
```diff
@@ -144,8 +144,12 @@ def maybe_console(*args: Union[str, int, float]) -> None:
         parsed_files = list(map(importer, tqdm(files) if is_console else files))
 
     maybe_console("Finished.")
-
-    if not isinstance(parsed_files, list):
+    # Sometimes we have a list of lists of AnnotationFile, sometimes we have a list of AnnotationFile
+    # We flatten the list of lists
+    if isinstance(parsed_files, list):
+        if isinstance(parsed_files[0], list):
```
Nathanjp91, line 150: if not doing anything else with the inner conditional, can flatten this, python supports conditional short circuiting
```python
if isinstance(parsed_files, list) and isinstance(parsed_files[0], list):
    ...
```
This won't raise if parsed_files isn't indexable because the first condition will short circuit.

tomvars: This change actually causes these tests to fail for some reason:

```
FAILED tests/darwin/importer/importer_mcpu_test.py::Test_find_and_parse::test_runs_single_threaded_if_use_multi_cpu_false - Asse...
FAILED tests/darwin/importer/importer_mcpu_test.py::Test_find_and_parse::test_uses_mpire_if_use_multi_cpu_true - AssertionError:...
```

Going to merge without it, can't spend any longer on this ticket.

PR #552, opened 2023-03-03
[IO-835][external] Fix for a logic flow bug in importer

owencjones:
**Summary**

@tomvars found a bug in the flow of the importer, by which logic for normal annotations, and logic for video annotations was being applied to both types in errata.  This has been corrected, along with unit test fixes and augmentations.

NB: A ticket is raised for refactoring the importer unit tests which have become untidy as a result of merging and refactoring in several areas: IO-841

# Changelog

* Refactored the importer to correctly separate flow of logic
* Refactored the tests to cover this.

### DIFF
--- a/darwin/importer/importer.py
+++ b/darwin/importer/importer.py
@@ -578,7 +578,7 @@ def _handle_annotators(annotation: dt.Annotation, import_annotators: bool) -> Li
     return []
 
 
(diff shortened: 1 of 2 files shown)


tomvars (approved):
Can't comment on the test update but the improved logic looks correct to me, I implemented something similar locally and it worked.

PR #554, opened 2023-03-07
[IO-437] Handling Uncaught import exceptions

Nathanjp91:
A range of exceptions can be thrown by the import pipeline code but are never handled. This can cause an issue on importing annotations in a range of scenarios. These errors are now captured and passed along to the user as console output. 

### Changelog
Capturing exceptions on imports to inform user on problems


darwin/importer/importer.py
```diff
@@ -491,6 +501,12 @@ def import_annotations(
                     import_annotators,
                     import_reviewers,
                 )
+                if potential_error:
+                    errors.append((parsed_file.full_path, potential_error))
+        if errors:
+            console.print(f"Encountered errors on {len(errors)} imports", style="error")
```
owencjones, line 507: This isn't your doing, but we need to improve our handling of `Console` objects, as they are only present for CLI code, not SDK.  The dataset carries a `console` property that allows us to tell whether this is an intereactive session or not, and we should probably move to using that at top end of the function.

Not something for now unless you want to do it.

owencjones (approved):
LGTM as long as there aren't any merge conflicts left

owencjones:
@Nathanjp91 Is this tested?  If so, let's merge

Nathanjp91:
> @Nathanjp91 Is this tested? If so, let's merge

Unclear on status of this actually, as per discussion on linear ticket it's closed as CNR. Should it still be merged?

owencjones:
> > @Nathanjp91 Is this tested? If so, let's merge
> 
> Unclear on status of this actually, as per discussion on linear ticket it's closed as CNR. Should it still be merged?

My take on that, re: latest input from other users, is that we should actually overhaul this whole area for good output.  Any thoughts?

owencjones:
Closing this as stale.

PR #556, opened 2023-03-10
[IO-874] removing annotation data as input to _get_annotation_data

Nathanjp91:
Removing the fourth input to _get_annotation_data as it's unnecessary and causes an issue with videoAnnotations.

### Changelog
fixed bug with video annotations import


owencjones (approved):
LGTM

PR #557, opened 2023-03-13
[IO-686] working darwin_demo

Nathanjp91:
Update the darwin_demo to a working state, changing from functions that no longer exist or were deprecated at some point in the past. 

### Changelog
Update demo code to a working state.


owencjones (approved):
Looks good, be good to have the link working again.

PR #558, opened 2023-03-20
[mod-720][Internal] Count the number of files that are actually downloaded.

brentgracey:
Info level logging to provide insight into large downloads.

### DIFF
--- a/darwin/dataset/remote_dataset.py
+++ b/darwin/dataset/remote_dataset.py
@@ -235,6 +235,9 @@ class RemoteDataset(ABC):
         ValueError
             If darwin in unable to get ``Team`` configuration.
         """
+
+        console = Console()
+
         if release is None:
             release = self.get_release()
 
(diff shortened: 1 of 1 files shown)


darwin/dataset/remote_dataset.py
```diff
@@ -235,6 +235,9 @@ def pull(
         ValueError
             If darwin in unable to get ``Team`` configuration.
         """
+
+        console = Console()
```
owencjones, line 239: See L120, there's a console built into this, which allows the user of these classes to define various console params, and although not used much _theoretically_ this is for allowing capturing of output to different forms etc.

```suggestion
        console = self.console or Console()
```

darwin/dataset/remote_dataset.py
```diff
@@ -329,7 +332,12 @@ def pull(
             if env_max_workers and int(env_max_workers) > 0:
                 max_workers = int(env_max_workers)
 
+            console.print(f"Going to download {str(count)} files to {self.local_images_path.as_posix()}.")
```
owencjones, line 335: I'd suggest _maybe_ these are good candidates for `console.log`?

owencjones (changes requested):
Looks good, decent improvement, and I'm requesting changes partly because of the first comment, but mainly because it's going to need @Nathanjp91's test fixes

owencjones (approved):
Good stuff, thanks @brentgracey

PR #563, opened 2023-03-23
[IO-901] Capturing and printing download errors

Nathanjp91:
Downloading of images would fail silently in both single and multi-threaded setups, causing issues when downloading large datasets and tokens expired, but would also fail generally for all other raised exceptions. Capturing these errors and printing to console to now provide a better UX and prevent blocking of models team when download fails

### Changelog
Errors downloading dataset items now captured and printed to console

### DIFF
--- a/darwin/dataset/remote_dataset.py
+++ b/darwin/dataset/remote_dataset.py
@@ -335,7 +335,7 @@ class RemoteDataset(ABC):
             if errors:
                 self.console.print(f"Encountered errors downloading {len(errors)} files")
             for error in errors:
-                self.console.print(f"\t - {str(error)}")
+                self.console.print(f"\t - {error}")
             return None, count
         else:
             return progress, count
--- a/darwin/dataset/remote_dataset.py
+++ b/darwin/dataset/remote_dataset.py
@@ -329,7 +329,13 @@ class RemoteDataset(ABC):
             if env_max_workers and int(env_max_workers) > 0:
                 max_workers = int(env_max_workers)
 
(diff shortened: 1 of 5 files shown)


brentgracey (approved):

darwin/dataset/remote_dataset.py
```diff
@@ -333,9 +333,15 @@ def pull(
                 max_workers = int(env_max_workers)
 
             console.print(f"Going to download {str(count)} files to {self.local_images_path.as_posix()}.")
-            exhaust_generator(progress=progress(), count=count, multi_threaded=multi_threaded, worker_count=max_workers)
+            successes, errors = exhaust_generator(
```
owencjones, line 336: Basically a monad like structure then in the end, like it

owencjones: I'm not sure if I'd rename the function overall now, but let's leave it for a less urgent time.

owencjones (approved):
LGTM, let's keep this stuff in mind for V2

darwin/dataset/remote_dataset.py
Nathanjp91: It does exhaust a generator and return values, so it is doing what it says on the box. We just happen to only ever use it to trigger downloads and the function was originally written such that the generator has to return a callable function, I haven't written any tests or checked whether the change I made to 'f(x)' will result in not requiring callable restriction because honestly, it didn't seem worth it given we only ever use it for this one purpose.

PR #564, opened 2023-03-24
Convert Darwin to Yolov8 format

kevinrigot:
Hi,
This will convert to Yolov8 format.
It will just create the dataset.yaml and the label files directly in the train dir.

To be complete, it should also copy the images and download the video sections. I'm not sure how to proceed as the converters seems to only converts label files.
It doesnt perform the split between train, valid and test. It should probably be manager in v7 directly?

### DIFF
--- a/darwin/exporter/formats/__init__.py
+++ b/darwin/exporter/formats/__init__.py
@@ -13,5 +13,6 @@ supported_formats: List[str] = [
     "semantic_mask_grey",
     "semantic_mask_index",
     "yolo",
+    "yolov8",
     "nifti",
 ]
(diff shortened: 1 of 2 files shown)


kevinrigot:
For the "copy the images"; I'm doing it manually at the moment.
For the split, I added a commit where I use the remote path if it is a folder "train", "valid" or "test". If it is something else, I push it to train.
I saw the "split" function, It's creating a txt file with the filenames that should go in the different split. Is this persisted, or just local in my computer? In other words, can I trust this split for incremental model trainings, or the validation list of files will always be different and wont provide me the same result everytime

PR #566, opened 2023-03-27
[V7-4587] [internal] Temporarily disabled multiprocessing in SDK unless explicitly set on

owencjones:
# Motivation

Ticket for CS points out that some customers experience issues with multiprocessing, as although disabled by default in CLI, it isn't in SDK.

# Solution

Have changed to be automatically not used unless user specifies the `use_multi_cpu` setting when invoking `import_annotations` and sets to `True`.  This means some imports may run slower, but the issues CS have seen will not happen.


Nathanjp91 (approved):
LGTM

PR #569, opened 2023-03-28
[DO NOT DELETE] [MOD-667][internal] Support creating benchmarks runs

whilefalse:
# Problem

We need to support starting a benchmarks run from darwin-py.

# Solution

A new command, which can be used like this:

```bash
darwin benchmarks run --target v7/darwin-py-benchmarks-test --name FooBar --format darwin /path/to/annotations
```


balysv (approved):

darwin/cli.py
```diff
@@ -167,6 +167,10 @@ def _run(args: Namespace, parser: ArgumentParser) -> None:
                 args.w,
                 args.h,
             )
+    elif args.command == "benchmarks":
```
owencjones, line 170: Is there to be any other benchmarks command - if not `benchmarks-run` might be more appropriate, but if it's just future proofing, np.

darwin/importer/importer.py
```diff
@@ -52,6 +54,16 @@
 """
 
 
+@dataclass
```
owencjones, line 57: Similar to something we had in mind anyway, so handy to have, cheers

tests/darwin/cli_functions_test.py
```diff
@@ -189,10 +204,170 @@ def error_mock():
...
-
                         delete_files(dataset_identifier, ["one.jpg", "two.jpg"], True)
 
                         get_remote_dataset_mock.assert_called_once_with(dataset_identifier=dataset_identifier)
                         fetch_remote_files_mock.assert_called_once_with({"filenames": ["one.jpg", "two.jpg"]})
                         mock.assert_called_once()
                         exception.assert_called_once_with(1)
+
+
+def describe_benchmarks():
```
owencjones, line 215: Good stuff, not enough things get tests added here.

owencjones (approved):
LGTM, liaise with QA about how this affects main use of importer mainly - feel free to point people my way

darwin/cli.py
whilefalse: I don't think there are any immediate plans, but I can imagine something like `benchmarks status <id>` or `benchmarks get-results <id>`.

PR #570, opened 2023-03-31
Update coco.py to support complex polygons

fgraffitti-cyberhawk:
I've updated the coco.py script to support import of complex polygons, according to the guide in:
https://docs.v7labs.com/docs/importing-complex-polygons


JBWilkie:
Support for this was added in #989

PR #575, opened 2023-04-13
[IO-751] api client v2

Nathanjp91:
Adding in the core api client functionality that can make generic http requets. Will be built on for the meta apiclient that is presented to the user and acts as a core base. 
- Generic call method that is invoked by oither calls, get/put/post etc
- Checks for known exceptions
### Changelog
Continuing work on darwin py v2 functionality. Added Core Client

### DIFF
--- a/darwin/future/core/client.py
+++ b/darwin/future/core/client.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
-
+from urlp
 from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Callable, Dict, List, Optional, Union, overload
+from urllib.parse import urlparse
 
 import requests
 from pydantic import BaseModel, validator
@@ -34,8 +35,9 @@ class Config(BaseModel):
         v = v.strip()
         if not v.endswith("/"):
             v += "/"
-        assert v.startswith("http") or v.startswith("https"), "base_url must start with http or https"
-        assert v.count("/") >= 3, "base_url must contain at least 3 slashes"
+        check = urlparse(v)
+        assert check.scheme in {"http", "https"}, "base_url must start with http or https"
+        assert check.netloc, "base_url must contain a domain"
         return v
 
     def from_env(cls) -> Config:
@@ -154,7 +156,7 @@ class Client:
         ...
 
     def _generic_call(self, method: Callable, endpoint: str, payload: Optional[dict] = None) -> dict:
-        endpoint = self._validate_endpoint(endpoint)
+        endpoint = self._sanitize_endpoint(endpoint)
         url = self.config.base_url + endpoint
         if payload is not None:
             response = method(url, payload)
@@ -184,12 +186,12 @@ class Client:
     def patch(self, endpoint: str, data: dict) -> dict:
         return self._generic_call(self.session.patch, endpoint, data)
 
-    def _validate_endpoint(self, endpoint: str) -> str:
+    def _sanitize_endpoint(self, endpoint: str) -> str:
         return endpoint.strip().strip("/")
 
 
 def raise_for_darwin_exception(response: requests.Response) -> None:
-    """Raises an exception if the response is not 200
+    """Raises an exception if the response.status_code matches a known darwin error
 
     Parameters
     ----------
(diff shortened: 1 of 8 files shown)


darwin/future/core/client.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+
+    api_key: Optional[str]
+    base_url: str
+    default_team: Optional[Team]
+
+    @validator("base_url")
+    def validate_base_url(cls, v: str) -> str:
+        v = v.strip()
+        if not v.endswith("/"):
```
owencjones, line 35: I see what you mean about Pydantic validators here, they are a bit wordy at times, and the mutation of the value in a validator feels almost kind of dirty.

No change suggested, just WRT prev convo.

darwin/future/core/client.py
```diff
@@ -0,0 +1,203 @@
...
+    def post(self, endpoint: str, data: dict) -> dict:
+        return self._generic_call(self.session.post, endpoint, data)
+
+    def delete(self, endpoint: str) -> dict:
+        return self._generic_call(self.session.delete, endpoint)
+
+    def patch(self, endpoint: str, data: dict) -> dict:
+        return self._generic_call(self.session.patch, endpoint, data)
+
+    def _validate_endpoint(self, endpoint: str) -> str:
```
owencjones, line 187: Naming of this one feels off, it's less validating, and more preparing/sanitising/somethingising.  Not a huge deal, consideration point.

darwin/future/core/client.py
```diff
@@ -0,0 +1,203 @@
...
+        return endpoint.strip().strip("/")
+
+
+def raise_for_darwin_exception(response: requests.Response) -> None:
+    """Raises an exception if the response is not 200
```
owencjones, line 192: This doesn't raise if the error is in a 500 range, or 318 etc.?  Is that deliberate?

pyproject.toml
```diff
@@ -72,12 +72,19 @@ profile = "black"
 plugins = [
     "pydantic.mypy"
 ]
+
+follow_imports = "silent"
+warn_redundant_casts = true
+warn_unused_ignores = true
+# disallow_any_generics = true
```
owencjones, line 79: Just remove the line with this one I think

owencjones (approved):
Yep, good stuff, few comments to look at.

darwin/future/core/client.py
Nathanjp91: You can seperate them into seperate functions and they become a one line call, I use that elsewhere. Mutation within the validator is the pydantic way, they combine both validation and conversion, if the function throws it's invalid and if it doesn't you mutate into the format you want.

darwin/future/core/client.py
Nathanjp91: It's a good catch on the documentation, it should instead state that it raises for known darwin exceptions rather than any non-200 code. The actual client function that uses it raises_for_status() after checking darwin codes, it'll catch every other code we don't have specific exceptions listed for.

darwin/future/core/client.py
```diff
@@ -0,0 +1,203 @@
...
+    api_key: Optional[str]
+    base_url: str
+    default_team: Optional[Team]
+
+    @validator("base_url")
+    def validate_base_url(cls, v: str) -> str:
+        v = v.strip()
+        if not v.endswith("/"):
+            v += "/"
+        assert v.startswith("http") or v.startswith("https"), "base_url must start with http or https"
```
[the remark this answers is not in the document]
Nathanjp91: Good point on http vs https. Should have caught that myself. 
In terms of urlib there's actually not a validator in that. Closest is a parser that is essentially doing what I'm doing here anyway although a bit more wordily and requiring an import. 

```
from urllib.parse import urlparse
url = urlparse("scheme://netloc/path;parameters?query#fragment")
assert url.scheme
assert url.netloc
assert url.path
```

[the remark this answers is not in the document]
owencjones: Yeah, I was thinking of `urlparser` - it not necessary obv.

darwin/future/core/client.py
owencjones: 👍🏻

PR #578, opened 2023-04-27
[Io 973] get_team DarwinV2

Nathanjp91:
Creation of the get_team method for v2 including tests and config/client changes to accomodate intendid uses

### Changelog
Darwin-py v2 futures get_team method


darwin/future/core/client.py
```diff
@@ -2,21 +2,27 @@
...
+JSONType = Dict[str, Any]  # type: ignore
 
 
-class Config(BaseModel):
+class TeamsConfig(BaseModel):
+    api_key: str
+    datasets_dir: Path
+
+
+class DarwinConfig(BaseModel):
```
Nathanjp91, line 25: Had to change from Config -> DarwinConfig here because pydantic models actually have an internal Config class which sets metaproperties and it causes mypy issues on functions like local() which return the 'Config' class. It can't figure out if you mean the base class or the internal class.

darwin/future/core/client.py
```diff
@@ -40,11 +49,42 @@ def validate_base_url(cls, v: str) -> str:
...
+        if path.suffix.lower() == ".yaml":
+            data = DarwinConfig._parse_yaml(path)
+            return DarwinConfig.parse_obj(data)
+        else:
+            return DarwinConfig.parse_file(path)
 
-    def from_file(cls, path: Path) -> Config:
-        pass
+    @staticmethod
+    def _parse_yaml(path: Path) -> dict:
```
Nathanjp91, line 84: as above, this will eventually be replaced

darwin/future/tests/fixtures.py
```diff
@@ -0,0 +1,15 @@
...
+import shutil
+from pathlib import Path
+from typing import Generator
+
+import pytest
+
+
+@pytest.fixture
+def test_directory() -> Generator[Path, None, None]:
+    path = Path.home() / "darwin-tests"
```
Nathanjp91, line 11: not sure why, but I couldn't get tests to create a hidden folder, ".darwin-tests" here, with permission denied exceptions. I think a mac thing. Non-hidden folders were fine.

PR #580, opened 2023-04-28
[IO-976][internal] Fix for bug in image naming

owencjones:
Addresses urgent bug, in which file names can clash when files are in different folders.  This adds numbering from JSON to images.


Nathanjp91 (approved):
LGTM

PR #581, opened 2023-04-28
Made numpy typing optional to support old versions of numpy

owencjones:
Small fix that unblocks databricks envs using darwin py.


Nathanjp91 (approved):
Good workaround, would be nice if databricks supported older numpy though but not in our control.

owencjones:
It would, but they're behind, quite significantly, and not moving.

PR #582, opened 2023-05-02
[io-952][internal] Break data objects out into separate files

owencjones:
Small refactoring to avoid circular deps, and create atomic deps.


darwin/future/data_objects/validators.py
```diff
@@ -1,3 +1,16 @@
...
+    ----------
+    name : str
+        The name to be parsed and validated
+
+    Returns
+    -------
+    str
+        The parsed and validated name
+    """
     assert isinstance(name, str)
```
Nathanjp91, line 15: I know this is actually my code but was written when first learning pydantic. Don't actually need these type assertions as pydantic does it already

Nathanjp91 (approved):
LGTM. Do agree they need to be broken out and I never liked the darwin_meta name I chose.

PR #591, opened 2023-05-09
[IO-1009][Internal] V2 Create dataset

owencjones:
IO-952 subticket for creation of datasets.


darwin/future/core/datasets/create_dataset.py
```diff
@@ -0,0 +1,37 @@
...
+
+    Returns
+    -------
+    Dataset
+        The created dataset
+
+    Raises
+    ------
+    HTTPError
+        Any HTTP errors returned by the API
```
Nathanjp91, line 28: It'll also raise validation from pydantic if the returned HTTP objects aren't valid, not sure if we want to mention that

Nathanjp91 (changes requested):
Check other PR for other changes might need to make.

darwin/future/core/datasets/create_dataset.py
owencjones: Good point


> Check other PR for other changes might need to make.

Will do, these are only really "consilidation" PRs, an attempt to avoid the massive PRs, but tbh, I don't think the strategy is going well.  I need to come up with a better gitflow.

owencjones:
Have made this part of #593

PR #592, opened 2023-05-09
[Io 764] TeamMemberQuery

Nathanjp91:
Adding in functionality for the TeamMemberQuery proposed functionality
Also fleshing out Query objects and changes to inherited methods and typing. Example of intendid workflow below
```python
members = TeamMemberQuery().where({"name": "role", "value": "owner"}).collect(client)
```
### Changelog
TeamMemberQuery proposed meta interaction method added


darwin/future/core/client.py
```diff
@@ -123,33 +121,6 @@ class Page(BaseModel):
     detail: PageDetail
 
 
-class Cursor(ABC):
```
owencjones, line 126: Yeah, good call, let's keep these out, unless we come up with a use case that is improved by them

darwin/future/core/types/query.py
```diff
@@ -1,10 +1,24 @@
...
+Param = Dict[str, Any]  # type: ignore
+
+
+class Modifiers(Enum):
+    GREATER_THAN = ">"
+    LESS_THAN = "<"
+    GREATER_EQUAL = ">="
+    LESS_EQUAL = "<="
+    NOT_EQUAL = "!="
+    CONTAINS = "contains"
```
owencjones, line 20: I'm assuming we don't need `NOT_CONTAINS` for now at least.  Do you think there are any others we do need for now.

darwin/future/core/types/query.py
```diff
@@ -1,10 +1,24 @@
...
 
-from darwin.future.data_objects.team import Team
+from darwin.future.core.client import Client
 from darwin.future.pydantic_base import DefaultDarwin
 
+T = TypeVar("T", bound=DefaultDarwin)
+Param = Dict[str, Any]  # type: ignore
+
+
+class Modifiers(Enum):
```
owencjones, line 14: This feels better singular?  Minor point, won't kill not to change.

darwin/future/core/types/query.py
```diff
@@ -13,40 +27,64 @@ class QueryFilter(DefaultDarwin):
...
+        elif self.modifier == Modifiers.NOT_EQUAL:
+            return attr != param
+        elif self.modifier == Modifiers.CONTAINS:
+            return param in attr
+        else:
+            raise ValueError(f"Unknown modifier {self.modifier}")
 
 
-class Query:
+class Query(Generic[T], ABC):
```
owencjones, line 58: Gotta love them Generic TypeVars, never cause any trouble with mypy. 😉

darwin/future/tests/core/test_query.py
```diff
@@ -47,9 +57,42 @@ def test_query_filter_functionality(basic_filters: List[Query.QueryFilter], test
...
+def test_query_iterable(basic_filters: List[Query.QueryFilter], non_abc_query: Type[Query.Query]) -> None:
+    q = non_abc_query(basic_filters)
     for i, f in enumerate(q):
         assert f == basic_filters[i]
     assert q.filters is not None
     assert len(q) == len(basic_filters)
+
+
+@pytest.mark.parametrize(
+    "mod,param,check,expected",
```
owencjones, line 69: Have you seen hypothesis btw - testing assertions set that may be useful for this type of test.

owencjones (approved):
Yeah, looking good man.  Sorry my refactor made it a PITA!

darwin/future/core/types/query.py
Nathanjp91: There could be others that we come up with but shouldn't be too hard to include when we do think of them.

darwin/future/core/types/query.py
Nathanjp91: getting this typing to work with mpyp was a major PITA actually and huge props to chatgpt for figuring out where I was going wrong.

PR #593, opened 2023-05-10
[IO-1011][internal] Retrieve dataset functions

owencjones:
Sub ticket to address R of CRUD.


darwin/future/core/datasets/list_datasets.py
```diff
@@ -0,0 +1,38 @@
...
+    Raises
+    ------
+    HTTPError
+        Any errors that occurred while making the request
+    ValidationError
+        Any errors that occurred while parsing the response
+
+    """
+    response = (
+        api_client.get("/datasets", QueryString({"team": str(team_slug)})) if team_slug else api_client.get("/datasets")
```
rslota, line 35: Is it even possible to get dataset for given team from out API? I'm pretty sure our backend ignores that and always returns datasets from team, that API_KEY is authorized for (one key = one team).

darwin/future/core/datasets/get_dataset.py
```diff
@@ -0,0 +1,31 @@
...
+    Returns
+    -------
+    Dataset
+
+    Raises
+    ------
+    HTTPError
+        Any errors that occurred while making the request
+    """
+    response = api_client.get("/datasets", QueryString({"team": TeamSlug(dataset_id)}))
```
rslota, line 29: Pretty sure this wasn't suppose to be `TeamSlug(dataset_id)` ;)

darwin/future/core/datasets/list_datasets.py
rslota: So, I think that choice of team happens when creating api_client, and then we can ofc filter datasets by name etc (locally because that API doesn't have actual usable filters on its own).

darwin/future/tests/core/datasets/fixtures.py
```diff
@@ -0,0 +1,49 @@
...
+        "releases": [],
+    }
+
+    return mock_client
+
+
+@fixture
+def happy_get_client() -> Client:
+    mock_client = MagicMock(Client)
+    mock_client.get.return_value = [
```
Nathanjp91, line 35: I've been using the responses module for this, maybe unify our approaches either way?

darwin/future/core/datasets/list_datasets.py
```diff
@@ -0,0 +1,38 @@
...
+    )
+
+    return parse_obj_as(DatasetList, response)
```
Nathanjp91, line 38: is parse_obj_as and DatasetLists.parse_obj functionally the same or is there some difference? I've been using the latter. If they're the same we should pick one and keep it standard.

darwin/future/core/types/common.py
```diff
@@ -0,0 +1,27 @@
...
+from darwin.future.data_objects import validators as darwin_validators
+
+
+class TeamSlug(ConstrainedStr):
+    """Team slug type"""
+
+    min_length = 1
+    max_length = 256
+
+    validator = darwin_validators.parse_name
```
Nathanjp91, line 14: I actually didn't know you could extend the pydantic subtypes like this, but I don't think this is working the way you intend it to? I can't find docs on extending the class on pydantic page but drilling down into ConstrainedStr in pydantic lib, it's got no `validator` function.

It does have a classmethod `validate` and an internal function `__get_validators__` which yield a bunch of different functions in a chain, including `validate`. 

I could be totally wrong though as I haven't played with extending pydantics types and it could be doing other things under the hood.

Nathanjp91: https://docs.pydantic.dev/latest/usage/types/#custom-data-types

Yeah, see this page on extending str.

darwin/future/core/datasets/get_dataset.py
owencjones: Fair point

darwin/future/tests/core/datasets/fixtures.py
owencjones: Yeah, sure, got a good link to look at?

Nathanjp91 (changes requested):
I think mostly good, need to review some things though.

darwin/future/core/datasets/list_datasets.py
owencjones: Will review.

darwin/future/tests/core/datasets/fixtures.py
Nathanjp91: https://github.com/v7labs/darwin-py/blob/090da8d105e4fdb4ab4411510d9448d243c48fb2/darwin/future/tests/core/test_client.py#L59

Can mock the return value and endpoints. There's a bunch of other assertions it gives access to as well but I haven't used too many yet.

owencjones: But the `Client` doesn't return a `Response` object does it?  I thought that was returning `JSONType`?

darwin/future/core/datasets/list_datasets.py
owencjones: `DatasetLists` doesn't have a `parse` method in this instance, because its def is just (roughly)

```python
DatasetList = List[Dataset]
```

So it's not inherently pydantic-y.  `parse_obj_as` is allowing us to co-erce a list of a pydantic type into validator iteratively, essentially - that said, this PR is pre-test, so I'm not convinced of the approach.

darwin/future/tests/core/datasets/fixtures.py
Nathanjp91: yeah it responds with `JSONType` but the response module you can tell it what an endpoint returns

so in the test below that uses it, might look something like 
```python
with responses.RequestsMock() as rsps:
       endpoint = base_client.config.api_endpoint + "datasets"
       with responses.RequestsMock() as rsps:
        rsps.add(responses.POST, endpoint, json=mock_json_result, status=200)
        dataset = create_dataset(client, mock_json_result["name"])
        assert dataset.name == mock_json_result["name"]
```

darwin/future/core/datasets/list_datasets.py
Nathanjp91: ah right I see the issue. I haven't created types for mine, I just leave them as List[DarwinObject] in the code

darwin/future/tests/core/datasets/fixtures.py
owencjones: Ah I'll take a look, cheers

darwin/future/core/types/common.py
owencjones: Yeah, so to be fair, maybe this isn't the perfect way to do this, I'm kind of on the fence tbh.  It's one I've used before, and it works - I haven't tested this one yet - but I've done the same in the past.

The `__get_validators__` dunder yields all the validators one at a time, and then adds the `self.validate` method to the yield list.

This overwrites this, with your validator, which only has to return `True` or `False` to validate the thing.

One possibility is to have an alternate like this:

```python
@classmethod
def validate(self) -> bool:
    self.super().validate and darwin_validators.parse_name
```

If it's too hacky, what would you do instead?  I can assign a normal `constr`, but not with a validator, and you've written one.

darwin/future/core/datasets/get_dataset.py
owencjones: Addressed now

darwin/future/core/datasets/list_datasets.py
owencjones: Have removed.

darwin/future/core/types/common.py
owencjones: Have committed a new version, using an init dunder to achieve the same on creation.

owencjones:
@Nathanjp91 I'll get the requests mock in, but as a strategy if I do that, and then get this one across the line, that addresses your considerations for most of the other PRs

darwin/future/core/types/common.py
```diff
@@ -5,14 +5,27 @@
...
     max_length = 256
 
-    def __new__(self) -> "TeamSlug":
-        return super().__new__(darwin_validators.parse_name(super().__str__))
+    @classmethod
+    def __get_validators__(cls): #type: ignore
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, v: str) -> "TeamSlug":
```
Nathanjp91, line 19: I think I like this approach better. It's a bit more verbose but shouldn't be too many instances where we have to extend base pydantic objects like this

owencjones: Yeah, I agree.  It's just a bit weird that it's not extended from BaseModel like usual!

darwin/future/core/types/common.py
```diff
@@ -0,0 +1,39 @@
...
+        if not isinstance(v, str):
+            raise TypeError("string required")
+        modified_value = darwin_validators.parse_name(v)
+        return cls(modified_value)
+
+    def __repr__(self) -> str:
+        return f"TeamSlug({super().__repr__()})"
+
+
+class QueryString:
```
Nathanjp91, line 29: Kind of a much of a muchness with this, but thoughts on making this a basemodel and using a validtor instead? To keep it in line with our other code format

```python
class QueryString(BaseModel):
    value: Dict[str, str] # will provide mypy type hinting here and validation, but not ValidationError on passing a Dict[Any, Any]
    @pydantic.validator('value')
    def dict_check(cls, v) -> Dict[str, str]:
        assert all(isinstance(k, str) and isinstance(v, str) for k, v in value.items())
        return v

File: darwin/future/tests/core/datasets/test_create_dataset.py [hunk 8]
```diff
@@ -0,0 +1,43 @@
...
+    with raises(HTTPError):
+        with responses.RequestsMock() as rsps:
+            rsps.add(
+                rsps.POST,
+                base_client.config.api_endpoint + "datasets",
+                json=basic_dataset,
+                status=400,
+            )
+
+            create_dataset(base_client, "test-dataset")
```
Nathanjp91, line 40: are these meant to be doubled?

darwin/future/tests/core/datasets/test_get_dataset.py
```diff
@@ -0,0 +1,45 @@
...
+        )
+        with raises(HTTPError):
+            get_dataset(base_client, "test-dataset")
+            get_dataset(base_client, "test-dataset")
+
+
+def test_it_raises_an_error_on_pydantic_error(sad_client_pydantic: Client) -> None:
+    with raises(ValidationError):
+        get_dataset(sad_client_pydantic, "test-dataset")
+        get_dataset(sad_client_pydantic, "test-dataset")
```
Nathanjp91, line 45: as above

darwin/future/core/types/common.py
owencjones: Yeah, I like it

Nathanjp91 (approved):
LGTM

PR #595, opened 2023-05-10
[IO-1072] fix for image extensions

Nathanjp91:
Fixes a bug on importing jpeg images in LocalDataset object

### Changelog
jpeg images now importing with localdataset object correctly


darwin/dataset/local_dataset.py
```diff
@@ -10,7 +10,7 @@
 from darwin.utils import SUPPORTED_IMAGE_EXTENSIONS, attempt_decode, parse_darwin_json
 
 
-class LocalDataset(object):
+class LocalDataset:
```
owencjones, line 13: I keep finding these in old code too

test.py
```diff
@@ -1,25 +1,6 @@
-from typing import Optional, Union
```
owencjones, line 1: Feels like this has crept in in a larger PR, and probably actually needs a `git rm`?

owencjones (changes requested):
Check the `test.py` file, remove or justify

test.py
Nathanjp91: I've actually `git rm`'ed it before but it keeps coming back for some reason. It's in the gitignore

owencjones (approved):
LGTM

NPatel1980:
Reviewed the fix with Nathan, QA passed

PR #597, opened 2023-05-11
[Io-759] Meta Api Client

Nathanjp91:
Creation of meta api client object for use with higher level and user centric interaction with darwin objects

### Changelog
Introduction of MetaClient object and associated functionality

### DIFF
--- a/darwin/future/core/client.py
+++ b/darwin/future/core/client.py
@@ -93,34 +93,6 @@ class Result(BaseModel):
     ...
 
 
-class PageDetail(BaseModel):
-    """Page details model for managing pagination
-
-    Attributes
-    ----------
-    count: int, current position
-    next: Optional[str], url for the next page
-    previous: Optional[str], url for the previous page
-    """
-
-    count: int
-    next: Optional[str]
-    previous: Optional[str]
-
-
-class Page(BaseModel):
-    """Page of results
-
-    Attributes
-    ----------
-    results: List[Result], list of results
-    detail: PageDetail, details about the page
-    """
-
-    results: List[Result]
-    detail: PageDetail
-
-
 class Client:
     """Client Object to manage and make requests to the Darwin API
     Attributes
--- a/darwin/future/core/client.py
+++ b/darwin/future/core/client.py
@@ -66,8 +66,28 @@ class DarwinConfig(BaseModel):
         return values
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


darwin/future/core/client.py
```diff
@@ -93,34 +124,6 @@ class Result(BaseModel):
     ...
 
 
-class PageDetail(BaseModel):
```
Nathanjp91, line 96: Don't need these if not using cursors, remove for cleaner codebase

darwin/future/meta/client.py
```diff
@@ -0,0 +1,27 @@
...
+    def __init__(self, config: DarwinConfig, retries: Optional[Retry] = None) -> None:
+        super().__init__(config, retries=retries)
+
+    @classmethod
+    def local(cls) -> Client:
+        config = DarwinConfig.local()
+        return MetaClient(config)
+
+    @classmethod
+    def from_api_key(cls, api_key: str, datasets_dir: Optional[Path] = None) -> Client:
```
owencjones, line 19: Function sig says this returns client, but it returns `MetaClient` - which obv is a subclass anyway, but any reason not to just say `MetaClient` as a return?

darwin/future/meta/client.py
```diff
@@ -0,0 +1,27 @@
...
+from requests.adapters import Retry
+
+from darwin.future.core.client import Client, DarwinConfig
+
+
+class MetaClient(Client):
+    def __init__(self, config: DarwinConfig, retries: Optional[Retry] = None) -> None:
+        super().__init__(config, retries=retries)
+
+    @classmethod
```
owencjones, line 13: Same as other comment

darwin/future/meta/client.py
```diff
@@ -0,0 +1,27 @@
...
+    def local(cls) -> Client:
+        config = DarwinConfig.local()
+        return MetaClient(config)
```
owencjones, line 16: And any reason these are `return MetaClient(config)` not `return cls(config)`?

owencjones (commented):
Some queries, looking good

darwin/future/meta/client.py
Nathanjp91: good catch, I extracted this out from the core client so that's why.

owencjones (approved):

PR #598, opened 2023-05-12
[IO-1046][IO-1035] Team method upates and lazy loading

Nathanjp91:
# Problem
Update darwin-py v2 to include Team Meta objects, lazy loading and querying.

# Solution
Refactor of existing code from Core to Meta objects to enable lazy loading and query chaining. 
Changes:
- Query object now returns Meta objects, this is so that chaining can occur
- Meta Objects use property fields to call query objects for subfields
- Query Object now requires second generic T -> MetaBase, R -> DefaultDarwin, where R = the Core data object that T manages, ie T = TeamMeta, R = Team
```python
class Query(Generic[T, R], ABC):
```
- MetaBase base object that manages underlying data for Meta Objects, including results if loaded, requires to be instantiated with generics as above, ie, `DatasetMeta(MetaBase[Dataset])` or generically `class T(MetaBase[R])`
- MetaBase manages most iteration, list and collecting logic, but does require overwriting underlying object with the method to retrieve the objects in question. Examples seen in `DatasetMeta.__next__()`. 
- Point of Note: MetaBase objects should never call the Query objects for that object, otherwise circular logic occurs. They can use Queries for subfields, but to get the underlying object they should call core methods. For eg
```python
class TeamMeta(MetaBase[Team]):
    client: Client

    def __init__(self, client: Client, teams: Optional[List[Team]]=None) -> None:
        # TODO: Initialise from chaining within MetaClient
        self.client = client
        if not teams:
            teams = [get_team(self.client)] # <- Using TeamQuery introduces circular logic, use core method instead
        super().__init__(teams)
    

    @property
    def members(self) -> TeamMemberQuery: # <- Fine to use a query object here 
        return TeamMemberQuery(self.client)
```
- Client now loads Meta team, allowing `client.team[0].datasets[0]...` functionality
# Changelog
Team + other meta objects now able to be chained and lazily loaded.


darwin/future/data_objects/team.py
```diff
@@ -51,5 +55,32 @@ class Team(DefaultDarwin):
...
+def get_team_members(client: Client) -> Tuple[List[TeamMember], List[Exception]]:
+    response = client.get(f"/memberships")
+    members = []
+    errors = []
+    for item in response:
+        try:
+            members.append(TeamMember.parse_obj(item))
+        except Exception as e:
+            errors.append(e)
+    return (members, errors)
```
owencjones, line 86: Think the brackets are unnecessary, although totally stylistic, and your choice on this one.

owencjones (approved):
Approved - there's a couple of changes, but nothing I'd want to block this moving on

darwin/future/core/types/query.py
```diff
@@ -106,47 +110,38 @@ def __len__(self) -> int:
...
         return self
 
-    def __next__(self) -> QueryFilter:
-        if self.filters is None:
-            self.filters = []
-        if self.n < len(self.filters):
-            result = self.filters[self.n]
+    def __next__(self) -> R:
+        if self.results is None:
+            self.results = list(self.collect())
```
Nathanjp91, line 119: Interaction here becomes
self.collect() -> MetaObject
Then MetaObject stores a list of R and implement iteration dunders

.vscode/settings.json
```diff
@@ -24,5 +24,6 @@
     "black"
   ],
   "python.testing.pytestEnabled": true,
-  "python.linting.enabled": true
+  "python.linting.enabled": true,
+  "python.analysis.typeCheckingMode": "basic"
```
Nathanjp91, line 28: Have absolutely no idea what introduced this

darwin/future/core/types/query.py
```diff
@@ -106,47 +110,38 @@ def __len__(self) -> int:
...
+        if self.n < len(self.results):
+            result = self.results[self.n]
             self.n += 1
             return result
         else:
             raise StopIteration
+        
+    def __getitem__(self, index: int) -> R:
```
owencjones, line 127: Are our keys definitely always integers?  `__setitem__` and `__getitem__` also take strings as native, but I can't immediately think whether we have any issue with that.

darwin/future/data_objects/team.py
```diff
@@ -51,5 +54,32 @@ class Team(DefaultDarwin):
...
+        if not team_slug:
+            team_slug = client.config.default_team
+        return get_team(client, team_slug)
+
 
 TeamList = List[Team]
+
+
+def get_team(client: Client, team_slug: Optional[str] = None) -> Team:
+    """Returns the team with the given slug"""
```
owencjones, line 69: Same comment

darwin/future/data_objects/team.py
```diff
@@ -51,5 +54,32 @@ class Team(DefaultDarwin):
...
+    if not team_slug:
+        team_slug = client.config.default_team
+    response = client.get(f"/teams/{team_slug}/")
+    return Team.parse_obj(response)
+
+
+def get_team_members(client: Client) -> Tuple[List[TeamMember], List[Exception]]:
```
owencjones, line 76: And then this one, sorry!

darwin/future/data_objects/team.py
```diff
@@ -51,5 +54,32 @@ class Team(DefaultDarwin):
...
+
 
 TeamList = List[Team]
+
+
+def get_team(client: Client, team_slug: Optional[str] = None) -> Team:
+    """Returns the team with the given slug"""
+    if not team_slug:
+        team_slug = client.config.default_team
+    response = client.get(f"/teams/{team_slug}/")
```
owencjones, line 72: We're starting to get a mix of functions that use the monad `exceptions, return` response, and those that raise.  Not something for now, but we should work out what our approach is for using one vs the other.

darwin/future/meta/client.py
```diff
@@ -27,3 +29,9 @@ def from_api_key(cls, api_key: str, datasets_dir: Optional[Path] = None) -> Meta
         if datasets_dir:
             config.datasets_dir = datasets_dir
         return cls(config)
+
+    @property
+    def team(self) -> TeamMeta:
```
owencjones, line 34: Nice, pretty straightforward in the end.

darwin/future/meta/objects/base.py
```diff
@@ -0,0 +1,40 @@
+from __future__ import annotations
+
+from typing import Generic, List, Optional, TypeVar
+
+from darwin.future.pydantic_base import DefaultDarwin
+
+R = TypeVar("R", bound=DefaultDarwin)
+
+class MetaBase(Generic[R]):
```
owencjones, line 9: I was 50/50 whether to put this, but "metabase" is the name of a popular BI/Dashboarding tool.  I wouldn't normally care, but we do actually use it elsewhere in the stack, so this _might_ be worthy of renaming.

darwin/future/meta/objects/dataset.py
```diff
@@ -60,9 +59,16 @@ def _delete_by_slug(client: MetaClient, slug: str) -> int:
...
-        assert_is(isinstance(client, MetaClient), "client must be a MetaClient")
+    def _delete_by_id(client: Client, dataset_id: int) -> int:
+        assert_is(isinstance(client, Client), "client must be a MetaClient")
         assert_is(isinstance(dataset_id, int), "dataset_id must be an integer")
 
         dataset_deleted = remove_dataset(client, dataset_id)
         return dataset_deleted
+
+
+    def __next__(self) -> Dataset:
```
owencjones, line 70: Cheers!

darwin/future/meta/queries/team_member.py
```diff
@@ -20,18 +20,18 @@ def where(self, param: Param) -> TeamMemberQuery:
...
 
-        return TeamMemberQuery(query.filters)
+        return TeamMemberQuery(self.client, query.filters)
 
-    def collect(self, client: Client) -> List[TeamMember]:
-        members, exceptions = get_team_members(client)
+    def collect(self) -> TeamMembersMeta:
+        members, exceptions = get_team_members(self.client)
         if exceptions:
             # TODO: print and or raise exceptions, tbd how we want to handle this
```
owencjones, line 28: We should have a think about this before we have many more of these.

owencjones (approved):
Comments to definitely consider, but no outright demanded changes

darwin/future/meta/objects/base.py
Nathanjp91: The Meta Base object is definitely required for the generics, but don't mind what it's called.

darwin/future/core/types/query.py
Nathanjp91: as in just implicitly converting a string to an int here, like query['6'] or would there be some other meaningful interaction with a string as an index that we'd want? Like query['<name>'] returning the dataset name for instance? I would worry that has too much overlap with like query.where('name=...') and we'd just be maintaining multiple sets of functionality that do the same job. IMO better to just have and maintain the one.

.vscode/settings.json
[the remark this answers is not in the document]
Nathanjp91: By the looks of it, turns on pylance highlighting and syntax checking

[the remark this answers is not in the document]
Nathanjp91: pylance being the built in checker that comes with vscode

darwin/future/data_objects/team.py
```diff
@@ -51,5 +55,32 @@ class Team(DefaultDarwin):
     # Data Validation
     _slug_validator = validator("slug", allow_reuse=True)(parse_name)
 
+    @staticmethod
+    def from_client(client: Client, team_slug: Optional[str] = None) -> Team:
+        """Returns the team with the given slug"""
```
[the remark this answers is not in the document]
Nathanjp91: Yep, I'll go back and write up docs for all the stuff I've added

darwin/future/data_objects/team.py
Nathanjp91: I think for me it comes down to when we get a single object back or a list of objects. If we're only returning a single 'block', ie something that's all related to each other and if one thing goes wrong it's all poisoned, then it makes sense to just return the object + raise if issues, but if we have a function that returns a list of objects and they're all seperate, then it makes sense to package up the good data and return it with the exceptions monad style.

PR #601, opened 2023-05-17
[Io-1103][Internal] --ignore-slots option added to dataset pull

Nathanjp91:
Adding in the option to forcibly ignore slots even if dataset includes slots. Defaults to taking the first slot item and building the regular flat structure

### Changelog
--ignore-slots option added to `dataset pull` command


owencjones (approved):
Yep, good stuff

PR #605, opened 2023-05-22
[IO-1067] Filtering by class ids

Nathanjp91:
Filtering on dataset export currently uses the wrong filter REST payload and instead uses a filtering system intended for UI only use. 

### Changelog
Filters changed to use the same system as UI 'current filters' option rather than 'optionally filter by class'


NPatel1980:
QA Passed, export created in Darwinpy is now the same as an export triggered in the UI with "Current Filters" applied.

PR #607, opened 2023-05-25
[IO-1154][V7-5112][internal] Fix annotation import with no source_info

rslota:
This fixes an issue where `source_info` is required if any slots are provided. This doesn't need to be the case.

### DIFF
--- a/darwin/importer/importer.py
+++ b/darwin/importer/importer.py
@@ -481,6 +481,8 @@ def import_annotations(
             for parsed_file in track(files_to_track):
 
                 image_id, default_slot_name = remote_files[parsed_file.full_path]
+                # We need to check if name is not-None as Darwin JSON 1.0 
+                # defaults to name=None
                 if parsed_file.slots and parsed_file.slots[0].name:
                     default_slot_name = parsed_file.slots[0].name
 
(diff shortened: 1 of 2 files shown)


Nathanjp91 (approved):
LGTM

darwin/importer/importer.py
```diff
@@ -481,7 +481,9 @@ def import_annotations(
             for parsed_file in track(files_to_track):
 
                 image_id, default_slot_name = remote_files[parsed_file.full_path]
-                if parsed_file.slots:
+                # We need to check if name is not-None as Darwin JSON 1.0 
+                # defaults to name=None
+                if parsed_file.slots and parsed_file.slots[0].name:
```
Nathanjp91, line 486: is `name` field always guaranteed to be in a slot (but could still return None/'' and fail the bool check)? If not then this will raise if it try's to access `name` and it's not there. 
`if parsed_file.slots and 'name' in parsed_file.slots[0]` is a more defensive approach.

Nathanjp91 (approved):
pending clarification on that comment, looks good.

darwin/importer/importer.py
rslota: > is name field always guaranteed to be in a slot (but could still return None/'' and fail the bool check)

Yes, the field has to be there (`dt.Slot` type). If it wasn't there that would be a bug on its own IMO, so I don't want to do more defensive check.

Issue #604, opened 2023-05-19
Bug: image naming


peds:
When pulling the dataset using the CLI command `darwin dataset pull`, if the image names contain a `.` _in addition to_ the `.` before the file extension, the downloaded image names are incorrect. This issue applies to versions > 0.8.21.

**Example:**
Expected image name: `20221127T112041.338Z.jpg`
Downloaded image name: `20221127T112041.jpg`

I investigated the issue, and found the bug. It was introduced in this pull request https://github.com/v7labs/darwin-py/pull/580/files.
The issue is  due to incorrect usage of `with_suffix` method of the `Path` object on [line 319 of in darwin/dataset/download_manager.py]( https://github.com/v7labs/darwin-py/pull/580/files#diff-c35262c01c8b2e8c33f6cd9cc0e2c4e82427fbeb4d5826d32784aa51f5ca32d9R319). The `with_suffix` method is meant to be used to _replace_ a suffix. But `stem` does not contain a suffix, so if there is a  `.` in `stem`, what comes after will be incorrectly treated as a suffix and replaced by the `with_suffix` method.

**The fix** is to change that line from
<pre><code>filename = str(Path(stem).with_suffix(suffix))</code></pre>
to
<pre><code>filename = stem + suffix</code></pre>

Run the following code in a Python terminal to illustrate the issue with [line 319 in darwin/dataset/download_manager.py]( https://github.com/v7labs/darwin-py/pull/580/files#diff-c35262c01c8b2e8c33f6cd9cc0e2c4e82427fbeb4d5826d32784aa51f5ca32d9R319)
<pre><code>from pathlib import Path
image_filename="20221127T112041.338Z.jpg"
annotation_path=Path("20221127T112041.338Z.json")

# lines 317-318 in darwin/dataset/download_manager.py
suffix = Path(image_filename).suffix
stem = annotation_path.stem

# line 319
filename = str(Path(stem).with_suffix(suffix))
print(filename)
# >>> 20221127T112041.jpg

# line 319 fix
filename = stem + suffix
print(filename)
# >>> 20221127T112041.338Z.jpg
</code></pre>

I would be happy to create a pull request for this if that's preferable.

owencjones:
Thanks for the contribution on this, we actually had our own ticket in progress for it, so it's been addressed in the place where you dealt with it, and also some other areas of code, so your PR was closed, but not because we don't appreciate the effort!

Many thanks

PR #608, opened 2023-05-31
Fix pulling images with . in stem

nrudakov:
As mentioned in #604 `filename = str(Path(stem).with_suffix(suffix))` should be replaced with `filename = stem + suffix`


owencjones:
@nrudakov Thanks for your contribution, as @Nathanjp91 mentioned this is now fixed by a coincidental branch, so I'm closing this branch, but we would continue to appreciate any future contributions. 😊

Thanks again

V7 Python Devs 🐍

PR #611, opened 2023-05-31
[IO-1081 ]fix for dot/period/. in filename

Nathanjp91:
dots/periods in filenames cause the last 'segment' between the periods to be lost, and thus causing issues with filenames being downloaded/named incorrectly. This stems from issues with the .with_suffix() function. Replaced with stem + suffix solution that generates the correct filename

### Changelog
filenames with multiple dots now saving filenames correctly


nrudakov:
@Nathanjp91 maybe it worth also to get rid of unnecessary str to Path to str conversion on the same line

PaulWhiteQA:
QA Pass - Happy with the file formats when pulling with Darwin cli

PR #612, opened 2023-06-02
[IO-1205] Semantic Mask annotations outside of frame

Nathanjp91:
## Problem
Polygons that extend beyond the frame of reference (image height and width) get points truncated and this causes issues with the exported mask

## Solution
Checks if any annotation exceeds the window frame, if so it creates a larger mask defined by the full extent of all polygons in the annotations, creating and then offsetting the polygons to capture the entire space, before subsampling the full mask with the image dimensions in the correct location

## Changelog
Resolved an issue where out of bounds polygons caused semantic mask exports to truncate points and generate a wrong mask

### DIFF
--- a/darwin/exporter/formats/mask.py
+++ b/darwin/exporter/formats/mask.py
@@ -255,22 +255,19 @@ def render_polygons(
 
     errors: List[Exception] = []
 
-    annotations = [a for a in annotations if not isinstance(a, dt.VideoAnnotation)]
-    beyond_window = annotations_exceed_window(annotations, height, width)
+    filtered_annotations: List[dt.Annotation] = [a for a in annotations if not isinstance(a, dt.VideoAnnotation)]
+    beyond_window = annotations_exceed_window(filtered_annotations, height, width)
     if beyond_window:
-        image_mask = mask
-        x_min, x_max, y_min, y_max = get_extents(annotations, height, width)
+        # If the annotations exceed the window, we need to offset the mask to fit them all in.
+        # Capture the offsets so we can shift the annotations back to their original positions later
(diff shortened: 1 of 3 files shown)


darwin/exporter/formats/mask.py
```diff
@@ -469,3 +486,69 @@ def export(annotation_files: Iterable[dt.AnnotationFile], output_dir: Path, mode
...
+    Args:
+        annotations (List[dt.Annotation]): List of annotations
+        height (int): height of image
+        width (int): width of image
+
+    Returns:
+        bool: True if any annotation exceeds window, false otherwise
+    """
+    for item in annotations:
+        bbox = item.data["bounding_box"]
```
rslota, line 503: Technically `bounding_box` isn't required on polygons. Our backend will always return it, but client-written files may not always include it and this code would crash for it. I don't think we need to complicate this by doing a fallback that computes this bounding box, but maybe skipping annotations that don't have it for this check?

rslota (changes requested):
In general looks good, but we need to be careful with the fact that bbox is optional for polygons and have to handle this case "somehow", either but not making this fix work for such polygons or doing a fallback that creates that bbox if needed.

darwin/exporter/formats/mask.py
rslota: I general, I would like files with no bbox to not crash because of this fix. I don't mind not fixing this for them "today" (we can ticket it with lower priority), but I don't want to break anything existing.

Nathanjp91: Just added checks for the field, in both functions that use it, it should default to not extending the window

rslota (approved):

PR #616, opened 2023-06-14
[IO-1104][internal] DatasetQuery

owencjones:
# Problem
Need to implement meta queries for Datasets

# Solution
Have implemented meta queries in same style to Teams for Datasets.

# Changelog
Introduced DatasetQuery and tests
Refactored `list_datasets` to use monad error styles.


darwin/future/core/backend.py
```diff
@@ -13,7 +13,7 @@ def get_team(client: Client, team_slug: Optional[str] = None) -> Team:
 
 
 def get_team_members(client: Client) -> Tuple[List[TeamMember], List[Exception]]:
-    response = client.get(f"/memberships")
+    response = client.get("/memberships")
```
Nathanjp91, line 16: press f to pay respects

darwin/future/data_objects/release.py
```diff
@@ -19,6 +19,9 @@ class Release(DefaultDarwin):
 
     name: str
 
+    def __str__(self) -> str:
```
Nathanjp91, line 22: Is there a purpose to overwriting this? Pydantic objects already print to console?

darwin/future/meta/queries/dataset.py
```diff
@@ -0,0 +1,59 @@
...
+    def where(self, param: Param) -> "DatasetQuery":
+        filter = QueryFilter.parse_obj(param)
+        query = self + filter
+
+        return DatasetQuery(query.filters)
+
+    def collect(self, client: Client) -> List[Dataset]:
+        datasets, exceptions = list_datasets(client)
+        if exceptions:
+            # TODO: print and or raise exceptions, tbd how we want to handle this
```
Nathanjp91, line 31: We should definitely have a discussion about exceptions soon, figure out where we want to handle them, collecting exceptions into lists, logging (somewhat seperate but related)

Nathanjp91 (approved):
I don't see anything that is blocking. Some stuff I think we need to discuss and work-out for future though.

darwin/future/meta/queries/dataset.py
owencjones: Yeah, so the issue we have is whether we are withing a CLI or SDK context, I suppose.  We could maybe add `console` member to `client` so that CLI usage would be detectable, or we could use `sys` to detect the call stack, but that may be less x-plat compatible.

PR #617, opened 2023-06-15
[IO-1203] Missing dataset fields

Nathanjp91:
# Problem
Dataset items created via parse method lack the workflow status, id and dataset slug fields

# Solution
add parsing these into the object, dataset_slug is not a a part of the json response and has to be passed into the parse function manually, added in as an optional field with default of "n/a" to preserve any existing functionality

# Changelog
Items created via underlying parse method should correctly identify workflows they are attached to and the dataset they are a part of.

### DIFF
--- a/darwin/item.py
+++ b/darwin/item.py
@@ -95,7 +95,7 @@ class DatasetItem(BaseModel):
                 "dataset_slug": dataset_slug,
                 "seq": None,
                 "current_workflow_id": raw.get("workflow_data", {}).get("workflow_id"),
-                "current_workflow": raw.get("workflow_status"),
+                "current_workflow": raw.get("workflow_data"),
                 "slots": raw["slots"],
             }
         else:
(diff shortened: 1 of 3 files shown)


darwin/item.py
```diff
@@ -92,12 +92,11 @@ def parse(cls, raw: Dict[str, Any]) -> "DatasetItem":
                 "archived": raw["archived"],
                 "filesize": sum(file.get("size_bytes", 0) for file in raw["slots"]),
                 "dataset_id": raw["dataset_id"],
-                "dataset_slug": "n/a",
+                "dataset_slug": dataset_slug,
                 "seq": None,
-                "current_workflow_id": None,
-                "current_workflow": None,
+                "current_workflow_id": raw.get("workflow_data", {}).get("workflow_id"),
+                "current_workflow": raw.get("workflow_status"),
```
rslota, line 98: `workflow_status` is not a workflow data really. I think this should simply be `raw.get("workflow_data")`.

rslota (changes requested):

PaulWhiteQA:
QA Pass - Went over this on a call with Nathan and can confirm the correct parameters are now returned

PR #618, opened 2023-06-15
[IO-1194] Matching colour palette generation to class

Nathanjp91:
# Problem
Colour palette generation in semantic mask exports is randomly mismatched due to two underlying issues
* duplicates in the annotation class list when an annotation file has multiples of the same class present
* indeterminate sorting via sets/dicts in python
which means annotations can get assigned out of order to the generation

# Solution
Remove duplicates found in the annotations list by name and then sort on the remaining list. Then use this list for remaining colour palette generation

# Changelog
Colours should now map to the correct classes in semantic mask exports


darwin/exporter/formats/mask.py
```diff
@@ -475,21 +475,16 @@ def export(annotation_files: Iterable[dt.AnnotationFile], output_dir: Path, mode
         writer = csv_writer(f)
         writer.writerow(["class_name", "class_color"])
 
-        # Create a palette if there was no palette created
-        try:
-            palette_rgb
-        except NameError:
-            palette_rgb = {"__background__": [0, 0, 0]}
```
owencjones, line 482: Ah yeah, I see the issue you were talking about now.

darwin/exporter/formats/mask.py
```diff
@@ -561,11 +556,9 @@ def offset_complex_polygon(polygons: List, offset_x: int, offset_y: int) -> List
...
 def offset_simple_polygon(polygon: List, offset_x: int, offset_y: int) -> List:
     new_polygon = []
     for point in polygon:
-        new_polygon.append({
-            'x': point['x'] + offset_x,
-            'y': point['y'] + offset_y
-        })
-    return new_polygon
\ No newline at end of file
+        new_polygon.append({"x": point["x"] + offset_x, "y": point["y"] + offset_y})
```
owencjones, line 563: This could be a listcomp I guess, and would be a bit faster as one, but is likely more readable this way, so only change it if you want to.

tests/darwin/exporter/formats/export_mask_test.py
```diff
@@ -648,5 +666,94 @@ def test_export(
...
+            colours, counts = np.unique(flat_image, axis=0, return_counts=True)  # type: ignore
+            assert len(colours) == len(counts)
+            assert len(colours) == len(sizes)
+
+            for index, colour in enumerate(colours):
+                # regardless of particular colours assigned, the pixel count should be the same for that (r,g,b) tuple
+                assert tuple(colour) in inverse_mapping
+                assert counts[index] == sizes[inverse_mapping[tuple(colour)]]
+
+
```
owencjones, line 757: Nice addition

owencjones (approved):
There's one change I would make - take out the commented code, but I'm not going to block this by requiring myself to re-review it, so you discretion on this.

darwin/exporter/formats/mask.py
```diff
@@ -84,6 +84,7 @@ def get_rgb_colours(categories: dt.MaskTypes.CategoryList) -> Tuple[dt.MaskTypes
     dt.MaskTypes.RgbPalette
         A dict of categories and their corresponding RGB palette value.
     """
+    # categories = set(categories)
```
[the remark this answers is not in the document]
Nathanjp91: So I actually tested something like this, but the problem here is that each time this function gets called, it would reset sorting, and then also mucks with colour mapping. Sets are unordered in python.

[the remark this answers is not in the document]
Nathanjp91: oh sorry I see it's to remove the comment, not to use a set here

rslota (approved):

PR #619, opened 2023-06-20
[IO-1019][external] Meta delete datasets

owencjones:
# Problem
Meta client needs ability to delete datasets

# Solution
Added `DatasetMeta` object for chaining, and added delete method to it that works with slug or id.

# Changelog
Introduced DatasetMeta Object
Introduced delete_dataset method
Added tests for these
Refactored core code that was incompatible.

### DIFF
--- a/darwin/future/core/datasets/remove_dataset.py
+++ b/darwin/future/core/datasets/remove_dataset.py
@@ -1,6 +1,7 @@
 from typing import Optional
 
-from darwin.future.core.client import Client, JSONType
+from darwin.future.core.client import Client
+from darwin.future.exceptions.core.datasets import DatasetNotFound
 
 
 def remove_dataset(api_client: Client, id: int, team_slug: Optional[str] = None) -> int:
@@ -27,6 +28,6 @@ def remove_dataset(api_client: Client, id: int, team_slug: Optional[str] = None)
     )
 
     if "id" not in response:
-        raise Exception(f"Dataset with id {id} not found")
+        raise DatasetNotFound(f"Dataset with id {id} not found")
 
-    return int(response["id"])  # type: ignore
+    return int(response["id"])
--- a/darwin/future/meta/objects/dataset.py
+++ b/darwin/future/meta/objects/dataset.py
@@ -3,6 +3,7 @@ from typing import List, Optional, Tuple, Union
 from darwin.future.core.datasets.get_dataset import get_dataset
 from darwin.future.core.datasets.remove_dataset import remove_dataset
 from darwin.future.data_objects.dataset import Dataset
+from darwin.future.helpers.assertion import assert_is
 from darwin.future.meta.client import MetaClient
 from darwin.future.meta.queries.dataset import DatasetQuery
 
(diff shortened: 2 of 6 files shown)


darwin/future/core/client.py
```diff
@@ -186,7 +186,7 @@ def _contain_qs_and_endpoint(self, endpoint: str, query_string: Optional[QuerySt
             return endpoint
 
         assert "?" not in endpoint
-        return endpoint + "?" + str(query_string)
+        return endpoint + str(query_string)
```
owencjones, line 189: Found this as I was going, it was adding ? in two locations.  Thank you social testing.

darwin/future/core/datasets/remove_dataset.py
```diff
@@ -1,29 +1,32 @@
...
+        team_slug = api_client.config.default_team
+
+    response = api_client.put(
+        f"/datasets/{id}/archive",
+        {"team_slug": team_slug},
     )
 
-    return response
+    if "id" not in response:
+        raise Exception(f"Dataset with id {id} not found")
```
Nathanjp91, line 30: We should move these to custom exceptions when we get a chance

darwin/future/meta/objects/dataset.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class DatasetMeta:
+    client: MetaClient
+
+    def __init__(self, client: MetaClient) -> None:
+        # TODO: Initialise from chaining within MetaClient
+        self.client = client
+
+    def datasets(self) -> DatasetQuery:
```
Nathanjp91, line 17: Does this make sense? Having a dataset object that returns a query for datasets.

darwin/future/meta/objects/dataset.py
```diff
@@ -0,0 +1,67 @@
...
+        if dataset and dataset.id:
+            dataset_deleted = remove_dataset(client, dataset.id)
+        else:
+            raise Exception(f"Dataset with slug {slug} not found")
+
+        return dataset_deleted
+
+    @staticmethod
+    def _delete_by_id(client: MetaClient, dataset_id: int) -> int:
+        assert isinstance(client, MetaClient), "client must be a MetaClient"
```
Nathanjp91, line 63: I recently learned that these asserts are deleted if someone runs python with debug optimizations. Thoughts on making them 
```{python}
if not isinstance(...):
    raise Exception
```
It's not as clean but is less able to be abused.

Nathanjp91 (changes requested):
Small comments.

darwin/future/meta/objects/dataset.py
owencjones: The idea I had in mind was that it would be for later, and would basically be refactored, when we have the chaining like this:

```python

client.datasets().where(etc...)
```

So it's a placeholder for now, the idea being that it might be useful in wiring it all up later.

I've got a vague idea in mind of using the `__getattribute__` dunder to lazy load these on `MetaClient`, but ATM that's just a blue sky thought.  Could allow invocation without invocation though.

darwin/future/meta/objects/dataset.py
owencjones: I'm ok with that change although I was aware that in debug mode assertion checking is suspended.

Might introduce a simple partial to make it just as clean anyway.

Nathanjp91 (approved):
LGTM

PR #620, opened 2023-06-20
[IO-1291][internal] Fix rester mask conversion to sem mask

rslota:
# Problem

After recent import changes that fix rasters imports, the internal structure of rasters changed. This broken SemMask conversion. 

# Solution

Fixed tests and adjusted implementation to rely more on class types, rather than data. 

# Changelog
Fixed Darwin Raster -> Semantic Mask conversion.


owencjones (approved):
Yep, LGTM.  Be good to get into release tomorrow if we can QA in time.

rslota:
@owencjones I can see you approved, but it's not approved 😕

owencjones:
@rslota That was strange!  I've approved it again!

PR #621, opened 2023-06-22
[IO-1018][internal] Create Dataset meta method

owencjones:
# Problem
Meta objects had no way of creating datasets, which blocks workflows.

# Solution
Added a simple create dataset function to the `DatasetMeta` object, with tests.

# Changelog
Implemented previously `NotRaisedException` raising function `create_dataset` on `DatasetMeta`
Added unit tests for creation of datasets.

### DIFF
--- a/darwin/future/meta/objects/dataset.py
+++ b/darwin/future/meta/objects/dataset.py
@@ -149,9 +149,9 @@ class DatasetMeta:
         ------
         AssertionError
         """
-        slug = slug.lower().strip()
-        assert_is(isinstance(slug, str), "slug must be a string")
-        assert_is(len(slug) > 0, "slug must not be empty")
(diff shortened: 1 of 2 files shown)


Nathanjp91 (approved):
LGTM. I think stuff like the Query's should be an `@property` so they can just be called like `dataset.items.where()` (ie without the function call). Although I'm not too sure on datasetmeta having a dataset field, maybe hidden field? Not part of this PR though.

owencjones:
> LGTM. I think stuff like the Query's should be an `@property` so they can just be called like `dataset.items.where()` (ie without the function call). Although I'm not too sure on datasetmeta having a dataset field, maybe hidden field? Not part of this PR though.

Fair comment, will add as a ticket

PR #623, opened 2023-07-03
[IO-1343] Upolgyon simplify_polygon update versioning

Nathanjp91:
# Problem
Upolygon simplify polygon updates require update to darwin-py versioning requirements

# Solution
Update pyroject.toml file for new version

# Changelog
fix to rare issues with simplify_polygon and find_contours functions in exports


owencjones (approved):
Need to ensure we change it back when upoly is merged, otherwise LGTM

PR #624, opened 2023-07-03
[IO-1173][external] Workflows retrieve functions - core

owencjones:
# Problem
To handle workflows, we need to retrieve them

# Solution
Added both a `get_workflows` and `get_workflow` (by ID) function to the `core` module.

# Changelog
Core module now has workflow retrieve functionality.


darwin/future/tests/core/workflows/test_get_workflows.py
```diff
@@ -0,0 +1,82 @@
...
+    # Call the function being tested
+    workflows = get_workflows(base_client, team_slug=team_slug)
+
+    # Assertions
+    assert isinstance(workflows, List)
+    assert len(workflows) == len(response_data)
+    assert all(isinstance(workflow, Workflow) for workflow in workflows)
+
+
+@responses.activate
```
Nathanjp91, line 55: I like this syntax. Cleaner than the context manager I've been using.

darwin/future/core/workflows/get_workflows.py
```diff
@@ -0,0 +1,13 @@
...
+
+from darwin.future.core.client import Client
+from darwin.future.data_objects.workflow import Workflow
+
+
+def get_workflows(client: Client, team_slug: Optional[str] = None) -> List[Workflow]:
+    team_slug = team_slug or client.config.default_team
+    response = client.get(f"/v2/teams/{team_slug}/workflows?worker=false")
+
+    return [parse_obj_as(Workflow, workflow) for workflow in response]
```
Nathanjp91, line 13: If one of these fails, it'll fail to return the entire list. Which kind of breaks a bit from what we've been doing recently with the Tuple(List[Object], List[Exception]).

Nathanjp91: We could maybe write a generic that solves this
```python
def expand_exception_parse(response, T) -> Tuple[T, Exception]:
    ...
```

Nathanjp91 (approved):
Nothing really blocking, although monad style exceptions could be standaradized.

PR #626, opened 2023-07-05
Test cleanup

Nathanjp91:
# Problem
Pytest-describe blocks VScode test integration

# Solution
Preliminary swap of all pytest-describe tests to pytest class based equivelants

# Changelog
Tests will now be picked up by VScode


tests/darwin/dataset/remote_dataset_test.py
```diff
@@ -306,27 +306,30 @@ def files_content() -> Dict[str, Any]:
     }
 
 
-def test_dataset_creation_from_id():
-    def it_should_set_id_correctly_from_id(darwin_client: Client):
-        dataset_id = "team_slug/dataset_name:test_release"
-        dataset = darwin_client.get_remote_dataset(dataset_id)
+# This test was never actually running
+# TODO: Fix this test
```
owencjones, line 310: Q: Is this ticketed in backlog?

owencjones (approved):
LGTM, need to ticket replacing tests that turned out not to be running

tests/darwin/dataset/remote_dataset_test.py
Nathanjp91: not yet, I'll add one

PR #627, opened 2023-07-06
[IO-1176][internal] WorkflowQuery

owencjones:
# Problem
Currently we can't query workflows

# Solution
Have implemented a Workflow query object in standard style

# Changelog
* Added WorkflowQuery and tests
* Refactored get_workflow, and list_workflows to be monadic
* Some other improvements of a minor nature.


darwin/future/core/client.py
```diff
@@ -12,7 +12,7 @@
 from darwin.future.core.types.common import QueryString
 from darwin.future.exceptions.client import NotFound, Unauthorized
 
-JSONType = Dict[str, Any]  # type: ignore
+JSONType = Union[Dict[str, Any], List[Dict[str, Any]]]  # type: ignore
```
owencjones, line 15: The JSON that is returned from the `workflows` endpoint is a JSON array of objects, so this needed updating.

darwin/future/data_objects/workflow.py
```diff
@@ -53,8 +53,15 @@ class WFEdge(DefaultDarwin):
 
     id: UUID
     name: str
-    source_stage_id: UUID
-    target_stage_id: UUID
+    source_stage_id: Optional[UUID]
```
owencjones, line 56: `WFEdge`s only have both `source_stage_id` and `target_stage_id` if they are in the middle of a workflow.  Those at the ends only have one, so added `Optionals` and a validator requiring one or the other.

darwin/future/helpers/exception_handler.py
```diff
@@ -0,0 +1,37 @@
+import logging
+import sys
+from typing import List, Optional, Union
+
+from rich.console import Console
+
+
```
owencjones, line 7: Added a _potential_ exception handler, after working out how to discover whether the lib is being used from CLI or SDK.

darwin/future/meta/queries/dataset.py
```diff
@@ -2,9 +2,8 @@
 
 from darwin.future.core.client import Client
 from darwin.future.core.datasets.list_datasets import list_datasets
-from darwin.future.core.types.query import Modifier, Param, Query, QueryFilter
+from darwin.future.core.types.query import Param, Query, QueryFilter
 from darwin.future.data_objects.dataset import Dataset
```
owencjones, line 6: Incidental fix along the way.

darwin/future/helpers/exception_handler.py
```diff
@@ -0,0 +1,37 @@
...
+def handle_exception(exception: Optional[Union[Exception, List[Exception]]]) -> None:
+    """
+    Handles an exception or list of exceptions by printing them to the terminal
+
+    Parameters
+    ----------
+    exception : Optional[Union[Exception, List[Exception]]]
+        The exception(s) to handle
+    """
+    IS_INTERACTIVE_SESSION = sys.stdout and sys.stdout.isatty()
```
Nathanjp91, line 17: This is neat, I didn't know you could infer this

darwin/future/helpers/exception_handler.py
```diff
@@ -0,0 +1,37 @@
...
+        if exception := getattr(exc_info, "[1]", None):
+            ...
+        else:
+            raise ValueError("No exception provided and no exception in sys.exc_info")
+
+    if IS_INTERACTIVE_SESSION:
+        console = Console()
+        handler = console.print
+    else:
+        logger = logging.getLogger(__name__)
```
Nathanjp91, line 30: We should definitely use more logging in our code, but we also have to define a good strategy of where, what to capture. I think our current strategy just captures the CLI command being run which is kind of useless as the customer can just tell us that anyway.

owencjones: Yeah, agreed, but it feels like a piecemeal job like this, rather than a big all in one job?

darwin/future/meta/queries/workflow.py
```diff
@@ -0,0 +1,88 @@
...
+
+        if not self.filters:
+            return workflows
+
+        for filter in self.filters:
+            workflows = self._execute_filters(workflows, filter)
+
+        return workflows
+
+    def _execute_filters(self, workflows: List[Workflow], filter: QueryFilter) -> List[Workflow]:
```
Nathanjp91, line 46: I know this is my devising but for these complex objects I'm less enamoured with this filtering approach... but I can't think of anything better that still allows for the customizability for all the different (and custom) types we will have. Open to any suggestions if you've got them.

darwin/future/meta/objects/workflow.py
```diff
@@ -0,0 +1,23 @@
+from typing import List, Optional
+
+from darwin.client import Client
+from darwin.future.data_objects.workflow import Workflow
+from darwin.future.meta.objects.base import MetaBase
+
+
+class WorkflowMeta(MetaBase[Workflow]):
```
Nathanjp91, line 8: I think we should swap this ordering around, people are going to be using meta objects more than they are Core, so think it makes more sense to have WorfklowCore and the Meta is just Workflow. Doesn't have to be done now though but I might do it at some point.

Nathanjp91 (approved):
Don't have any changes requested. Some comments we should discuss and think about though.

darwin/future/meta/objects/workflow.py
owencjones: Ah yeah, that makes sense.  Feels like a all of them at once job, otherwise it doesn't make much sense - they fall out of sync with each other.

darwin/future/meta/queries/workflow.py
owencjones: Best I can think of is maybe a member taking a list of filtering functions?

Nathanjp91: The order should be reversed, so we only iterate over the list once rather than once per filter. Doesn't really matter here but for big datasets for instance it might become an issue. With the inheritance and typing though it can easily be implemented later for objects that need it. Just have to overwrite (or not even use) generic_execute_filter with one that is O(n) instead. 

Maybe filters becomes a list of functions that get called so something like 

`filtered = [x for x in unfiltered if any([y(x) for y in filters])]`

is possible

PR #629, opened 2023-07-10
[IO-1341] YoloV2 file overwriting bug.

owencjones:
File overwriting bug in YOLO conversions.

# Problem
In a scenario where two files are in a directory structure with the same name:

```
|-image.json
|-image_2.json
|
|--folder_1
|    |-image.png
|--folder_2
|    |-image.png
```

The use of the `AnnotationFile` attribute `filename` is used to define the output filenames (`image.txt` and `image.txt`), and this causes a clash, in which one file output _overwrites_ the other.

# Solution
Instead of using `AnnotationFile.filename`, it's better to use the `.stem` attribute of the path to the json file, as these are in the same directory, and can't share a name.

# Changelog
Changed file to use `Path.stem` instead of `AnnotationFile.filename`

### DIFF
--- a/darwin/exporter/formats/yolo.py
+++ b/darwin/exporter/formats/yolo.py
@@ -31,7 +31,7 @@ def export(annotation_files: Iterable[dt.AnnotationFile], output_dir: Path) -> N
 
 def _export_file(annotation_file: dt.AnnotationFile, class_index: ClassIndex, output_dir: Path) -> None:
     txt = _build_txt(annotation_file, class_index)
-    output_file_path = (output_dir / annotation_file.filename).with_suffix(".txt")
+    output_file_path = (output_dir / annotation_file.path.stem).with_suffix(".txt")
     output_file_path.parent.mkdir(parents=True, exist_ok=True)
     with open(output_file_path, "w") as f:
         f.write(txt)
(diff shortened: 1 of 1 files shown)


JWV7 (approved):
Works well for me

Quick side note, we do the same thing when converting to pascalvoc

Should we change line 188 in pascalvoc.py as well?

darwin/exporter/formats/yolo.py
```diff
@@ -31,7 +31,7 @@ def export(annotation_files: Iterable[dt.AnnotationFile], output_dir: Path) -> N
 
 def _export_file(annotation_file: dt.AnnotationFile, class_index: ClassIndex, output_dir: Path) -> None:
     txt = _build_txt(annotation_file, class_index)
-    output_file_path = (output_dir / annotation_file.filename).with_suffix(".txt")
+    output_file_path = (output_dir / annotation_file.path.stem).with_suffix(".txt")
```
Nathanjp91, line 34: There was an issue (not here I don't think) re using stem. I believe it was for filenames that included '.' character, 
`one.two.jpeg` for instance the .stem function loses that middle section IIRC becoming `one.jpeg`.  This did crop up as an issue elsewhere (I think for Brent, but I don't recall the ticket). Might want to verify that double dotted filenames also work.

Nathanjp91 (changes requested):

darwin/exporter/formats/yolo.py
JWV7: Well remembered, that was [IO-1081](https://linear.app/v7labs/issue/IO-1081/bug-pulling-a-release-with-cli-on-darwin-py-0822-and-0823-truncates)

Could we just apply that fix here? Instead of using `stem.with_suffix`, just `(stem + suffix)`

brain-geek (approved):

Nathanjp91 (approved):

PR #631, opened 2023-07-11
[IO-1331][internal] Introduce E2E tests module

owencjones:
# Problem
No provision for E2E tests in darwin-py

# Solution
Added configured pytest module.

# Changelog
Added pytest module
Added pytest.ini fule
Added minor docs.

### DIFF
--- a/e2e_tests/e2e_test.md
+++ b/e2e_tests/e2e_test.md
@@ -2,3 +2,4 @@
 
 This module is setup in `pytest.ini` to _not run_.  If you want to run the E2E tests,
 you will need to run `pytest e2e_tests` rather than `pytest`.
+


pytest.ini
```diff
@@ -0,0 +1,2 @@
+[pytest]
+addopts = --ignore=e2e_tests
```
Nathanjp91, line 2: So the intention here is that it'll only be run via Github Actions (or if you want to manually test)?

owencjones: Exactly, yeah

owencjones: You can run `pytest e2e_tests` and it will run them, but still run `pytest` on your local machine

PR #632, opened 2023-07-13
[IO-1332][internal] E2E Test harness

owencjones:
# Problem
We lack a simple way to run darwin commands and take output for assertions

# Solution
Write a simple wrapping function that runs darwin commands, and returns exit code, and `STDOUT` and `STDERR` streams outputs

# Changelog
Added `run_cli_command` function, which wraps CLI commands
Added tests for `run_cli_command`


tests/e2e_test_internals/test_run_cli_command.py
```diff
@@ -0,0 +1,58 @@
...
+    mock_subprocess_run.assert_called_once()
+
+    assert return_code == 137
+    assert std_out == "stdout"
+    assert std_err == "stderr"
+
+
+@mock.patch("e2e_tests.helpers.run")
+def test_does_not_pass_working_directory_to_run_cli_command(mock_subprocess_run: mock.Mock) -> None:
+    mock_subprocess_run.reset_mock()
```
Nathanjp91, line 49: this shouldn't be necessary, patch resets automatically after exiting a decorator

Nathanjp91 (approved):

tests/e2e_test_internals/test_run_cli_command.py
owencjones: That's true, originally, I was wrapping several tests in a `with mock.patch` but removed it because it's very messy.  I'll correct in the next ticket because it's in the same area

PR #633, opened 2023-07-13
Readme updates

Nathanjp91:
# Problem
No listed development env instructions

# Solution
Write some instructions

# Changelog
Development environment recommendations added to readme


DEV.md
```diff
@@ -0,0 +1,55 @@
+# Development Environment
```
owencjones, line 1: Would make more sense in the `docs` directory?

owencjones (changes requested):
See request to move file, but useful contribution overall

DEV.md
owencjones: I think you've `cp`d this instead of `mv`ing it. 😉

owencjones (changes requested):
Remove extra file, otherwise good.  I think after the meeting the other day we need to minimalise this as much as we can, but this is a good reflection of how it is.

owencjones (approved):
I like this, lets merge it unless there's a reason not to.

PR #634, opened 2023-07-14
[IO-1174][IO-1282] Workflow + Stage Meta objects + Stage Query

Nathanjp91:
# Problem
Workflow Meta objects and sub objects don't exist

# Solution
Create the meta objects for these associated items

# Changelog
Meta Workflows added to dpy2 functionality


darwin/future/core/types/query.py
```diff
@@ -2,7 +2,7 @@
```
owencjones, line 2: Just a generally well factored file eh

darwin/future/meta/objects/dataset.py
```diff
@@ -162,9 +155,3 @@ def _validate_slug(slug: str) -> None:
 
         VALID_SLUG_CHARS = "abcdefghijklmnopqrstuvwxyz0123456789-_"
         assert_is(all(c in VALID_SLUG_CHARS for c in slug_copy), "slug must only contain valid characters")
-
```
owencjones, line 165: It's nice doing a review that is largely deleting code that has been made unnecessary by simplification. 👍🏻

darwin/future/meta/queries/workflow.py
```diff
@@ -43,38 +44,38 @@ def collect(self) -> List[Workflow]:
...
-    def _execute_filters(self, workflows: List[Workflow], filter: QueryFilter) -> List[Workflow]:
+    def _execute_filters(self, workflows: List[WorkflowMeta], filter: QueryFilter) -> List[WorkflowMeta]:
         if filter.name == "id":
             id_to_find = UUID(filter.param)
             return [w for w in workflows if w.id == id_to_find]
 
         if filter.name == "inserted_at_start":
             start_date = datetime.fromisoformat(filter.param)
-            return [w for w in workflows if self._date_compare(w.inserted_at, start_date)]
+            return [w for w in workflows if w._item is not None and self._date_compare(w._item.inserted_at, start_date)]
```
owencjones, line 54: Maybe the `is not None` could've been reasonably added to the `_date_compare` function, but equally I'm not too bothered

darwin/future/core/types/query.py
Nathanjp91: I know, so much refactoring. At least it doesn't require 2 generics now though which makes everything easier to create.

owencjones (approved):
Yeah, looks good.  Large PR, but largely methodical refactor to simplify moving forward.

darwin/future/meta/queries/workflow.py
Nathanjp91: This is actually mostly for mypy here. It complains because _item is Optional[T] and thus it'll squiggly line everything without some check in the list comprehension.

Nathanjp91: but we could change the typing on _date_compare to take an Optional and check it in there as well

PR #637, opened 2023-07-18
user-agent update

Nathanjp91:
# Problem
We aren't capturing metrics around python version

# Solution
Add in user-agent info

# Changelog
Python version now captured in metrics


owencjones (approved):
If anything I'd go one further and add in platform ;)

PR #639, opened 2023-07-19
[MED-235][external] removed slot from argument to fix error message issue

tomvars:
# Problem
Current version of this in prod is failing leading to bad Exports hanging instead of failing gracefully - Error message creation is broken here due to an extra argument. This is a hotfix but we should add unit tests for this functionality.

# Changelog
Fixed bug associated with bad export error message creation

### DIFF
--- a/darwin/exporter/formats/nifti.py
+++ b/darwin/exporter/formats/nifti.py
@@ -63,7 +63,6 @@ def export(annotation_files: Iterable[dt.AnnotationFile], output_dir: Path) -> N
         output_volumes = build_output_volumes(video_annotation)
         slot_map = {slot.name: slot for slot in video_annotation.slots}
         for annotation in video_annotation.annotations:
-            # print(slot)
             populate_output_volumes(annotation, output_dir, slot_map, output_volumes, image_id)
         write_output_volume_to_disk(output_volumes, image_id=image_id, output_dir=output_dir)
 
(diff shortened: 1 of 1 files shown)


darwin/exporter/formats/nifti.py
```diff
@@ -158,8 +160,8 @@ def check_for_error_and_return_imageid(video_annotation: dt.AnnotationFile, outp
             str(filename),
```
owencjones, line 160: `create_error_message_json` no longer accepts 4 arguments.

darwin/exporter/formats/nifti.py
```diff
@@ -331,13 +360,11 @@ def process_affine(affine):
         return affine
 
 
-def create_error_message_json(error_message: str, output_dir: str, image_id: str, slot_name: str):
-    output_path = Path(output_dir) / f"{image_id}_{slot_name}_error.json"
+def create_error_message_json(error_message: str, output_dir: str, image_id: str):
```
owencjones, line 363: There's still a reference to this that calls it with the 4 original arguments.

Only a minor change, but one that would throw an error, so a necessary one.

owencjones (changes requested):
You've removed the final argument from `create_error_message_json`, but on L176 it's still called.

Given the hotfix context, I don't mind if you either:
* Remove argument on L176
* Add a discarded argument in function call declaration

Will re-review with priority if this needs it.

darwin/exporter/formats/nifti.py
owencjones: Oh right you are there, didn't spot that lack of a comma.  Ok, NP here then 😊

owencjones (approved):
All looks fine, the issue I originally raised was just unusual formatting, assumedly to reduce line length, and as a frequent two-pane window user, I appreciate that!

LGTM, ready for progression.

PR #640, opened 2023-07-21
[IO-1279] Functions to move items around stages

Nathanjp91:
# Problem
Don't have any way to move items around workflows in darwin-py at the moment

# Solution
Add in functions to stage items as well as backend core functions to move items

# Changelog
Can move items in workflows, see item's attached to a stage and manage workflows better

### DIFF
--- a/darwin/future/core/client.py
+++ b/darwin/future/core/client.py
@@ -155,7 +155,7 @@ class Client:
 
     @property
     def headers(self) -> Dict[str, str]:
-        http_headers: Dict[str, str] = {"Content-Type": "application/json"}
+        http_headers: Dict[str, str] = {"Content-Type": "application/json", "Accept": "application/json"}
         if self.config.api_key:
             http_headers["Authorization"] = f"ApiKey {self.config.api_key}"
         return http_headers
@@ -172,7 +172,7 @@ class Client:
         endpoint = self._sanitize_endpoint(endpoint)
         url = self.config.api_endpoint + endpoint
         if payload is not None:
-            response = method(url, payload)
+            response = method(url, json=payload)
         else:
             response = method(url)
 
--- a/darwin/future/core/items/get.py
+++ b/darwin/future/core/items/get.py
@@ -1,0 +1,32 @@
+
+
+from typing import List
+from uuid import UUID
+
+from darwin.future.core.client import Client
+from darwin.future.core.types.common import QueryString
+
+
+def get_item_ids(api_client: Client, team_slug: str, dataset_id: str) -> List[UUID]:
+    """
+    Returns a list of item ids for the dataset
+
+    Parameters
+    ----------
+    client: Client
+        The client to use for the request
+    team_slug: str
+        The slug of the team to get item ids for
+    dataset_id: str
+        The id or slug of the dataset to get item ids for
+
+    Returns
+    -------
+    List[UUID]
+        A list of item ids
+    """
+    
+    response = api_client.get(f"/v2/teams/{team_slug}/items/ids", QueryString({"not_statuses": "archived,error", "sort[id]": "desc", "dataset_ids": dataset_id}))
+    assert type(response) == dict
+    uuids = [UUID(uuid) for uuid in response["item_ids"]]
+    return uuids
--- a/darwin/future/core/items/get.py
+++ b/darwin/future/core/items/get.py
@@ -1,13 +1,11 @@
-
-
-from typing import List
+from typing import List, Union
 from uuid import UUID
 
 from darwin.future.core.client import Client
 from darwin.future.core.types.common import QueryString
 
 
-def get_item_ids(api_client: Client, team_slug: str, dataset_id: str) -> List[UUID]:
+def get_item_ids(api_client: Client, team_slug: str, dataset_id: Union[str, int]) -> List[UUID]:
     """
     Returns a list of item ids for the dataset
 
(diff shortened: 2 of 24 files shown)


darwin/future/core/client.py
```diff
@@ -155,7 +155,7 @@ def _setup_session(self, retries: Retry) -> None:
 
     @property
     def headers(self) -> Dict[str, str]:
-        http_headers: Dict[str, str] = {"Content-Type": "application/json"}
+        http_headers: Dict[str, str] = {"Content-Type": "application/json", "Accept": "application/json"}
```
Nathanjp91, line 158: This was required for the stage api, but if we need to set this for other api's we'll need to think about a syntax for setting headers of the client, either passing headers as an optional parameter to the client.get(...)/REST functions or changing headers before certain API calls

darwin/future/core/client.py
```diff
@@ -172,7 +172,7 @@ def _generic_call(self, method: Callable, endpoint: str, payload: Optional[dict]
         endpoint = self._sanitize_endpoint(endpoint)
         url = self.config.api_endpoint + endpoint
         if payload is not None:
-            response = method(url, payload)
+            response = method(url, json=payload)
```
Nathanjp91, line 175: data didn't work here for some reason

darwin/future/core/types/query.py
```diff
@@ -193,5 +193,21 @@ def where(self, *args: object, **kwargs: str) -> Query[T]:
     def collect(self) -> List[T]:
         raise NotImplementedError("Not implemented")
 
+    def collect_one(self) -> T:
```
Nathanjp91, line 196: Both here added as dogfooding session with Simon

darwin/future/data_objects/workflow.py
```diff
@@ -173,7 +179,7 @@ class Workflow(DefaultDarwin):
     inserted_at: datetime
     updated_at: datetime
 
-    dataset: WFDataset
+    dataset: Optional[WFDataset]
```
Nathanjp91, line 182: Some workflows on platform don't have a dataset attached, prevents collecting these items if it's required field. Although intuitively you would think it should be

darwin/future/meta/objects/base.py
```diff
@@ -1,20 +1,22 @@
...
 R = TypeVar("R", bound=DefaultDarwin)
+Param = Dict[str, object]
 
 
 class MetaBase(Generic[R]):
     _item: Optional[R]
     client: Client
 
-    def __init__(self, client: Client, item: Optional[R] = None) -> None:
+    def __init__(self, client: Client, item: Optional[R] = None, meta_params: Optional[Param] = None) -> None:
```
Nathanjp91, line 16: meta_params have to be added to the MetaBase here so that chaining can happen properly with backend API calls.
`client.team.workflows.first().stages.first().push_items_to_next_stage()`

pushing items to stage requires other parameters like workflow_id, dataset_id, team_slug. This need to be passed from Meta -> Query -> Meta as it goes down the chain

darwin/future/core/client.py
owencjones: I think it only works if the data type isn't "app/json" or on certain methods.

darwin/future/core/items/get.py
```diff
@@ -0,0 +1,64 @@
...
+    Returns
+    -------
+    List[UUID]
+        A list of item ids
+    """
+    response = api_client.get(
+        f"/v2/teams/{team_slug}/items/ids",
+        QueryString({"workflow_stage_ids": str(stage_id), "dataset_ids": str(dataset_id)}),
+    )
+    assert type(response) == dict
```
owencjones, line 62: Could consider using the other function we have for this which is impervious to `__DEBUG__` flag, but not the most important now - esp if this is just for making mypy happy!

darwin/future/core/items/get.py
```diff
@@ -0,0 +1,64 @@
...
+    -------
+    List[UUID]
+        A list of item ids
+    """
+
+    response = api_client.get(
+        f"/v2/teams/{team_slug}/items/ids",
+        QueryString({"not_statuses": "archived,error", "sort[id]": "desc", "dataset_ids": str(dataset_id)}),
+    )
+    assert type(response) == dict
```
owencjones, line 31: Same as my other (confusingly later) comment - could be `assertis` or just an `if`, but dw if this is just mypy cheering.

darwin/future/core/items/move_items.py
```diff
@@ -0,0 +1,41 @@
...
+    dataset_id: str
+        The id or slug of the dataset to move items for
+    stage_id: str
+        The id or slug of the stage to move items to
+    item_ids: List[UUID]
+        A list of item ids to move to the stage
+
+    Returns
+    -------
+    None
```
owencjones, line 28: Doesn't return `None`

darwin/future/data_objects/workflow.py
owencjones: Yep, intuitively I _did_ think so!  Good catch

darwin/future/meta/client.py
```diff
@@ -36,3 +38,7 @@ def team(self) -> TeamMeta:
         if self._team is None:
             self._team = TeamMeta(self)
         return self._team
+
+    @property
```
owencjones, line 42: Nice.  I'm really liking these chainables.

darwin/future/meta/objects/base.py
```diff
@@ -1,20 +1,22 @@
...
         self.client = client
         self._item = item
+        self.meta_params = meta_params or {}
```
owencjones, line 19: There's some argument for this being `or dict()` just because `{}` is also the syntax for set delimitation, but now I'm really scraping the barrel I guess.

darwin/future/meta/objects/dataset.py
```diff
@@ -1,10 +1,17 @@
-from typing import List, Optional, Tuple, Union
+from __future__ import annotations
```
owencjones, line 1: With this, I tend to just quote the returnable I need it for, but this is valid too, especially with our current support matrix.

darwin/future/meta/objects/dataset.py
```diff
@@ -155,3 +178,20 @@ def _validate_slug(slug: str) -> None:
 
         VALID_SLUG_CHARS = "abcdefghijklmnopqrstuvwxyz0123456789-_"
         assert_is(all(c in VALID_SLUG_CHARS for c in slug_copy), "slug must only contain valid characters")
+
+    def upload_files(
```
owencjones, line 182: Good move using the CLI function - I was looking at which to do when I went sick last week.

darwin/future/meta/objects/workflow.py
```diff
@@ -23,4 +43,37 @@ def id(self) -> UUID:
     def name(self) -> str:
         if self._item is None:
             raise ValueError("WorkflowMeta has no item")
-        return self._item.name
\ No newline at end of file
+        return self._item.name
+
+    def push_from_dataset_stage(self) -> WorkflowMeta:
+        assert self._item is not None
```
owencjones, line 49: Are all of these to satiate mypy?

darwin/future/meta/queries/stage.py
```diff
@@ -14,9 +14,10 @@ def collect(self) -> List[StageMeta]:
         if not self.meta_params:
             raise ValueError("Must specify workflow_id to query stages")
         workflow_id: UUID = self.meta_params["workflow_id"]
+        meta_params = self.meta_params.copy()
```
owencjones, line 17: Pass-by-ref was mutilating params as they were passed around?

owencjones (approved):
Yeah, looks good - I've commented a lot, but there's nothing that _needs_ changing.

darwin/future/core/items/get.py
Nathanjp91: it's mostly for mypy type checking, but we can use the assert_is in these cases too.

darwin/future/meta/objects/workflow.py
Nathanjp91: pretty much

PR #641, opened 2023-07-31
Metaname refactor

Nathanjp91:
# Problem
Naming of Meta/Core objects is a bit backwards in how we would like to encourage use

# Solution
Refactor naming conventions
Core Objects now have 'model' appended to signify they're a data model: Workflow -> "WorkflowModel"
Meta Objects drop the 'Meta', making them easier to import and use: WorkflowMeta -> Worfklow
# Changelog
Meta objects now renamed to be easier to use


pyproject.toml
```diff
@@ -61,7 +61,6 @@ python = ">=3.7.0,<3.12"
 argcomplete = "^2.0.0"
 humanize = "^4.4.0"
 pillow = "^9.4.0"
-pyyaml = ">=5.1"
```
Nathanjp91, line 64: This was breaking for me for some reason unless pyyaml was removed and then added back. Unsure why.

owencjones (commented):
It all looks good to me, only not approving because of how GH behaves with approved drafts.

owencjones:
Oh, but change the title to reflect a ticket if there is one.

PR #643, opened 2023-08-04
[IO-1405][external] YoloV8 Segementation dataset support

owencjones:
# Problem
We don't support YoloV8 Segmentation

# Solution
Have implemented YoloV8 Segmentation datasets.

# Changelog
Added new Yolo segementation exporter
Added Yolo shared helpers
Added social test for the exporter

### DIFF
--- a/darwin/exporter/formats/yolo_segmented.py
+++ b/darwin/exporter/formats/yolo_segmented.py
@@ -1,4 +1,5 @@
 from collections import namedtuple
+from enum import Enum, auto
 from logging import getLogger
 from pathlib import Path
 from typing import Iterable, List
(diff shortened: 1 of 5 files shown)


darwin/exporter/formats/yolo_segmented.py
```diff
@@ -0,0 +1,236 @@
...
+    """
+    The YoloV8 annotation types
+    """
+
+    UNKNOWN = auto()
+    BOUNDING_BOX = auto()
+    POLYGON = auto()
+
+
+def _build_text(annotation_file: AnnotationFile, class_index: ClassIndex) -> str:
```
owencjones, line 78: Extensive logging which largely only shows at `DEBUG` level, so won't usually show to users, but they could opt to see.  Allows for good diagnostics if this turns out not to work as intended.

darwin/exporter/formats/yolo_segmented.py
```diff
@@ -0,0 +1,236 @@
...
+        if "x" in annotation.data and "y" in annotation.data and "w" in annotation.data and "h" in annotation.data:
+            annotation_type = YoloSegmentedAnnotationType.BOUNDING_BOX
+        elif "points" in annotation.data:
+            if isinstance(annotation.data["points"][0], list):
+                logger.warn(f"Skipped annotation at index {annotation_index} because it's a complex polygon'")
+                continue
+
+            annotation_type = YoloSegmentedAnnotationType.POLYGON
+        else:
+            annotation_type = YoloSegmentedAnnotationType.UNKNOWN
```
hfcorreia, line 122: might be useful to extract to a helper func, improves the readability then we can do something like:

```python
annotation_type = get_annotation_type(annotation.data)
```

darwin/exporter/formats/yolo_segmented.py
```diff
@@ -0,0 +1,236 @@
...
+
+            except KeyError as exc:
+                logger.warn(
+                    f"Skipped annotation at index {annotation_index} because an"
+                    "expected key was not found in the data."
+                    f"Error occured while calculating point at index {last_point}."
+                    if last_point
+                    else "Error occured while enumerating points.",
+                    exc_info=exc,
+                )
```
hfcorreia, line 215: also would extract to helper to improve readability

darwin/exporter/formats/yolo_segmented.py
```diff
@@ -0,0 +1,236 @@
...
+
+                if CLOSE_VERTICES:
+                    points.append(Point(x=n_x5, y=n_y5))
+
+            except KeyError as exc:
+                logger.warn(
+                    f"Skipped annotation at index {annotation_index} because an"
+                    "expected key was not found in the data.",
+                    exc_info=exc,
+                )
```
hfcorreia, line 191: also would extract to helper to improve readability

darwin/exporter/formats/yolo_segmented.py
owencjones: Yeah, agreed.  In fact I only didn't because of the time frame.  The original plan had a helper func for this, and separate predicate functions for handling bounding boxes and polygons

darwin/exporter/formats/yolo_segmented.py
owencjones: Yeah, originally I did, but time was against me on this one

hfcorreia (approved):

PR #648, opened 2023-08-22
[IO-1335][internal] E2E setup and teardown

owencjones:
# Problem
Need to be able to run E2E tests with appropriate setup and teardown in place

# Solution
Added setup, teardown, config, and data seeding functions.

# Changelog
Added data seeding
Added Setup functions
Added teardown functions
Added pytest configuration functions
Added before and after run hooks


e2e_tests/conftest.py
```diff
@@ -22,21 +27,49 @@ def pytest_sessionstart(session: pytest.Session) -> None:
...
+        raise E2EEnvironmentVariableNotSet("E2E_TEAM")
+
     if not isinstance(session.config.cache, pytest.Cache):
         raise TypeError("Pytest caching is not enabled, but E2E tests require it")
 
     session.config.cache.set("server", server)
     session.config.cache.set("api_key", api_key)
+    session.config.cache.set("team_slug", team_slug)
+
+    global datasets
```
owencjones, line 48: Normally a very bad practice, this is one of the few exceptions, as this function can only run once in scope.  It's also the only way to solve this issue without storing the item in a custom format, and this is not pickleable.

e2e_tests/conftest.py
```diff
@@ -22,21 +27,49 @@ def pytest_sessionstart(session: pytest.Session) -> None:
...
+
+    print("Sleeping for 10 seconds to allow the server to catch up")
+    sleep(10)
+
+
+def pytest_sessionfinish(session: pytest.Session, exitstatus: int) -> None:
+    if not isinstance(session.config.cache, pytest.Cache):
+        raise TypeError("Pytest caching is not enabled, but E2E tests require it")
+
+    global datasets
```
owencjones, line 59: Normally a very bad practice, this is one of the few exceptions, as this function can only run once in scope.  It's also the only way to solve this issue without storing the item in a custom format, and this is not pickleable.

e2e_tests/conftest.py
```diff
@@ -22,21 +27,49 @@ def pytest_sessionstart(session: pytest.Session) -> None:
...
 
     session.config.cache.set("server", server)
     session.config.cache.set("api_key", api_key)
+    session.config.cache.set("team_slug", team_slug)
+
+    global datasets
+    datasets = setup_tests(ConfigValues(server=server, api_key=api_key, team_slug=team_slug))
+
+    print("Sleeping for 10 seconds to allow the server to catch up")
+    sleep(10)
```
owencjones, line 52: Sleep is also usually bad, but for test setup to account for latency, it's that or polling.

Nathanjp91 (commented):
So is the intendid use here for a test something along the lines of 

```python
def test_upload():
    pytest_sessionstart()
    run_workflow_I_want_to_test()
    pytest_sessionfinish()
    
```

If so, perhaps should add it as a decorator that does startup and teardown

```python

@e2e.basic
def test:
    run_workflow_I_want_to_test()
```

owencjones:
> So is the intendid use here for a test something along the lines of
> 
> ```python
> def test_upload():
>     pytest_sessionstart()
>     run_workflow_I_want_to_test()
>     pytest_sessionfinish()
>     
> ```
> 
> If so, perhaps should add it as a decorator that does startup and teardown
> 
> ```python
> @e2e.basic
> def test():
>     run_workflow_I_want_to_test()
> ```

Nah, `pytest` automatically runs the `pytest_sessionstart` and `pytest_sessionfinish` in the `conftest.py` file at the start and end of the test run, so these will happen before and after running all tests (but not each test) respectively.

TBH, Pytest is maybe not perfect for this kind of integration testing but it can be persuaded into it effectively, and given we are already familiar with it elsewhere, it seems to make better sense.

Nathanjp91 (approved):

PR #649, opened 2023-08-23
[IO-1607][internal] Hotfix for Yolov8

owencjones:
# Problem
YoloV8 was misidentifying items as complex polygons and refusing to export them.

# Solution
Improved identification logic, and added test case and tolerated old scenario in case it's needed for anything.

# Changelog
Refactor to YoloV8 exporter
Test added

### DIFF
--- a/darwin/exporter/formats/yolo_segmented.py
+++ b/darwin/exporter/formats/yolo_segmented.py
@@ -5,7 +5,8 @@ from multiprocessing.pool import CLOSE
 from pathlib import Path
 from typing import Iterable, List
 
-from darwin.datatypes import Annotation, AnnotationFile, VideoAnnotation
+from darwin.datatypes import Annotation, AnnotationFile, JSONType, VideoAnnotation
+from darwin.exceptions import DarwinException
 from darwin.exporter.formats.helpers.yolo_class_builder import (
     ClassIndex,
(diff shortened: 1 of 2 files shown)


hfcorreia (approved):

michalwski (approved):
LGTM

darwin/exporter/formats/yolo_segmented.py
```diff
@@ -78,14 +79,25 @@ class YoloSegmentedAnnotationType(Enum):
...
+    Returns
+    -------
+    YoloSegmentedAnnotationType
+        The annotation type.
+    """
+    type = annotation.annotation_class.annotation_type
+
+    if type == "bounding_box":
+        return YoloSegmentedAnnotationType.BOUNDING_BOX
+    elif type == "polygon":
```
Nathanjp91, line 100: is complex_polygon/circle/other annotation types not applicable to YOLO?

owencjones: Largely speaking, Yolo has historically been bounding boxes, and now supports polygons too, but ellipses, points, and other types are not supported - with many (2D plane) types, we could still process the bounding box and include it, but this wasn't in scope of the feature, and it was decided to include only bbox and simple polys for now.

PR #651, opened 2023-08-24
python 3.11 support

Nathanjp91:
# Problem
darwin-py doesn't support python 3.11

# Solution
Change dependencies to allow 3.11 support

# Changelog
Python 3.11 support added. Python 3.7 support dropped


owencjones (dismissed):
LGTM, some offline discussion to be had in house

pyproject.toml
```diff
@@ -72,20 +72,20 @@ jsonschema = ">=4.0.0"
 deprecation = "^2.1.0"
```
owencjones, line 72: Due to [Github limitation](https://github.com/orgs/community/discussions/4452) I can't comment on the exact line I need to, but on `L69` can be entirely removed now, as it's only needed by Python 3.7

Arguably, we could keep it for people who forcibly install on 3.7, but I think I err on the side of not doing personally, especially given that you've changed other python version dependant installs.

owencjones (changes requested):
I think remove `pyproject.toml:L69`, or at least consider it and respond. 🙂

owencjones (approved):

Nathanjp91:
Working via this PR instead
https://github.com/v7labs/darwin-py/pull/692

PR #652, opened 2023-08-24
[IO-1561] Mypy cleanup

Nathanjp91:
# Problem
Lots of mypy typing errors exist within the old code

# Solution
Begin cleanup of mypy typing issues

# Changelog
Some existing mypy typing issues resolved


darwin/torch/transforms.py
```diff
@@ -8,8 +8,7 @@
 
 from darwin.torch.utils import convert_segmentation_to_mask, flatten_masks_by_category
 
-
-TargetKey = Union["boxes", "labels", "mask", "masks", "image_id", "area", "iscrowd"]
+TargetKey = Literal["boxes", "labels", "mask", "masks", "image_id", "area", "iscrowd"]
```
owencjones, line 11: Nice addition.  Always love a Literal

darwin/torch/utils.py
```diff
@@ -11,6 +11,8 @@
 from darwin.dataset.identifier import DatasetIdentifier
 from darwin.datatypes import Segment
 
+NumpyCastable = Union[np.ndarray, torch.Tensor, List[Union[int, float]]]
```
owencjones, line 14: Numpy has types included in newer versions that can be used instead of `np.ndarray` but don't do it if it breaks anything.

`numpy.typing` contains:
* `ArrayLike`
* `DTypeLike`
* `NBitBase`
* `NDArray`

pyproject.toml
```diff
@@ -46,6 +46,13 @@ warn_return_any = false
 warn_unreachable = true
 pretty = true
 
+[[tool.mypy.overrides]]
+module = ["darwin.*", "tests.*"]
+disallow_any_unimported = false
+disallow_any_explicit = false
+disallow_untyped_calls = false
```
owencjones, line 53: What was it that necessitated these two (☝🏻 and 👇🏻), because this is not ideal to disallow globally, but I'm guessing you encountered something significant to change these flags.

tests/darwin/torch/dataset_test.py
```diff
@@ -4,8 +4,10 @@
 from unittest.mock import patch
 
 import numpy as np
+import pytest
```
owencjones, line 7: Unclear how these worked without this tbh!  I guess they only worked in a pytest session where it's already a global.

owencjones (changes requested):
Some stuff to respond to before approval, but no outright issues to change.

We should address pytest tests without imports and and run blocks at the bottom globally soon too - not directly relevant to this

darwin/torch/transforms.py
Nathanjp91: Annoyingly I think it's caused other issues to crop up elsewhere. Which is half the problem changing some of these. It's like whack-a-mole

darwin/torch/utils.py
Nathanjp91: Might give these a try, I suspect they'll cause issues because we use List[Union[int, float]] which might not be covered

tests/darwin/torch/dataset_test.py
Nathanjp91: Yeah I was confused too.

pyproject.toml
Nathanjp91: So the thing here is that these cause numpy and any third party library use to error if the third party doesn't have typing. To me there's no use being informed about errors we can't change, it just clogs up the visual and makes it harder to distinguish from things we can resolve.

tests/darwin/torch/dataset_test.py
owencjones: Global scope I guess

PR #653, opened 2023-08-24
[IO-1464] add in unknown parsing for WFTypes

Nathanjp91:
# Problem
WFType enum doesn't accept unknown values and this breaks certain testing with older/experimental Workflow types

# Solution
Introduce default parsing that accepts unknown values and casting them to an UNKNOWN enum type

# Changelog
Workflow stage enum no longer breaks on experimental and testing types


owencjones (approved):
Yep, all good.

PR #654, opened 2023-08-25
[IO-1554][internal] Refactoring github actions

owencjones:
# Problem
Workflows have actions missing, updates that need doing, and could be clearer

# Solution
Replace/refactor the workflows

# Changelog
Added new bash scripts to run each action
Refactor flows to trigger bash scripts where applicable.


owencjones:
Closing to prevent runs costing minutes, now the PR action is tested.

PR #655, opened 2023-08-30
[IO-1626][external] Client fails when data structure is unexpected

owencjones:
# Problem
On occasion, the API returns a nullish response instead of 0 for a zero count item, image or video count, which causes a crash

# Solution
Have written a single function that calculates the item count for each dataset, based on appropriate elimination logic

# Changelog
New function to calculate item counts, applied to all creations of `RemoteDatasetV1` and `RemoteDatasetV2` items.
Tests for new function
Tests to ensure that client and new function behave correctly _together_
Some cleanup of old typing and linting

### DIFF
--- a/darwin/client.py
+++ b/darwin/client.py
@@ -17,7 +17,14 @@ from darwin.dataset import RemoteDataset
 from darwin.dataset.identifier import DatasetIdentifier
 from darwin.dataset.remote_dataset_v1 import RemoteDatasetV1
 from darwin.dataset.remote_dataset_v2 import RemoteDatasetV2
-from darwin.datatypes import DarwinVersionNumber, Feature, ItemId, Team
+from darwin.datatypes import (
+    DarwinVersionNumber,
+    Feature,
+    ItemId,
+    JSONFreeForm,
+    JSONType,
+    Team,
+)
 from darwin.exceptions import (
     InsufficientStorage,
     InvalidLogin,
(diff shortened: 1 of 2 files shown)


darwin/client.py
```diff
@@ -5,19 +5,20 @@
 import zlib
```
owencjones, line 5: There are many changes in here which are just formatting or typing
* `Any` -> `UnknownType` are equivalent, but one doesn't cause issues with linter.
* Reformatting of lines has occurred in several unrelated places.

Only the implementation of the `_get_item_count` and its' use and testing are relevant to the fix

tests/darwin/client_test.py
```diff
@@ -73,11 +74,53 @@ def test_returns_list_of_datasets(self, darwin_client: Client):
             assert_dataset(remote_datasets[0], expected_dataset_1)
             assert_dataset(remote_datasets[1], expected_dataset_2)
 
+    @responses.activate
```
owencjones, line 77: This is a socialised test, testing that the client is using the `_get_item_count` function correctly.

There is a functional unit test as well, testing the function in a pure functional way.

simedw (approved):
overall LGTM 

normally I don't try to test `_` methods directly but I think it's fine.
Also _get_item_count doesn't really belong as a class method on Client, it's more of a utility function thing or could even live inside the RemoteDataset class?

PR #656, opened 2023-08-31
[IO-1626][external] Unexpected data from API causes crash

owencjones:
# Problem
On occasion, the API returns a nullish response instead of 0 for a zero count item, image or video count, which causes a crash

# Solution
Have written a single function that calculates the item count for each dataset, based on appropriate elimination logic

# Changelog
New function to calculate item counts, applied to all creations of `RemoteDatasetV1` and `RemoteDatasetV2` items.
Tests for new function
Tests to ensure that client and new function behave correctly _together_
Some cleanup of old typing and linting


Nathanjp91 (approved):
LGTM: Test coverage and changed in the relevant places

PR #662, opened 2023-09-17
Update local datasets to darwin 2.0

almazan:
# Problem
`LocalDataset` and other functions that read and parse local files (generally used to train models) do not support Darwin 2.0 format datasets. As it is implemented now, these functions assume datasets are 1.0 so they'll crash when loading a 2.0 dataset due to the mismatch of fields.

# Solution
Standardize the use of the util function `parse_darwin_json` across all these functions, which already abstracts dataset formats and supports both 1.0 and 2.0.

# Changelog
- Support for Darwin 2.0 datasets in `LocalDataset`
- Support for Darwin 2.0 datasets in `get_coco_format_record`

### DIFF
--- a/darwin/dataset/utils.py
+++ b/darwin/dataset/utils.py
@@ -301,16 +301,30 @@ def get_coco_format_record(
         new_obj = {"bbox_mode": box_mode, "category_id": category, "iscrowd": 0}
 
         if annotation_type == "polygon":
-            px, py = [], []
-            for point in obj.data["path"]:
-                px.append(point["x"])
-                py.append(point["y"])
-            poly = [(x, y) for x, y in zip(px, py)]
-            if len(poly) < 3:  # Discard polygons with less than 3 points
-                continue
-            segmentation = [list(itertools.chain.from_iterable(poly))]
+            # Support for complex polygons
+            if "paths" in obj.data:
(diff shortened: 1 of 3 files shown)


darwin/utils/utils.py
```diff
@@ -289,7 +289,7 @@ def find_files(
 
     for f in files:
         path = Path(f)
-        if path.is_dir() == True:
+        if path.is_dir() is True:
```
Nathanjp91, line 292: minor but this could just be `if path.is_dir()`

darwin/dataset/utils.py
```diff
@@ -277,41 +278,55 @@ def get_coco_format_record(
...
+            all_px, all_py = [], []
+            segmentation = []
+
+            for path in paths:
+                px, py = [], []
+                for point in path:
+                    px.append(point["x"])
+                    py.append(point["y"])
+                poly = [(x, y) for x, y in zip(px, py)]
+                if len(poly) < 3:  # Discard polygons with less than 3 points
```
Nathanjp91, line 320: This could be at the start of the function and it would skip any processing, `if len(path) < 3` rather than poly.

Nathanjp91 (approved):
Looks Good. Nothing major.

PR #664, opened 2023-09-21
[IO-1557] basic Export E2E test

Nathanjp91:
# Problem
Lacking any e2e testing for basic exports

# Solution
Adding in import annotations and export tests. Basic annotations file which is used to import for a test, then exporting on that dataset

Further, when developing and failing tests, the testing team on staging can get filled with datasets, refactored to teardown and datasets with a set naming format, including disconnecting and deleting any workflows added.

E2E testing also adds tests on PR when changes detected in e2e folder, won't run normally

# Changelog
Basic Export and import tests added


e2e_tests/test_darwin.py
```diff
@@ -63,6 +95,55 @@ def test_darwin_push(local_dataset_with_images: E2EDataset) -> None:
...
+    url = f"{config_values.server}/api/teams/{config_values.team_slug}/annotation_classes?include_tags=true"
+    response = api_call('get', url, None, config_values.api_key)
+    if not response.ok:
+        raise Exception(f"Failed to get annotation classes: {response.text}")
+    classes = response.json()['annotation_classes']
+    class_ids = [c["id"] for c in classes]
+    class_str = " ".join([str(c) for c in class_ids])
+    # Test darwin export
+    sleep(2)
+    result = run_cli_command(
```
rslota, line 143: I think it would be good to check if the export was actually created (not necessarily completed). I think you can do that with `releases` CLI command.

Nathanjp91: A little easier said than done, darwin-py truncates command line output if it's too long. Oddly, testing in debug environment exhibits different behaviour than running with pytest command. I think the cli output library we have might be influenced by local terminal settings on truncation size.

Nathanjp91: as in, an assert "test_darwin_export" in output can both pass and fail depending on where it's run, export name is truncated below for instance when run via pytest command, but passes when I run this in the debugger
image

Nathanjp91: Checking that there's not no releases will have to do

rslota: I guess for now... 
But eventually we'll need to figure this one out. Maybe special test mode where output is unformatted?

rslota: How does this output work when you redirect it to file?

Nathanjp91: Unfortunately the same result, the output is truncated in the file too, but also differs based on if you run from local command line or in debugging mode.

e2e_tests/test_darwin.py
```diff
@@ -50,19 +83,78 @@ def test_darwin_create(local_dataset: E2EDataset) -> None:
...
     assert result[0] == 0
 
+def test_darwin_import(local_dataset_with_annotations: E2EDataset) -> None:
+    """
+    Test importing a dataset via the darwin cli, dataset created via fixture with annotations added to objects
+    """
+    assert local_dataset_with_annotations.id is not None
+    assert local_dataset_with_annotations.name is not None
+    assert local_dataset_with_annotations.directory is not None
+    result = run_cli_command(
```
owencjones, line 106: Totally up to you, but you can write these pretty neatly as:

```python
exit_level, std_out, std_err = run_cli_command(
```

and then your asserts become really clear:
```python
assert exit_level == 0
```

Or, in your case here,

```python
exit_level, _, __ = run_cli_command(
```

Your call on this though, I can see why you've gone for this approach too.

e2e_tests/test_darwin.py
owencjones: Rich has a formatter class that you can pass to it, that overrides the formatting - there's probably a plainer one that could be used for specific cases if needed, but we'd have to make changes to cli code to allow us to switch to it, which is changing the code to suit the test a bit.

owencjones: Theoretically, you can create a virtual VT of any display width, and pipe to it, but that's a lot of work for a small gain I think.

owencjones: I think we need a `--porcelain` type command that allows darwin to run simple formatted output.

e2e_tests/helpers.py
```diff
@@ -46,7 +46,9 @@ def run_cli_command(command: str, working_directory: Optional[str] = None) -> Tu
             capture_output=True,
             shell=True,
         )
+
     try:
         return result.returncode, result.stdout.decode("utf-8"), result.stderr.decode("utf-8")
-    except:
+    except UnicodeDecodeError:
```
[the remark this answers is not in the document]
Nathanjp91: I actually tried encoding, `subprocess.run` has an argument to specify if you want an encoded text format back (or bytes default) and windows terminal still fails on that output so I had to go back

owencjones (approved):
Re-review approved

PR #665, opened 2023-09-22
[AI-1190] Implement and test augmentations

ChristofferEdlund:
# Problem
We want to introduce the possibility to use Albumentation transforms with darwin-py torch datasets. 

# Solution
Introducing an `AlbumentationsTransform` class in torch.transforms that can be be used in the following manner:

```
from darwin.torch.dataset import (
    ClassificationDataset,
    InstanceSegmentationDataset,
    ObjectDetectionDataset,
    SemanticSegmentationDataset,
)

from darwin.torch.transforms import AlbumentationsTransform

dataset_path = Path(r"/path/to/local/darwin/dataset")
transform = AlbumentationsTransform.from_path('/tmp/transform.json')

inst_dataset = InstanceSegmentationDataset(dataset_path=dataset_path, transform=transform)
```
One can initilize the `AlbumentationsTransform` in three ways:
- AlbumentationsTransform(albumentation_transform) <- needs an albumentations transformation
- AlbumentationsTransform.frompath(path) <- a path pointing to a .json or .yaml file defining the transformation
- AlbumentationsTransform.fromdict(dict) <- a dictionary defining the transform

To read more about the dictionary and file formats supported, we refer to the [albumentations documentation.](https://albumentations.ai/docs/examples/serialization/#Serializing-an-augmentation-pipeline-to-a-JSON-or-YAML-file)

Further.
- Instance segmentation dataset output is changed for bounding boxes to be `coco` format (X, Y, W, H) to be consistent with darwin-json annotations and the ObjectDetecion torch dataset. [BREAKING CHANGE]
- Clamping is introduced to ObjectDetection bboxes that is outside of image for more robust data loading.

# Changelog
- Introduced albumentation transform support for darwin torch datasts
- [BREAKING CHANGE] darwin.torch.dataset.InstanceSegmentationDataset has bbox coordinates changes from `pascal_voc` to `coco` format (X, Y, H, W)
- darwin.torch.dataset.ObjectDetectionDataset clamps bbox coordinates out-of-bound.

### DIFF
--- a/darwin/torch/transforms.py
+++ b/darwin/torch/transforms.py
@@ -1,6 +1,8 @@
 import random
 from typing import Any, Dict, Optional, Tuple, Union
 
+import albumentations as A
+import numpy as np
 import torch
 import torchvision.transforms as transforms
 import torchvision.transforms.functional as F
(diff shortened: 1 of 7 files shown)


darwin/torch/transforms.py
```diff
@@ -209,12 +197,12 @@ def __call__(self, image: PILImage.Image, target: TargetType) -> Tuple[PILImage.
             if num_keypoints:
                 keypoints = keypoints.view(num_keypoints, -1, 3)
 
-        keep = (boxes[:, 3] > boxes[:, 1]) & (boxes[:, 2] > boxes[:, 0])
-        boxes = boxes[keep]
-        classes = classes[keep]
-        masks = masks[keep]
-        if keypoints is not None:
-            keypoints = keypoints[keep]
+        # keep = (boxes[:, 3] > boxes[:, 1]) & (boxes[:, 2] > boxes[:, 0])
```
dorfmanrobert, line 200: can remove commented out lines

owencjones:
Updated title so that it's easier for me on deployment 😉

almazan (approved):
All changes look good to me.

almazan (changes requested):
Some comments need to be addressed before approval

darwin/torch/transforms.py
```diff
@@ -209,13 +215,6 @@ def __call__(self, image: PILImage.Image, target: TargetType) -> Tuple[PILImage.
             if num_keypoints:
                 keypoints = keypoints.view(num_keypoints, -1, 3)
 
-        keep = (boxes[:, 3] > boxes[:, 1]) & (boxes[:, 2] > boxes[:, 0])
-        boxes = boxes[keep]
-        classes = classes[keep]
-        masks = masks[keep]
-        if keypoints is not None:
-            keypoints = keypoints[keep]
```
almazan, line 217: I believe this logic was added for cases with bboxes of either width or height of size 0px. Should we add it back? The easiest way to add it back is to compute `keep` using a `tmp_boxes` converted to `x,y,x,y` using `tmp_boxes[:, 2:] += boxes[:, :2]` and then filter `boxes` using `keep`.

almazan (changes requested):

darwin/torch/transforms.py
```diff
@@ -191,7 +198,6 @@ def __call__(self, image: PILImage.Image, target: TargetType) -> Tuple[PILImage.
         boxes = [obj["bbox"] for obj in annotations]
         # guard against no boxes via resizing
         boxes = torch.as_tensor(boxes, dtype=torch.float32).reshape(-1, 4)
-        boxes[:, 2:] += boxes[:, :2]
         boxes[:, 0::2].clamp_(min=0, max=w)
         boxes[:, 1::2].clamp_(min=0, max=h)
```
almazan, line 202: This is not clamping bboxes anymore, since now indexes 2 and 3 are height and width instead of x2 and y2. Boxes now could go outside of the image.

almazan (approved):

PR #666, opened 2023-09-25
[IO-1565] video capture check

Nathanjp91:
# Problem
CV2 can fail to import videocapture function and fail elsewhere that isn't captured

# Solution
Make sure cv2 and videocapture function are imported

# Changelog
cv2 failed import should now print the right error message if VideoCapture function isn't found


owencjones (approved):
Simple bug fix, LGTM

Issue #603, opened 2023-05-19
unexpected behavior of consecutive use of RemoteDataset.pull() with use_folders=True


filippocastelli:
###  Issue description:
Pulling two different releases using `RemoteDataset.pull()` using `with_folders=True` can result in images missing from the second release pull unless `force_replace=True` is specified.

When making consecutive calls to `RemoteDataset.pull(release)` with different releases, if any file in the second release shares  filename with one in the first (even with different paths), pull is going to ignore those files unless `force_replace=True` 

`release_A` references the following files:
- `/dir_A/000.png`
- `/dir_A/001.png`

`release_B` references:
- `/dir_B/000.png`
- `/dir_B/002.png`

After calling
```[python]
RemoteDataset.pull(release_A, with_folders=True)
RemoteDataset.pull(release_B, with_folders=True)
```

The resulting `images` directory will be

```
images:
    - dir_A:
        - 000.png
        - 001.png
    - dir_B:
        - 002.png
```

### Expected Behavior:

Since `/dir_B/000.png` is not supposed to overwrite `/dir_A/000.png` , it's expected that the second pull should download it as well.
`force_replace=True` should only be necessary to download files with conflicting local paths. 

### Workaround

Specify `force_replace=True` when pulling data

### Environment
- `python==3.10`
- `darwin-py==0.8.24`

owencjones:
Hi Fillippo,

I will raise an internal ticket for this.  If you are a customer also reach out to Customer Experience.

Because of the current state management within `RemoteDataset`, this will require a bit of a refactor, so I'm not sure when this will be fixed.  In the meantime, really appreciate the heads up on the workaround 👍🏻 

Owen, V7

owencjones:
This is now largely done, so I expect it to be released in v0.8.44, scheduled for 10th October.

Closing now, please reach out if you experience this issue after that date, though.

Owen, V7

PR #667, opened 2023-09-25
[IO-1586][external] Changes to RemoteDataset.pull() to account for identically named files in different export releases

JWV7:
# Problem
As described in [this](https://github.com/v7labs/darwin-py/issues/603) GitHub issue, pulling two different releases using `RemoteDataset.pull()` passing `with_folders=True` **can** result in missing files from the second release unless `force_replace=True` is passed as well. This is because `pull()` does not download files if the stem of the filename is 
already present anywhere in the `~/.darwin/datasets/{team_slug}/{dataset_slug}/images` folder.

A further breakdown of the cause is available in the Linear ticket [IO-1586
](https://linear.app/v7labs/issue/IO-1586/bug-external-issue-with-sequential-dataset-pull-operations)

# Solution
Instead of checking the stem of existing filenames in the images folder, check the full local filepath of each file against the "planned" local path for each file in the export release. This way, if the "planned" local path already exists, we know that file doesn't need to be downloaded again.

This led to the discovery of an issue with the `remove_extra` flag, the purpose of which is to remove files from the `images` folder that aren't contained within the export release currently being pulled. Similarly to above, it was using filename stems to make decisions about which files should be removed, which meant it couldn't account for situations where a release contained duplicate filenames.

To resolve this, the functionality was moved up a level to `remote_dataset.py` where it has access to build a list of full local filepaths from the complete annotation files. Files in the `images` folder are then compared against this list, and those without a path contained in this list are removed.

# Changelog
Changes to RemoteDataset.pull() to prevent situations where identically named files in different folders could be overwritten when pulling different export releases

### DIFF
--- a/darwin/dataset/download_manager.py
+++ b/darwin/dataset/download_manager.py
@@ -104,7 +104,7 @@ def download_all_images_from_annotations(
 
         if not force_replace:
             # Check the planned path for the image against the existing images
-            planned_image_path = images_path / Path(annotation.remote_path.lstrip('/')).resolve().absolute() / Path(annotation.filename)
+            planned_image_path = images_path / Path(annotation.remote_path.lstrip('/\\')).resolve().absolute() / Path(annotation.filename)
             if planned_image_path in existing_images:
                 continue
 
--- a/darwin/dataset/download_manager.py
+++ b/darwin/dataset/download_manager.py
@@ -94,7 +94,7 @@ def download_all_images_from_annotations(
 
     # Verify that there is not already image in the images folder
     unfiltered_files = images_path.rglob(f"*") if use_folders else images_path.glob(f"*")
-    existing_images = [image for image in unfiltered_files if is_image_extension_allowed(image.suffix)]
+    existing_images = {image for image in unfiltered_files if is_image_extension_allowed(image.suffix)}
 
     annotations_to_download_path = []
     for annotation_path in annotations_path.glob(f"*.{annotation_format}"):
@@ -116,6 +116,14 @@ def download_all_images_from_annotations(
             if len(slot.source_files) > 1:
                 force_slots = True
 
+    if remove_extra:
+        # Removes existing images for which there is not corresponding annotation
+        annotations_downloaded_stem = [a.stem for a in annotations_path.glob(f"*.{annotation_format}")]
+        for existing_image in existing_images:
+            if existing_image.stem not in annotations_downloaded_stem:
+                print(f"Removing {existing_image} as there is no corresponding annotation")
+                existing_image.unlink()
+    
     # Create the generator with the partial functions
     download_functions: List = []
     for annotation_path in annotations_to_download_path:
--- a/darwin/dataset/download_manager.py
+++ b/darwin/dataset/download_manager.py
@@ -94,7 +94,7 @@ def download_all_images_from_annotations(
 
     # Verify that there is not already image in the images folder
     unfiltered_files = images_path.rglob(f"*") if use_folders else images_path.glob(f"*")
-    existing_images = {image.stem: image for image in unfiltered_files if is_image_extension_allowed(image.suffix)}
+    existing_images = [image for image in unfiltered_files if is_image_extension_allowed(image.suffix)]
 
     annotations_to_download_path = []
     for annotation_path in annotations_path.glob(f"*.{annotation_format}"):
@@ -103,11 +103,11 @@ def download_all_images_from_annotations(
             continue
 
         if not force_replace:
-            # Check collisions on image filename and json filename on the system
-            if annotation.filename in existing_images:
+            # Check the planned path for the image against the existing images
+            planned_image_path = images_path / Path(annotation.remote_path).relative_to('/') / Path(annotation.filename)
+            if planned_image_path in existing_images:
                 continue
-            if sanitize_filename(annotation_path.stem) in existing_images:
-                continue
+
         annotations_to_download_path.append(annotation_path)
         if len(annotation.slots) > 1:
             force_slots = True
@@ -116,13 +116,6 @@ def download_all_images_from_annotations(
             if len(slot.source_files) > 1:
                 force_slots = True
 
-    if remove_extra:
-        # Removes existing images for which there is not corresponding annotation
-        annotations_downloaded_stem = [a.stem for a in annotations_path.glob(f"*.{annotation_format}")]
-        for existing_image in existing_images.values():
-            if existing_image.stem not in annotations_downloaded_stem:
-                print(f"Removing {existing_image} as there is no corresponding annotation")
-                existing_image.unlink()
     # Create the generator with the partial functions
     download_functions: List = []
     for annotation_path in annotations_to_download_path:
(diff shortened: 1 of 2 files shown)


darwin/dataset/remote_dataset.py
```diff
@@ -351,7 +351,24 @@ def pull(
...
+                for annotation_path in annotations_dir.glob(f"*.json"):
+                    annotation = parse_darwin_json(annotation_path, count=0)
+                    local_paths.append(Path(annotation.slots[0].source_files[0]['local_path']))
+
+                for existing_image in existing_images:
+                    if existing_image not in local_paths:
+                        print(f"Removing {existing_image} as there is no corresponding annotation in this release")
+                        existing_image.unlink()
+
+            downloaded_file_count = len([f for f in self.local_images_path.rglob("*") if f.is_file() and not f.name.startswith('.')])
```
JWV7, line 370: I also discovered an issue where this line was falsely including `.DS_Store` files in the count if they existed and reporting these to the terminal after pulling the release

darwin/dataset/remote_dataset.py
```diff
@@ -351,7 +351,24 @@ def pull(
             for error in errors:
                 self.console.print(f"\t - {error}")
 
-            downloaded_file_count = len([f for f in self.local_images_path.rglob("*") if f.is_file()])
+            # Remove images that don't have a corresponding annotation file in the release
```
Nathanjp91, line 354: Is this necessary? The underlying `download_all_images_from_annotations` function gets passed the remove_extra flag and is meant to already unlink/delete files that don't have an annotation. If that's not doing it's job and still needs to be cleaned up afterwards then perhaps this is a separate ticket?

darwin/dataset/download_manager.py
```diff
@@ -94,7 +94,7 @@ def download_all_images_from_annotations(
 
     # Verify that there is not already image in the images folder
     unfiltered_files = images_path.rglob(f"*") if use_folders else images_path.glob(f"*")
-    existing_images = {image.stem: image for image in unfiltered_files if is_image_extension_allowed(image.suffix)}
+    existing_images = [image for image in unfiltered_files if is_image_extension_allowed(image.suffix)]
```
Nathanjp91, line 97: is there a reason to drop the dictionary here? Dictionaries have constant time lookup, so the checks later like if annotation.filename in existing_images are very quick on large datasets, but over an array it has to iterate over until it finds it (or the whole array worst case when it's not)

Nathanjp91: If just need the name rather than key:value for a dictionary, a set is the same as dictionary lookup and has the same comprehension notation just without the key:value `{x for x in list}`

Nathanjp91 (changes requested):
Comment to clarify

darwin/dataset/remote_dataset.py
Nathanjp91: good spot

darwin/dataset/download_manager.py
JWV7: The reason I changed to list is that the way this dict is constructed results in overwriting of keys. If there is more than 1 identical `image.stem` object in the release, then that key is written twice. This isn't a problem for appending to a list since we're only recording paths w/out stems

Given this, it sounds like a set is the best option, thanks for the knowledge!

darwin/dataset/remote_dataset.py
JWV7: I actually want to clarify one thing here - Is the intended behaviour of `remove_extra=True` to remove files that don't have a corresponding annotation file, or to remove files that are not part of the release currently being pulled?

Currently, the behaviour is the latter, because we remove all files that don't have a stem matching any stem in the release-specific annotations folder

If the latter is the intended behaviour, it works unless there are duplicate filename stems in the release, because it relies on the `existing_images` dict that I changed to a list (see comment above for why a dict was problematic). If `existing_images` is changed to a list or a set of paths where there are duplicate stems in the release, I can't think of a way to get a guaranteed mapping between annotation file and corresponding image/video etc. other than by inspecting the `local_path` field inside `source_files` for each annotation file. However, this isn't populated in annotation files that are part of the current pull **until** the download function is actually finished (i.e. exhaust_generator() is run). As far as I can tell, this means that the removal of extra files has to take place after this  (I chose to put it inside `remote_dataset.py`, but it could easily be put into a function)

Would this be good to put into a new ticket? There was already an issue with `remove_extra` before my changes (problematic `existing_images` dict), but my changes broke it more

darwin/dataset/download_manager.py
```diff
@@ -103,11 +103,11 @@ def download_all_images_from_annotations(
             continue
 
         if not force_replace:
-            # Check collisions on image filename and json filename on the system
-            if annotation.filename in existing_images:
-                continue
-            if sanitize_filename(annotation_path.stem) in existing_images:
+            # Check the planned path for the image against the existing images
+            planned_image_path = images_path / Path(annotation.remote_path).relative_to('/') / Path(annotation.filename)
```
owencjones, line 107: `Path(something).relative_to('/')` is POSIX specific (windows may not like it) and should be equivalent to `Path(something).resolve().absolute()`

You need the `resolve` step to prevent it including incorrect symlinks or `..` segs that are unresolved.

darwin/dataset/remote_dataset.py
owencjones: I don't mind this here, but creating an extra ticket and progressing it with this merge would be good for tracking.

owencjones (commented):
Some comments for now, good first PR though :)

darwin/dataset/remote_dataset.py
JWV7: Have made changes to ensure that `remove_extra` functions exactly as it did before the ticket, will raise it separately

darwin/dataset/download_manager.py
JWV7: `Path(something).resolve().absolute()` doesn't strip the leading `/` in from `annotation.remote_path` which prevents concatenation, so instead I'll remove the leading `/` using string methods

darwin/dataset/remote_dataset.py
JWV7: @rslota Would you be able to clarify on my comment above? In working on this issue, I discovered a problem with the remove_extra flag, and neither Nathan nor I are able to tell what it's intended behaviour is

I'll create a new ticket to work on it

Nathanjp91 (approved):
LGTM with the change to Set + full filename rather than overwrite on partial filename in dict

darwin/dataset/download_manager.py
JWV7: Just made a change to account for this

darwin/dataset/remote_dataset.py
rslota: Sorry for late reply, I missed the notification. 
`remove_extra` afaik is here to remove any image files that is already downloaded that would not get overridden by the current release being pulled. In other words, it should work the same as wiping image directory before pulling the release, however, it can't be implemented this way, in case we have issues downloading "new" images, and customer would ended up with empty target directory.

PR #668, opened 2023-09-26
[ANN-2555][External] Fixed bug when it is not possible to import mask annotations for a video

AndriiKlymchuk:
# Problem
When importing mask annotations for a video item, darwin-py skips them with next warning:
`[WARNING] Unsupported annotation type: 'dict_keys(['keyframe', 'mask', 'name', 'id'])'`

# Solution
Add logic for reorganizing such annotations as valid once. 

# Changelog
Fixed bug when it is not possible to import mask annotations for a video

### DIFF
--- a/darwin/datatypes.py
+++ b/darwin/datatypes.py
@@ -959,6 +959,74 @@ def make_graph(
         AnnotationClass(class_name, "graph"), {"nodes": nodes, "edges": edges}, subs or [], slot_names=slot_names or []
     )
 
+def make_mask(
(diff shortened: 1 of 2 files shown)


AndriiKlymchuk:
> Looks good as a bug fix, untested. Could you raise an IO ticket, with label 'darwin-py' for adding a happy path / sad path test and we'll handle that?

Done - https://linear.app/v7labs/issue/IO-1812/cover-datatypes-module-by-tests

PR #669, opened 2023-09-27
[IO-1739] Slot names mistakenly required for mask/rasters

Nathanjp91:
# Problem
Slot name is a required field, contrary to other annotation types 

# Solution
Remove the requirement checks that fail

# Changelog
Rasters and Masks now no longer require slot names, in line with schema


owencjones (approved):
Yep, nice simple fix.  One possible addition is a test confirming a "happy" path of not requiring the value, but it's largely completist and probably unnecessary.

PR #674, opened 2023-10-02
[IO-1661][external] Error messages were not clear to customers

owencjones:
# Problem
COCO `iscrowd` error messages were not clear to customers.

# Solution
Have changed the messages to explain more verbosely what the issue is with a crowd import.

# Changelog
`coco.py` now uses `logger`, and the message is more verbose

### DIFF
--- a/darwin/importer/formats/coco.py
+++ b/darwin/importer/formats/coco.py
@@ -1,5 +1,6 @@
+from logging import getLogger
 from pathlib import Path
-from typing import Any, Dict, Iterator, List, Optional
+from typing import Dict, Iterator, List, Optional
 
 import deprecation
 import orjson as json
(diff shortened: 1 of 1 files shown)


darwin/importer/formats/coco.py
```diff
@@ -112,7 +118,10 @@ def parse_annotation(annotation: Dict[str, Any], category_lookup_table: Dict[str
     iscrowd = annotation.get("iscrowd") == 1
 
     if iscrowd:
-        print("Warning, unsupported RLE, skipping")
+        logger.warn(
+            f"Skipping annotation {annotation.get('id')} because it is a crowd"
+            "annotation, and this is not supported from Coco."
+        )
```
Nathanjp91, line 124: Unless I'm mistaken, coco format supports iscrowd? Unless we just mean our coco importer/exporter? In which case maybe want to differentiate between coco the format and our code.

Nathanjp91 (approved):
One thing to consider

darwin/importer/formats/coco.py
owencjones: Yeah, I can clarify this - Coco _does_ support `iscrowd`, _we_ don't import it - in fact that's why this ticket exists.

NPatel1980:
QA passed, paired with Owen and observed the new message.

PR #677, opened 2023-10-05
[IO-1560] e2e Dataset Delete test

Nathanjp91:
# Problem
No e2e test for dataset delete

# Solution
Add dataset delete method to testing suite

# Changelog
e2e dataset deletion now tested


owencjones (approved):
LGTM, I'm happy for us to merge now if you've tested this works

PR #678, opened 2023-10-05
[IO-1445][external] Changes to LocalDataset() & get_annotations() to account for local releases pulled with folders

JBWilkie:
… pulled with folders

# Problem
When local releases are pulled with folders, construction of LocalDataset objects & the get_annotations() function fail due to a mismatch between annotation file names and item names. A more in-depth breakdown is available in the [IO-1445](https://linear.app/v7labs/issue/IO-1445/bug-classification-model-failed-a-potential-customer-trial-contacted) Linear ticket

# Solution
Instead of comparing the file names of the JSON with items, parse each JSON file to guarantee the correct local path for each item is checked against. This has the side effect of introducing significantly higher runtimes when parsing large JSON files. Therefore, make LocalDataset & get_annotations() use a streaming library to prevent this.

# Changelog
- Changes to LocalDataset construction & get_annotations() function to account for local releases pulled with folders, including releases with identically named files.
- Addition of the json_stream library to darwin-py to improve runtimes when reading certain fields from JSON files


JWV7:
Failing tests, need to investigate

Nathanjp91 (approved):
LGTM. Revisions seem fine with the QA and testing done.

PR #679, opened 2023-10-09
[AI-1260][internal] add loading of polygon support for object detection datasets

ChristofferEdlund:
# Problem
In the initial implementation, when loading classes for datasets labeled with the `bounding_box` annotation type, it wasn't taking into consideration that the `polygon` annotations could also be relevant because they contain bounding box information in their annotations.

# Solution
We modified the load_classes function to incorporate the following logic:

1.   When the annotation_type is specified as bounding_box, the function now additionally considers polygon annotations.
2.   It attempts to load class labels from both classes_bounding_box.txt and classes_polygon.txt files, combining the two lists if both exist.
3.   The changes ensure a more comprehensive loading of classes that may be relevant to object detection tasks.

Aka, we now load BOTH bounding_box and polygon classes that has been exported per default.

# Changelog

-   Enhanced the load_classes function to load polygon class labels when the bounding_box annotation type is specified.
-   Ensured a combined list of classes from both bounding boxes and polygons for object detection datasets.


darwin/dataset/utils.py
```diff
@@ -160,10 +173,26 @@ def get_classes(
     dataset_path = Path(dataset_path)
```
owencjones, line 173: Is there a way we could take the opportunity to separate this function for torch dataloaders and remote datasets and only change this for the dataloaders?  I assume that's what you need?

The `get_classes` function touches a lot of code, so changes like this might have far reaching consequences we don't see here, and Darwin py is not yet with a proper set of regression tests, so I'm maybe being overcautious, but.

tests/darwin/torch/dataset_test.py
```diff
@@ -74,8 +74,12 @@ def test_should_correctly_create_a_object_detection_dataset(
         self, team_slug: str, team_extracted_dataset_path: Path
     ) -> None:
         root = team_extracted_dataset_path / team_slug / "coco"
+        print(root)
```
owencjones, line 77: Prefer `Logging` lines, but in the torch code, I leave it to the AI team a bit more!

owencjones (changes requested):
Some changes, and one query

tests/darwin/torch/dataset_test.py
ChristofferEdlund: Sorry, that is just code that should not be there.

darwin/dataset/utils.py
```diff
@@ -131,6 +131,19 @@ def make_class_lists(release_path: Path) -> None:
                 f.write("\n".join(classes_names))
 
 
+def get_classes_from_file(path):
```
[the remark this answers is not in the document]
ChristofferEdlund: Great, thanks! Both was already imported.

darwin/dataset/local_dataset.py
```diff
@@ -64,20 +63,6 @@ def __init__(
         split_type: str = "random",
         release_name: Optional[str] = None,
     ):
-        assert dataset_path is not None
```
ChristofferEdlund, line 67: Refactoring the __init__ function to make ruff happy

ChristofferEdlund (commented):
The major changes is that extract_classes and get_classes has the same default behavior. 

FThe logic of loading polygon annotations with bounding_box are placed a level higher in stratified sampling and local_dataset. This is done by enabling loading of a list of annotation_types.

tests/darwin/torch/dataset_test.py
owencjones: Ok, can we remove it?

darwin/dataset/utils.py
owencjones: Ok, after discussion, this seems ok.

darwin/dataset/local_dataset.py
```diff
@@ -107,16 +141,24 @@ def __init__(
...
+                    f"Image ({stem}) is present with multiple extensions. This is forbidden."
+                )
             self.images_path.append(images[0])
             self.annotations_path.append(annotation_path)
 
-        if len(self.images_path) == 0:
-            raise ValueError(f"Could not find any {SUPPORTED_IMAGE_EXTENSIONS} file", f" in {images_dir}")
-
-        assert len(self.images_path) == len(self.annotations_path)
+    def _initial_setup(self, dataset_path, release_name):
```
owencjones, line 154: I like the extraction here, even if it is only to make Ruff happy!

darwin/dataset/utils.py
```diff
@@ -238,102 +285,97 @@ def get_coco_format_record(
     image_id: Optional[Union[str, int]] = None,
     classes: Optional[List[str]] = None,
 ) -> Dict[str, Any]:
-    """
```
owencjones, line 241: What was the reason behind removing the docblock?  Internal function?

owencjones (approved):
Looks sound to me, I think QA needs to be solid, and Jon should ideally have a glance, because this changes code I'm not as familiar with, that he _probably wrote_ :)

darwin/dataset/utils.py
ChristofferEdlund: That's strange, its there in my code. Will see if I can get it back on the github as well.

darwin/dataset/utils.py
```diff
@@ -93,7 +100,10 @@ def extract_classes(annotations_path: Path, annotation_type: str) -> Tuple[Dict[
             continue
 
         for annotation in annotation_file.annotations:
-            if annotation.annotation_class.annotation_type != annotation_type:
+            if (
```
dorfmanrobert, line 103: I believe this may be inconsistent with [this logic](https://github.com/v7labs/darwin-py/blob/0f799a589c884cebd699fb62b2713758f4031464/darwin/dataset/local_dataset.py#L328). Here it seems "complex_polygon" will not be considered in creating the class list, while in the linked logic "complex_polygon" is considered for an annotation.annotation_class.annotation_type.

dorfmanrobert: Not sure which is better: remove complex_polygon from linked logic or add it here

ChristofferEdlund: Great input, I have not changed anything considering that logic. So the behavior should be the same as main darwin-py in that regard. BUT, that does not stop us from fixing it if there is an issue.

ChristofferEdlund: I suggest, we skip this for now. It might be a bug, but it would be a pre-existing condition and we have the same behavior as before. Besides, I think that complex-polygons are getting deprecated. 

We could raise a bug report thought.

almazan (commented):
Looks good.

Disclamer: really long PR to review, so I might have missed things. Haven't tested it myself.

darwin/dataset/split_manager.py
```diff
@@ -146,6 +146,10 @@ def split_dataset(
     train_size: int = dataset_size - val_size - test_size
     split_id = f"{train_size}_{val_size}_{test_size}"
 
+    assert train_size > 0, f"Found {train_size} train examples, we need at least 1"
+    assert val_size > 0, f"Found {val_size} validation examples, we need at least 1"
+    assert test_size > 0, f"Found {test_size} test examples, we need at least 1"
```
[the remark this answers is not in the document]
ChristofferEdlund: I think I changed that, so we only enforce it on train

almazan (approved):

PR #681, opened 2023-10-10
[IO-1692][internal] Made import more tolerant of potential errors

owencjones:
# Problem
In version 0.8.38, CX reported issues with intolerance of errors in some empty annotations, where these should simply be skipped.

# Solution
Added logic that checks for `None` values in the chained subscripted lookups, so that these will be logged and skipped in future.
Added use of logger for the output logging.

# Changelog
Added error checking block to importer


darwin/importer/importer.py
```diff
@@ -39,6 +39,8 @@
 from darwin.utils import flatten_list, secure_continue_request
```
owencjones, line 39: **NB: Most changes are Black reformatting the file, I've marked the section involved**

darwin/importer/importer.py
```diff
@@ -647,6 +645,16 @@ def _import_annotations(
     for annotation in annotations:
         annotation_class = annotation.annotation_class
         annotation_type = annotation_class.annotation_internal_type or annotation_class.annotation_type
+
```
owencjones, line 648: **Main body of changes here**

darwin/importer/importer.py
```diff
@@ -26,7 +25,7 @@
 
 if TYPE_CHECKING:
     from darwin.client import Client
-    from darwin.dataset import RemoteDataset
+    from darwin.dataset.remote_dataset import RemoteDataset
```
Nathanjp91, line 28: Confused by this change? Was there a reason?

Nathanjp91 (approved):
Small q but not important

darwin/importer/importer.py
owencjones: Mypy complaint - not direct import

PR #685, opened 2023-10-12
[IO-1454][IO-1455][IO-1456][IO-1458] Streamlining of Team/Dataset deletion and creation methods

Nathanjp91:
# Problem
Dogfooding revealed ways we'd like to use team and dataset objects regarding deletion and creation

# Solution
Changes include
- Team object now has a way to create a dataset via `team.create_dataset(slug)`
- Team object now has a delete_dataset method
- Dataset now has a delete method, allowing interaction with queries `Team.dataset.where(...)[0].delete()`
- Changes to handling of exceptions
- Ruff formatting changes
- Black formatter unification
# Changelog
Team and Dataset methods now include new ways to create/delete


darwin/future/meta/objects/stage.py
```diff
@@ -17,20 +17,29 @@ class Stage(MetaBase[WFStageCore]):
 
     @property
     def item_ids(self) -> List[UUID]:
-        """_summary_
+        """Item ids attached to the stage
```
owencjones, line 20: Actually quite surprised Sphinx didn't moan about this (pre-this change) but maybe we have it excluded in sphinx steps atm

darwin/future/tests/data_objects/workflow/test_wfedge.py
```diff
@@ -1,6 +1,5 @@
 from pathlib import Path
 
-import pytest
```
owencjones, line 3: This being unused means this file is missing a run block at the bottom, so it'll only run in set.  Not the end of the world, but worthy of note.

darwin/future/tests/meta/objects/test_datasetmeta.py
```diff
@@ -74,128 +57,41 @@ def test_create_dataset_returns_dataset_created_if_dataset_created(base_config:
...
-    exceptions, _ = Dataset.delete_dataset(valid_client, "test_dataset")
-
-    assert exceptions is None
-    assert _delete_by_slug_mock.call_count == 1
-    assert _delete_by_id_mock.call_count == 0
-
-
-def test_delete_dataset_calls_delete_by_id_as_appropriate(
-    base_config: DarwinConfig, _delete_by_id_mock: Mock, _delete_by_slug_mock: Mock
+@mark.parametrize(
```
owencjones, line 60: 👍🏻

PR #686, opened 2023-10-13
[Ai-1263] darwin py converts polygon annotations incorrectly

ChristofferEdlund:
# Problem
Darwin-py's converter to V1 format from V2 converts polygons incorrectly, since all annotations used by gust utilizes V1 format (at least for now, we should migrate) the annotations become faulty for our models.

Affected annotation types are polygon and complex_polygon. 

What is wrong? 

Tags such as bounding_box gets places inside of the polygon tag in the dictionary.

How it should look like:

```
correct_darwin_v1_annotation_dict = {
  "polygon" : {"path" : x, y ....}
  "bounding_box" : {"x":x, "y": y, ...}
  "tag" : {...}
}

```
How it currently is exported as:
```

wrong_darwin_v1_annotation_dict = {
  "polygon" : {
        "path" : x, y ....}
        "bounding_box" : {"x":x, "y": y, ...}
         "tag" : {...}
  }
}
```

Why does this matter? The dataloaders now have to infer the bounding_box from polygons, but the data is already there.

# Solution
Update the `darwin_v_1_0.py` to better handle the conversion

# Changelog
- Fixed bug in to-darwin-v1-convertion of annotations. Will affect `darwin convert darwin_1_0 v7_dataset_folder target_folder`


ChristofferEdlund (commented):
Manual testing has been done. Downloaded v2 datasets, converted to V1 and tested to use our torch dataloaders for the following classes.

- tag
- bounding_box
- polygon

owencjones (approved):

PR #688, opened 2023-10-17
tweaks to older code checks

Nathanjp91:
# Problem
typechecking should ignore older code for now until better system can be merged

# Solution
typechecking no longer enforces on older code

# Changelog
typechecking no longer enforces on older code

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ warn_untyped_fields = true
 [tool.ruff]
 select = ["E", "F", "C"]
 ignore = ["E203", "E402"]
-line-length = 160
+line-length = 88
 
 [tool.ruff.per-file-ignores]
 "__init__.py" = ["E402", "F401"]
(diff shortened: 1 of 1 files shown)


deploy/_filter_files.py
```diff
@@ -11,7 +11,7 @@ def main(argv: List[str]) -> None:
     if file_extension.startswith("."):
```
owencjones, line 11: This file is used by other steps as well as typechecking, so will affect those too - is that deliberate?

Nathanjp91: Yeah this is meant to stop it forcing people to rewrite a whole file as per ruff/black guidelines for code in the older modules, as our discussion on thursday/friday.

pyproject.toml
```diff
@@ -57,16 +57,15 @@ warn_untyped_fields = true
 [tool.ruff]
 select = ["E", "F", "C"]
 ignore = ["E203", "E402"]
-line-length = 160
+line-length = 88
 
 [tool.ruff.per-file-ignores]
 "__init__.py" = ["E402", "F401"]
-"path/to/file.py" = ["E402"]
```
Nathanjp91, line 64: What was this even doing here?

owencjones (approved):
Approved after discussion

PR #690, opened 2023-10-18
[IO-1496][external] Expose the name, type, and edge_ids of Stage objects

JBWilkie:
# Problem
`Stage` objects currently don't expose their `name`, `type`, or `edge_id`s

# Solution
Write `Stage` class properties to expose these fields

# Changelog
Added the ability to access the name, type, and edge_ids of a Stage object


darwin/future/tests/meta/objects/test_stagemeta.py
```diff
@@ -81,3 +81,15 @@ def test_move_attached_files_to_stage(
...
+def test_get_stage_id(stage_meta):
+    assert stage_meta.id == UUID("00000000-0000-0000-0000-000000000000")
+
+def test_get_stage_name(stage_meta):
+    assert stage_meta.name == "test-stage"
+
+def test_get_stage_type(stage_meta):
+    assert stage_meta.type == "annotate"
+
+def test_get_stage_edges(stage_meta):
```
JBWilkie, line 94: I'm concerned about the `test_get_stage_edges` test. The test uses the `stage_meta` fixture, which has `edges=[]`, but this doesn't actually check that the code in the edges property works, since it loops over each set of IDs within edges

Should the fixture be adjusted to include a more fleshed out structure of UUIDs, or should a new edges fixture be defined that can be passed into `stage_meta`, or is the test fine as is?

owencjones: Try making the test instantiate an object of type `Stage` and then setting and confirming the bits you've added, to show they carry.  It's not going to do much, but it confirms that you've done the change the ticket requires.

```python
def test_confirm_stage_edges():
    test_stage = Stage(
        # Other members set above
        edges=[
            [], # put edge data in this
            [], # and this
        ]
    )

    assert len(test_stage.edge) == 2
    # some other checks of the data here maybe
```

darwin/future/meta/objects/stage.py
```diff
@@ -50,4 +50,23 @@ def move_attached_files_to_stage(self, new_stage_id: UUID) -> Stage:
...
+
+    @property
+    def type(self) -> str:
+        """Stage type."""
+        return self._element.type.value
+
+    @property
+    def edges(self) -> List[List[UUID]]:
+        """Edge ID, source stage ID, target stage ID."""
+        edges = []
```
owencjones, line 69: Not obligatory, but you could do this with a list comprehension, which are a bit faster, and worth getting used to.
```python
    return [
        [edge.id, edge.source_stage_id, edge.target_stage_id]
        for edge in in self._element.edges
    ]
```

Optional, but consider it, and also have a think about listcomps in general.

owencjones (commented):
I've not approved yet, but this is good.  If you DM me when you've considered all the stuff, then I'll approve.

darwin/future/tests/meta/objects/test_stagemeta.py
owencjones: Yeah, looking at that, that's really good as an improvement.  Nicely done.

owencjones (approved):
LGTM after PR changes

PR #691, opened 2023-10-19
[IO-1278][IO-1277] Core Items/Folders apis

Nathanjp91:
# Problem
No ability to retrieve or manage items and folders

# Solution
- Introduce basic item object
- backend calls for item and folders endpoints

# Changelog
Beggining of support for items object, including apis

### DIFF
--- a/darwin/future/data_objects/item.py
+++ b/darwin/future/data_objects/item.py
@@ -1,0 +1,71 @@
+# @see: GraphotateWeb.Schemas.DatasetsV2.ItemRegistration.ExistingItem
+from typing import Dict, List, Literal, Optional, Union
+from uuid import UUID
+
+from pydantic import Field, validator
+
+from darwin.datatypes import NumberLike
+from darwin.future.data_objects.pydantic_base import DefaultDarwin
+from darwin.future.data_objects.typing import UnknownType
+
+ItemFrameRate = Union[NumberLike, Literal["native"]]
+
+
+def validate_no_slashes(v: UnknownType) -> str:
+    assert isinstance(v, str), "Must be a string"
+    assert len(v) > 0, "cannot be empty"
+    assert r"^[^/].*$".find(v) == -1, "cannot start with a slash"
+
+    return v
+
+
+class ItemSlot(DefaultDarwin):
+    # GraphotateWeb.Schemas.DatasetsV2.ItemRegistration.ExistingSlot
+
+    # Required fields
+    slot_name: str
+    file_name: str
+
+    # Optional fields
+    storage_key: Optional[str]
+    as_frames: Optional[bool] = Field(None, alias="asFrames")
+    extract_views: Optional[bool] = Field(None, alias="extractViews")
+    fps: Optional[ItemFrameRate] = Field(None, alias="fps")
+    metadata: Optional[Dict[str, UnknownType]] = Field({}, alias="metadata")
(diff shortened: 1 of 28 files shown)


darwin/future/meta/objects/stage.py
```diff
@@ -66,7 +66,9 @@ def type(self) -> str:
     @property
     def edges(self) -> List[List[UUID]]:
         """Edge ID, source stage ID, target stage ID."""
+        assert self._element.edges is not None
         return [
             [edge.id, edge.source_stage_id, edge.target_stage_id]
             for edge in self._element.edges
+            if edge.source_stage_id and edge.target_stage_id
         ]
```
Nathanjp91, line 74: @JBWilkie source_stage_id and target_stage_id are optional so they can be None. Which means this comprehension without the if check actually returns a List[List[UUID | None]]. Let me know if this change doesn't make sense though and you want the None's

darwin/future/data_objects/item.py
owencjones, line 1: I like this stuff especially 😉

darwin/future/data_objects/item.py
```diff
@@ -0,0 +1,78 @@
...
+
+def validate_no_slashes(v: UnknownType) -> str:
+    assert isinstance(v, str), "Must be a string"
+    assert len(v) > 0, "cannot be empty"
+    assert r"^[^/].*$".find(v) == -1, "cannot start with a slash"
+
+    return v
+
+
+class ItemSlot(DefaultDarwin):
```
owencjones, line 22: This is the main one I've changed in my branch

darwin/future/data_objects/item.py
```diff
@@ -0,0 +1,78 @@
...
+        assert isinstance(v, (int, float, str)), "fps must be a number or 'native'"
+        if isinstance(v, (int, float)):
+            assert v >= 0.0, "fps must be a positive number"
+        if isinstance(v, str):
+            assert v == "native", "fps must be 'native' or a number greater than 0"
+        return v
+
+
+class Item(DefaultDarwin):
+    name: str
```
owencjones, line 59: And this is where we'll have to figure it out between us.

darwin/future/tests/core/items/fixtures.py
```diff
@@ -3,6 +3,53 @@
...
+            unfiltered_item_count=1,
+        )
+        for i in range(10)
+    ]
+
+
+@pytest.fixture
+def base_items_json(base_items: List[Item]) -> List[dict]:
+    items = [item.dict() for item in base_items]
+    # json library doesn't support UUIDs so need to be str'd
```
owencjones, line 43: Yeah, what's with this?  Every time I hit this, I'm tempted to write:

```python
class JsonableUUID(UUID):
    ...fix for this...
```

darwin/future/meta/objects/stage.py
owencjones: Can't see a use case for why the `None`s would be needed but you never know

owencjones (approved):
Looking good, hold off merge until we've worked out data structs in call later, and potentially until @JBWilkie responds on the points about none filtering, although I don't think that's as blocking.

darwin/future/meta/objects/stage.py
JBWilkie: I actually changed this so that it returns the `WFEdgeCore` items instead of the IDs within them, not sure why I didn't do this originally

This means we don't need to worry about this if check

PR #693, opened 2023-10-19
[IO-1276] 3-step upload

owencjones:
# Problem
Currently, we use the old code for upload

# Solution
Implemented DTOs and functions to upload using the V2 API.

# Changelog
Added item uploads functions
Added item DTOs and validators
Added relevant tests


Nathanjp91 (approved):
LGTM as per our prior call. Like the streamlining of the ItemLayout into singular object and think we should use this approach where we can rather than having seperate objects for everything. It means more up front code in the validators I think but means later code won't have to require so many checks on different objects and Union[v1, v2] typing.

pyproject.toml
```diff
@@ -56,7 +56,7 @@ warn_untyped_fields = true
 
 [tool.ruff]
 select = ["E", "F", "C"]
-ignore = ["E203", "E402"]
+ignore = ["E203", "E402", "E501"]
```
owencjones, line 59: This allows us to leave this rule to mypy

PR #694, opened 2023-10-19
[IO-1463][external] __str__ & __repr__ dunders for Meta objects

JBWilkie:
# Problem
There's currently no __str__ & __repr__ overrides for meta objects. This prevents users from quickly seeing useful information about these objects

# Solution
- Wrote generalisable __str__ & __repr__ methods for the MetaBase class from which they all inherit
- Began writing tests for these functions but this is WIP

# Changelog
Implemented __str__ & __repr__ overrides for Meta objects.


darwin/future/meta/objects/base.py
```diff
@@ -21,4 +21,43 @@ def __init__(
...
+- Dataset Slug: {self._element.slug}\n\
+- Dataset ID: {self._element.id}\n\
+- Dataset Releases: {self._element.releases if self._element.releases else 'No releases'}"
+
+        elif class_name == "Workflow":
+            return f"Workflow\n\
+- Workflow Name: {self._element.name}\n\
+- Workflow ID: {self._element.id}\n\
+- Connected Dataset ID: {self._element.dataset.id}\n\
+- Conneted Dataset Name: {self._element.dataset.name}"
```
JBWilkie, line 51: I thought about checking for a disconnected workflow - But it seems that when we create a workflow we raise an error if it's disconnected, so it should never happen here

JBWilkie:
As a general query on testing - I wrote tests for dataset, stage, and team. These use pre-existing object_meta fixtures

As far as I could see, no fixtures of this type were available for workflows or team members yet. I had a go at writing them but started to go down a bit of a rabbit hole. I'd appreciate a quick walk-through on how this might be done

darwin/future/meta/objects/base.py
```diff
@@ -21,4 +21,44 @@ def __init__(
...
+
+        elif class_name == "Stage":
+            return f"Stage\n\
+- Stage Name: {self._element.name}\n\
+- Stage Type: {self._element.type.value}\n\
+- Stage ID: {self._element.id}"
+
+        else:
+            raise Exception(f"Class name {class_name} not found in __str__ method")
```
owencjones, line 61: Might be useful to add a `pprint` of anything that makes it to this point.  I'm on the fence whether this should raise, because the most likely case of this happenning is if we forget to implement a new one.  I guess we'd find out quickly, though.

owencjones (approved):
Yeah, LGTM, one consider point, no change needed

darwin/future/meta/objects/base.py
JBWilkie: Good point, I nearly didn't include this.

Will just pprint anything that reaches it w/out raising

PR #696, opened 2023-10-23
fixed the siutation when a user pass only a image

ChristofferEdlund:
# Problem
One should be able to user the albumentations transform in this way:
```

transform = AlbumentationsTransform(transform_compse())
img_tensor = transform(img)

```

Right now the transform output's a (img, target) tuple even when only an image is passed.

# Solution

Add a check if there is only an image as input, and return only image if that is the case.

# Changelog
- albumentation transform now returns only img and not (img, target) when only an image is present as input

### DIFF
--- a/darwin/torch/transforms.py
+++ b/darwin/torch/transforms.py
@@ -332,7 +332,7 @@ class AlbumentationsTransform:
         transformed_data = self.transform(**albu_data)
         image, transformed_annotation = self._post_process(transformed_data, annotation)
 
-        if list(albu_data.keys()) == ["image"]:
+        if len(annotation) < 1:
             return image
 
         return image, transformed_annotation
--- a/darwin/torch/transforms.py
+++ b/darwin/torch/transforms.py
@@ -332,6 +332,9 @@ class AlbumentationsTransform:
         transformed_data = self.transform(**albu_data)
         image, transformed_annotation = self._post_process(transformed_data, annotation)
 
+        if list(albu_data.keys()) == ["image"]:
+            return image
+
         return image, transformed_annotation
 
     def _pre_process(self, image: np.ndarray, annotation: dict) -> dict:
--- a/darwin/torch/transforms.py
+++ b/darwin/torch/transforms.py
@@ -143,7 +143,9 @@ class ColorJitter(transforms.ColorJitter):
     def __call__(
         self, image: PILImage.Image, target: Optional[TargetType] = None
     ) -> Union[PILImage.Image, Tuple[PILImage.Image, TargetType]]:
-        transform = self.get_params(self.brightness, self.contrast, self.saturation, self.hue)
+        transform = self.get_params(
+            self.brightness, self.contrast, self.saturation, self.hue
+        )
         image = transform(image)
         if target is None:
             return image
@@ -198,7 +200,9 @@ class ConvertPolygonsToInstanceMasks(object):
     Converts given polygon to an ``InstanceMask``.
     """
 
-    def __call__(self, image: PILImage.Image, target: TargetType) -> Tuple[PILImage.Image, TargetType]:
+    def __call__(
+        self, image: PILImage.Image, target: TargetType
+    ) -> Tuple[PILImage.Image, TargetType]:
         w, h = image.size
 
         image_id = target["image_id"]
@@ -255,7 +259,9 @@ class ConvertPolygonsToSemanticMask(object):
     Converts given polygon to an ``SemanticMask``.
     """
 
-    def __call__(self, image: PILImage.Image, target: TargetType) -> Tuple[PILImage.Image, TargetType]:
+    def __call__(
+        self, image: PILImage.Image, target: TargetType
+    ) -> Tuple[PILImage.Image, TargetType]:
         w, h = image.size
         image_id = target["image_id"]
         image_id = torch.tensor([image_id])
@@ -282,7 +288,9 @@ class ConvertPolygonToMask(object):
     Converts given polygon to a ``Mask``.
     """
 
-    def __call__(self, image: PILImage.Image, annotation: Dict[str, Any]) -> Tuple[PILImage.Image, PILImage.Image]:
+    def __call__(
+        self, image: PILImage.Image, annotation: Dict[str, Any]
+    ) -> Tuple[PILImage.Image, PILImage.Image]:
         w, h = image.size
         segmentations = [obj["segmentation"] for obj in annotation]
         cats = [obj["category_id"] for obj in annotation]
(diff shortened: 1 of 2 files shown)


tests/darwin/exporter/formats/export_darwin_1_0_test.py
dorfmanrobert, line 1: side question regarding these tests (do know this isn't part of the PR): isn't "paths" the new "complex_polygon"?

darwin/torch/transforms.py
```diff
@@ -332,6 +340,9 @@ def __call__(self, image, annotation: dict = None) -> tuple:
         transformed_data = self.transform(**albu_data)
         image, transformed_annotation = self._post_process(transformed_data, annotation)
 
+        if len(annotation) < 1:
+            return image
```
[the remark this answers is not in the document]
ChristofferEdlund: It is how other transforms seems to be called.


    def __getitem__(self, index: int) -> Tuple[Tensor, Tensor]:
        """
        See superclass for documentation.

        Parameters
        ----------
        index : int
            The index of the image.

        Returns
        -------
        Tuple[Tensor, Tensor]
            A tuple of tensors, where the first value is the image tensor and the second is the
            target's tensor.
        """
        img: PILImage.Image = self.get_image(index)
        if self.transform is not None:
            img_tensor = self.transform(img)
        else:
            img_tensor = to_tensor(img)

        target = self.get_target(index)

        return img_tensor, target


https://github.com/v7labs/darwin-py/blob/5b7ad8b4064eac4672cbe950f40855c2be2ea009/darwin/torch/dataset.py#L113C10-L113C10

So lets enable that behavior in our AlbumentationsTransform class as well.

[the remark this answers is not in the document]
dorfmanrobert: maybe could default the value to None and use that instead of len() > 1?

tests/darwin/exporter/formats/export_darwin_1_0_test.py
dorfmanrobert: gotcha thanks

dorfmanrobert (approved):

darwin/torch/transforms.py
```diff
@@ -349,7 +363,7 @@ def _pre_process(self, image: np.ndarray, annotation: dict) -> dict:
             albumentation_dict["labels"] = labels.tolist()
 
         masks = annotation.get("masks")
-        if masks is not None:
+        if masks is not None and masks.numel() > 0:  # using numel() to check if tensor is non-empty
```
dorfmanrobert, line 366: if masks is None this will error

darwin/torch/transforms.py
```diff
@@ -364,8 +378,15 @@ def _post_process(self, albumentation_output: dict, annotation: dict) -> tuple:
         bboxes = albumentation_output.get("bboxes")
         if bboxes is not None:
             output_annotation["boxes"] = torch.tensor(bboxes)
+
             if "area" in annotation and "masks" not in albumentation_output:
```
dorfmanrobert, line 382: just wondering, why is it necessary to check for "masks" not being in the dict here?

darwin/torch/transforms.py
dorfmanrobert, line 1: what do you think abt moving the "area" logic outside the logic of boxes and masks in `_post_process()`? might make things cleaner

darwin/torch/transforms.py
```diff
@@ -378,7 +399,13 @@ def _post_process(self, albumentation_output: dict, annotation: dict) -> tuple:
             else:
                 output_annotation["masks"] = torch.stack(masks)
             if "area" in annotation:
-                output_annotation["area"] = torch.sum(output_annotation["masks"], dim=[1, 2])
+                output_annotation["area"] = torch.sum(
+                    output_annotation["masks"], dim=[1, 2]
+                )
+        elif "masks" in annotation:
```
dorfmanrobert, line 405: what is this checking for/why is it needed?

darwin/torch/transforms.py
dorfmanrobert: think the logic is this currently:

```
If "area" in annotation:
	if bboxes is not None and "masks" not in albumentation_output:
		if len(bboxes) > 0:
			output_annotation["area"] = (
                        output_annotation["boxes"][:, 2] * output_annotation["boxes"][:, 3]
                    )
		else:
			output_annotation["area"] = []
		
	elif masks is not None:
		output_annotation["area"] = torch.sum(
                    output_annotation["masks"], dim=[1, 2]
                )
                
	elif "masks" in annotation:
		output_annotation["area"] = []
```

darwin/torch/transforms.py
ChristofferEdlund: Python evaluates from left to right, since it is a AND and not an OR it will check if it is None and then return false (since it does not need to check the statement to the right).

darwin/torch/transforms.py
ChristofferEdlund: If masks are present, area calculated from the masks are more accurate then an area from bounding boxes.

darwin/torch/transforms.py
ChristofferEdlund: Great, i like it!

darwin/torch/transforms.py
dorfmanrobert: oh nice good to know!

darwin/torch/transforms.py
ChristofferEdlund: We needed to remove the masks tag to get around an albumentation bug, but we add it back again with an empty list if there are no annotations. This is so any code on the user side that assumes there to be a mask key wont break. (It might break for empty masks, but at least not a keyerror).

darwin/torch/transforms.py
```diff
@@ -377,8 +394,20 @@ def _post_process(self, albumentation_output: dict, annotation: dict) -> tuple:
                 output_annotation["masks"] = torch.tensor(np.array(masks))
             else:
                 output_annotation["masks"] = torch.stack(masks)
-            if "area" in annotation:
-                output_annotation["area"] = torch.sum(output_annotation["masks"], dim=[1, 2])
+        elif "masks" in annotation:
+            output_annotation["masks"] = torch.tensor([])
+
+        if "area" in annotation:
+            if "masks" in output_annotation and output_annotation["masks"].numel() > 0:
```
dorfmanrobert, line 401: stylistic choice: could use `masks` and `bboxes` variables since have them but up to you

dorfmanrobert:
left one stylistic comment abt using existing variables or not that I leave up to you, otherwise all good!

darwin/torch/transforms.py
ChristofferEdlund: I think we keep at as it is, masks and bboxes are not the same since we do some conversion.  masks for example is a list, and does not have .numel

PR #697, opened 2023-10-23
[IO-1800] Set priority

owencjones:
# Problem
No current access to setting priority of items

# Solution
Implemented priority setter.

# Changelog
Priority setting method in core
Tests

### DIFF
--- a/darwin/future/core/items/set_priority.py
+++ b/darwin/future/core/items/set_priority.py
@@ -1,0 +1,43 @@
+from typing import Dict, Optional
+
+from darwin.future.core.client import ClientCore
+from darwin.future.core.types.common import JSONType
+from darwin.future.data_objects.typing import UnknownType
+
+
+def set_item_priority(
+    api_client: ClientCore,
+    team_slug: str,
+    priority: int,
+    filters: Optional[Dict[str, UnknownType]] = None,
+) -> JSONType:
+    """
+    Sets the priority of an item
+
+    Parameters
+    ----------
+    client: Client
+        The client to use for the request
+    team_slug: str
+        The slug of the team to set the priority for
+    priority: int
+        The priority to set
+
+    Returns
+    -------
+    JSONType
+    """
+    payload: Dict["str", UnknownType] = {
+        "priority": priority,
+    }
+
+    if filters:
+        payload = {
+            **payload,
+            "filters": filters,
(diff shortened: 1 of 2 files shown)


darwin/future/core/items/set_priority.py
```diff
@@ -0,0 +1,43 @@
...
+from darwin.future.core.client import ClientCore
+from darwin.future.core.types.common import JSONType
+from darwin.future.data_objects.typing import UnknownType
+
+
+def set_item_priority(
+    api_client: ClientCore,
+    team_slug: str,
+    priority: int,
+    filters: Optional[Dict[str, UnknownType]] = None,
```
Nathanjp91, line 12: This is not blocking at all but interested here on your thoughts on 
`filters: Optional[...] = None` vs `filters: ... = {}` Is it more pythonic to assume None's or empty objects? The former does tend to lead to requiring more typecheck guards to satisfy mypy/pylance with like `assert filters is not None` etc, which does feel more verbose than required sometimes.

darwin/future/tests/core/items/test_set_priority.py
```diff
@@ -0,0 +1,93 @@
...
+import pytest
+
+from darwin.exceptions import DarwinException
+from darwin.future.core.client import ClientCore
+from darwin.future.core.items.set_priority import set_item_priority
+
+
+def test_set_item_priority() -> None:
+    # Create a mock API client
+    api_client = Mock(spec=ClientCore)
```
Nathanjp91, line 12: We do have the response library for this, although not at issue by any means.

Nathanjp91 (approved):
Nothing blocking

darwin/future/core/items/set_priority.py
Nathanjp91: but obviously, it is a bit more clear that filters isn't required via the `Optional[...] = None` approach

owencjones: Nah, good point, I like the default in this backend code, will take this suggestion

darwin/future/tests/core/items/test_set_priority.py
owencjones: Yeah, actually I'll rewrite with that in order to maintain our test style

PR #699, opened 2023-10-24
[IO-1800][internal] Set Priority of items

owencjones:
# Problem
The initial set priority had opportunity for some improvements

# Solution
Made the improvements

### DIFF
--- a/darwin/future/core/items/set_item_priority.py
+++ b/darwin/future/core/items/set_item_priority.py
@@ -1,4 +1,5 @@
 from typing import Dict, Optional
+from uuid import UUID
 
 from darwin.future.core.client import ClientCore
 from darwin.future.core.types.common import JSONType
@@ -8,8 +9,10 @@ from darwin.future.data_objects.typing import UnknownType
 def set_item_priority(
     api_client: ClientCore,
     team_slug: str,
+    dataset_id: int,
+    item_id: UUID,
     priority: int,
-    filters: Optional[Dict[str, UnknownType]] = None,
+    filters: Dict[str, UnknownType] = {},
 ) -> JSONType:
     """
     Sets the priority of an item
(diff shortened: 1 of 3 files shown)


darwin/future/core/items/set_item_priority.py
```diff
@@ -0,0 +1,45 @@
...
+        The priority to set
+
+    Returns
+    -------
+    JSONType
+    """
+    payload = {
+        "priority": priority,
+        "filters": {
+            "item_ids": [str(item_id)],
```
JBWilkie, line 36: I think the [set priority endpoint](https://docs.v7labs.com/reference/items-set-priority) can accept a list of `item_ids`, would it made sense to allow this to as well?

If so, maybe:
- `set_item_priority()` accepts `item_ids: List[UUID]`, then:
- `"item_ids": [str(item_id) for item_id in item_ids]` at this point

darwin/future/tests/core/items/test_set_priority.py
```diff
@@ -0,0 +1,88 @@
...
+            "item_ids": ["00000000-0000-0000-0000-000000000000"],
+            "status": "open",
+        },
+    }
+
+    # Define the expected endpoint
+    expected_endpoint = "/v2/teams/test-team/items/priority"
+
+    # Define the expected response
+    expected_response = {"status": "success"}
```
JBWilkie, line 48: I might be missing something, but if successful doesn't the endpoint always return `{"affected_item_count": n}`?

I couldn't get it to return `{"status": "success"}`

JBWilkie (changes requested):

darwin/future/tests/core/items/test_set_priority.py
owencjones: If you pass it a 422 inducing load of rubbish, it returns a response including `errors`

`Graphotate.Common.APIError` example:
```
{
  "errors": {
    "code": "EXTERNAL_STORAGE_LIMIT_REACHED",
    "detail": {},
    "message": "Error message",
    "status": 422
  }
}```

JBWilkie (approved):

PR #701, opened 2023-10-24
[IO-1461][IO-1820] Code Documentation + exception refactoring + mypy checking

Nathanjp91:
# Problem
Some undocumented code and dogfooding issues have accrued

# Solution
fix the following:
- added documentation as well as examples
- refactored and unified approach to exception handling
- swapped str/repr locations to conform with mypy typechecking

The List[Exception] Paradigm now only exists for things that parse multiple objects. It should raise on other issues like exceptions from underlying http request, but will attempt to parse all objects before failing. 

# Changelog
Added documentation and dogfooding refactors to futures code

### DIFF
--- a/darwin/future/exceptions.py
+++ b/darwin/future/exceptions.py
@@ -1,4 +1,6 @@
-from typing import List, Optional
+from __future__ import annotations
+
+from typing import Optional, Sequence
 
 from darwin.future.data_objects.typing import KeyValuePairDict, UnknownType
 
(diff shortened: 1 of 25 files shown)


darwin/future/meta/queries/stage.py
```diff
@@ -14,7 +14,7 @@ def _collect(self) -> List[Stage]:
             raise ValueError("Must specify workflow_id to query stages")
         workflow_id: UUID = self.meta_params["workflow_id"]
         meta_params = self.meta_params
-        workflow, exceptions = get_workflow(self.client, str(workflow_id))
+        workflow = get_workflow(self.client, str(workflow_id))
```
owencjones, line 17: Maybe need handling of the exceptions here?  Consideration point

owencjones (approved):
Yep, a lot of it, but good stuff.

darwin/future/exceptions.py
```diff
@@ -43,6 +45,30 @@ def from_exception(cls, exc: Exception) -> "DarwinException":
 
         return instance
 
+    @classmethod
```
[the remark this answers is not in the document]
Nathanjp91: we do in the old code version, I just copied it across and modified a little

darwin/future/meta/queries/stage.py
Nathanjp91: good point, I'll add something in. At least a print on failed validations

PR #702, opened 2023-10-26
[IO-1801][external] Core item archive method & tests

JBWilkie:
# Problem
Currently no way to archive items in `darwin.future`

# Solution
Define the core method to archive a list of items by their UUIDs by using the archive endpoint

# Changelog
Added core item archive method & tests

### DIFF
--- a/darwin/future/core/client.py
+++ b/darwin/future/core/client.py
@@ -10,7 +10,7 @@ from pydantic import BaseModel, root_validator, validator
 from requests.adapters import HTTPAdapter, Retry
 
 from darwin.future.core.types.common import JSONType, QueryString
-from darwin.future.exceptions import NotFound, Unauthorized
+from darwin.future.exceptions import NotFound, Unauthorized, UnprocessibleEntity
 
 
 class TeamsConfig(BaseModel):
@@ -245,3 +245,5 @@ def raise_for_darwin_exception(response: requests.Response) -> None:
         raise Unauthorized(response)
     if response.status_code == 404:
         raise NotFound(response)
+    if response.status_code == 422:
+        raise UnprocessibleEntity(response)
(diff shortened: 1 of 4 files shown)


darwin/future/core/client.py
```diff
@@ -245,3 +245,5 @@ def raise_for_darwin_exception(response: requests.Response) -> None:
         raise Unauthorized(response)
     if response.status_code == 404:
         raise NotFound(response)
+    if response.status_code == 422:
```
JBWilkie, line 248: We may want to surface / deal / pprint with 422s later, especially if because the endpoints generally do a good job of pointing out exactly **what** causes 422s

darwin/future/tests/core/items/test_archive_items.py
```diff
@@ -0,0 +1,54 @@
...
+from darwin.future.tests.core.fixtures import *
+
+
+@responses.activate
+def test_archive_items_including_filters(base_client: ClientCore) -> None:
+    # Define the expected response
+    responses.add(
+        responses.POST,
+        base_client.config.api_endpoint + "v2/teams/test-team/items/archive",
+        json={"affected_item_count": 0},
```
owencjones, line 19: Really minor point, but your affect item count here is mocked as 0, when if successful, it would've been 2.

Honestly, only fix if you want to.

darwin/future/tests/core/items/test_archive_items.py
```diff
@@ -0,0 +1,54 @@
...
+            "not_statuses": ["uploading", "annotate"],
+            "not_assignees": [123, 456, 789],
+        },
+    )
+
+    # Check that the response mathces what we expect
+    assert response == {"affected_item_count": 0}
+
+
+def test_archive_items_with_error_response(base_client: ClientCore) -> None:
```
owencjones, line 41: `base_client` fixture is included but not used here.

owencjones (changes requested):
Only one change, and it's really minor lint change at the bottom - this is good stuff mate.

owencjones (approved):

PR #704, opened 2023-10-29
[CSE-10][external] Fixes for LocalDataset object instantiation

JBWilkie:
# Problem
- 1: When dataset items in folders are pulled in a flat structure, `LocalDataset` instantiation and the `get_annotations()` function doesn't work because we are constructing & checking image paths on the assumption they were pulled with folders
- 2: Changes introduced in IO-1445 broke Darwin JSON 1.0 compatibility with LocalDataset objects & get_annotations() in some scenarios

Much more problem detail is available in a comment in [CSE-10](https://linear.app/v7labs/issue/CSE-10/multiple-stems-error-when-using-get-annotations)

# Solution
- 1: When instantiating a `LocalDataset` object or using `get_annotations()`, check if the release was pulled with folders and pass this down to the image path construction
- 2: When instantiating a `LocalDataset` object or using `get_annotations()`, check if the release is Darwin JSON 1.0 or 2.0, then pass this down to the relevant function(s)

Much more solution detail is available in a comment in [CSE-10](https://linear.app/v7labs/issue/CSE-10/multiple-stems-error-when-using-get-annotations)

# Changelog
- Fixed false assumption that dataset releases with folders are always pulled with folders. This assumption led to issues creating LocalDataset objects and with the use of the get_annotations() function
- Fixed issue where Darwin JSON 1.0 was not compatible with LocalDataset objects of the get_annotations() function


JBWilkie:
**Note:** This has failing tests at the moment, I need to:
- 1: Resolve them - There appears to be an issue where one test is loading a V2 dataset of 20 items, but one of the items is actually a V1 item
- 2: Resolve an issue with json_stream for video files with large frame_url fields. There appears to be an issue where if frame_url exceeds a certain size, then the stream breaks

I will raise this with someone soon to unblock the work

darwin/client.py
```diff
@@ -908,7 +908,7 @@ def move_to_stage(
         dataset_slug: str,
         team_slug: str,
         filters: Dict[str, UnknownType],
-        stage_id: int,
+        stage_id: str,
```
JBWilkie, line 911: Unrelated to the original ticket, but `stage_id` is `str`

PR #705, opened 2023-10-31
[IO-2040][external] Fix for NifTI exports that include filenames with dots

JBWilkie:
# Problem
As described in [IO-2040](https://linear.app/v7labs/issue/IO-2040/bug-there-is-an-issue-exporting-nifti-that-is-not-an-issue-for-darwin), NifTI exports of files with filenames containing dots would fail due to there being too many suffixes

# Solution
Add logic to extract the last 2 suffixes of each filename while preserving the existing error checking logic

# Changelog
 Fixed a bug preventing NifTI exports that include filenames with dots

### DIFF
--- a/darwin/exporter/formats/nifti.py
+++ b/darwin/exporter/formats/nifti.py
@@ -331,10 +331,7 @@ def write_output_volume_to_disk(
                 img_ornt, orig_ornt
             )  # Get transform from RAS to current affine
             img = img.as_reoriented(from_canonical)
-        output_path = (
-            Path(output_dir)
-            / f"{image_id}_{volume.series_instance_uid}_{volume.class_name}.nii.gz"
-        )
+        output_path = Path(output_dir) / f"{image_id}_{volume.class_name}.nii.gz"
         if not output_path.parent.exists():
             output_path.parent.mkdir(parents=True)
         nib.save(img=img, filename=output_path)
--- a/darwin/exporter/formats/nifti.py
+++ b/darwin/exporter/formats/nifti.py
@@ -63,8 +63,12 @@ def export(annotation_files: Iterable[dt.AnnotationFile], output_dir: Path) -> N
         output_volumes = build_output_volumes(video_annotation)
         slot_map = {slot.name: slot for slot in video_annotation.slots}
         for annotation in video_annotation.annotations:
-            populate_output_volumes(annotation, output_dir, slot_map, output_volumes, image_id)
-        write_output_volume_to_disk(output_volumes, image_id=image_id, output_dir=output_dir)
+            populate_output_volumes(
+                annotation, output_dir, slot_map, output_volumes, image_id
+            )
+        write_output_volume_to_disk(
+            output_volumes, image_id=image_id, output_dir=output_dir
+        )
 
 
 def build_output_volumes(video_annotation: dt.AnnotationFile) -> Dict:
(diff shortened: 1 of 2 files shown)


darwin/exporter/formats/nifti.py
```diff
@@ -135,16 +143,19 @@ def check_for_error_and_return_imageid(video_annotation: dt.AnnotationFile, outp
...
-    if len(suffixes) > 2:
-        return create_error_message_json(
-            "Misconfigured filename, contains too many suffixes", output_dir, str(filename)
-        )
-    elif len(suffixes) == 2:
+    try:
+        suffixes = filename.suffixes[-2:]
+    except IndexError:
+        suffixes = filename.suffixes
+    if len(suffixes) == 2:
```
JBWilkie, line 150: Extract the lasst 2 suffixes, which should be `.nii` and `.gz`, but handle the case that there are <2 suffixes

darwin/exporter/formats/nifti.py
```diff
@@ -290,11 +323,15 @@ def unnest_dict_to_list(d: Dict) -> List:
...
+            orig_ornt = io_orientation(
+                volume.original_affine
+            )  # Get orientation of current affine
             img_ornt = io_orientation(volume.affine)  # Get orientation of RAS affine
-            from_canonical = ornt_transform(img_ornt, orig_ornt)  # Get transform from RAS to current affine
+            from_canonical = ornt_transform(
+                img_ornt, orig_ornt
+            )  # Get transform from RAS to current affine
             img = img.as_reoriented(from_canonical)
-        output_path = Path(output_dir) / f"{image_id}_{volume.series_instance_uid}_{volume.class_name}.nii.gz"
```
JBWilkie, line 297: Sufficiently long filenames resulted in `OSError: [Errno 63] File name too long`, so on the [advice of Jade](https://www.loom.com/share/67a579881e8b418d927c639545779f43?sid=189cb828-b6ec-418f-9bfe-e2096d102f32), remove this UUID (which is V7-generated) from the filenames to reduce the chance of this happening

Nathanjp91: Is this breaking? Won't we be storing filename as part of an annotation that will include the UUID? Then when try to re-read this file elsewhere it won't exist.

Nathanjp91 (changes requested):
Possible issue but rest of the code looks good.

darwin/exporter/formats/nifti.py
```diff
@@ -135,16 +143,19 @@ def check_for_error_and_return_imageid(video_annotation: dt.AnnotationFile, outp
...
-        )
-    elif len(suffixes) == 2:
+    try:
+        suffixes = filename.suffixes[-2:]
+    except IndexError:
+        suffixes = filename.suffixes
+    if len(suffixes) == 2:
         if suffixes[0] == ".nii" and suffixes[1] == ".gz":
-            image_id = str(filename).strip("".join(suffixes))
+            image_id = str(filename).rstrip("".join(suffixes))
```
JBWilkie, line 152: `strip()` was removing any character in any item in suffixes from the start and end of the string

Since we expect suffixes to contain ".nii", I noticed that for example a file beginning with "nyu" would become "yu"

`rstrip()` strips only trailing characters, which will be the suffixes

Nathanjp91 (approved):

darwin/exporter/formats/nifti.py
```diff
@@ -97,7 +101,9 @@ def build_output_volumes(video_annotation: dt.AnnotationFile) -> Dict:
     for slot in video_annotation.slots:
         slot_metadata = slot.metadata
         assert slot_metadata is not None
-        series_instance_uid = slot_metadata.get("SeriesInstanceUID", "SeriesIntanceUIDNotProvided")
+        series_instance_uid = slot_metadata.get(
+            "SeriesInstanceUID", "SeriesIntanceUIDNotProvided"
```
NooshinGhavami, line 105: We don't want to remove these two then in the end? Or did you split it into a different PR?

darwin/exporter/formats/nifti.py
NooshinGhavami: Is it still handling case where there are >2 suffixes also?

NooshinGhavami (approved):

PR #707, opened 2023-11-02
[IO-1746] Introduction of Pagination queries and changes to item_ids endpoint

Nathanjp91:
# Problem
item ids collection uses an older endpoint, need to swap it over to the newer list_item_ids endpoint, which then needs to filter into changes on the meta objects that implement these. This however necessitates the creation of a solution to pagination on both the Core/Meta levels. 

# Solution
Introduce some major changes to accomodate Pagination, including
- Core data object for keeping track of pages, `darwin.future.data_objects.page.Page` which has defaults that we use on the API
- PaginatedQuery extension, which sets up pagination and changes relevant dunders to allow for pagination
  - paginated objects are lazy loading, iterable and will remember the current page, only collecting when needed
  - includes some new helper functions like collect_all to force total collection
  - still allows for indexing, `PaginatedQuery[76]` will detect the required page to send and collect results if not stored
  - `len(PaginatedQuery)` will collect_all and is something to note behaviour wise, this is a requirement to allow list(PaginatedQuery) to work, otherwise I would have made it Raise on an uncollected object
  - Changes to underlying Query object required to allow mapping, reflected in changes to dunder functions (but not behaviour, which remains the same). Results now stored as a dictionary map of {index: result} indicating the offset location of the item
- V7ID object now a wrapper for UUID
  - Required to interact with Query objects, despite being overkill on UUID list.
  - This approach will allow for other interaction later, like deletion of objects via IDs and any extensions we want to add 
- Anything that uses item_ids prior now uses ItemIDQuery, the first PaginatedQuery object

# Changelog
New Item_ids behaviour and introduction of PaginatedQuery Objects inheritable for default pagination behaviour.

### DIFF
--- a/darwin/future/core/client.py
+++ b/darwin/future/core/client.py
@@ -135,7 +135,11 @@ class ClientCore:
     team: Team, team to make requests to
     """
 
-    def __init__(self, config: DarwinConfig, retries: Optional[Retry] = None) -> None:
+    def __init__(
+        self,
+        config: DarwinConfig = DarwinConfig.local(),
+        retries: Optional[Retry] = None,
+    ) -> None:
         self.config = config
         self.session = requests.Session()
         if not retries:
--- a/darwin/future/core/client.py
+++ b/darwin/future/core/client.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from pathlib import Path
-from typing import Callable, Dict, Optional, overload
+from typing import Callable, Dict, Optional
 from urllib.parse import urlparse
 
 import requests
@@ -165,21 +165,6 @@ class ClientCore:
         if self.config.api_key:
             http_headers["Authorization"] = f"ApiKey {self.config.api_key}"
         return http_headers
-
-    @overload
-    def _generic_call(
-        self, method: Callable[[str], requests.Response], endpoint: str
-    ) -> dict:
-        ...
-
-    @overload
-    def _generic_call(
-        self,
-        method: Callable[[str, dict], requests.Response],
-        endpoint: str,
-        payload: dict,
-    ) -> dict:
-        ...
 
     def _generic_call(
         self, method: Callable, endpoint: str, payload: Optional[dict] = None
--- a/darwin/future/core/client.py
+++ b/darwin/future/core/client.py
@@ -227,7 +227,7 @@ class ClientCore:
         return self._generic_call(
             self.session.delete,
             self._contain_qs_and_endpoint(endpoint, query_string),
-            data if data is not None else {},
+            data,
         )
 
     def patch(self, endpoint: str, data: dict) -> JSONType:
--- a/darwin/future/core/types/common.py
+++ b/darwin/future/core/types/common.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 from typing import Any, Dict, List, Union
 
-from darwin.future.core.types.query import Query, QueryFilter
 from darwin.future.data_objects import validators as darwin_validators
 from darwin.future.data_objects.typing import UnknownType
 
@@ -76,9 +75,6 @@ class QueryString:
 
     value: Dict[str, str]
 
-    def from_filter(self, filter: QueryFilter) -> QueryString:
-        return QueryString({filter.name: filter.param}) 
-
     def dict_check(self, value: UnknownType) -> Dict[str, str]:
         assert isinstance(value, dict)
         assert all(isinstance(k, str) and isinstance(v, str) for k, v in value.items())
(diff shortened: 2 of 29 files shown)


darwin/future/core/client.py
```diff
@@ -166,21 +170,6 @@ def headers(self) -> Dict[str, str]:
             http_headers["Authorization"] = f"ApiKey {self.config.api_key}"
         return http_headers
 
-    @overload
```
Nathanjp91, line 169: Totally unsure why I even had this overload, it didn't matter for the typing.

darwin/future/core/types/common.py
```diff
@@ -83,3 +85,6 @@ def __init__(self, value: Dict[str, str]) -> None:
 
     def __str__(self) -> str:
         return "?" + "&".join(f"{k}={v}" for k, v in self.value.items())
+
```
Nathanjp91, line 88: Useful for combining parameters in the core functions when required for pagination which I treat seperately

darwin/future/core/types/query.py
```diff
@@ -104,6 +106,12 @@ def _from_kwarg(cls, key: str, value: str) -> QueryFilter:
             modifier = None
         return QueryFilter(name=key, param=value, modifier=modifier)
 
+    def to_dict(self, ignore_modifier: bool = True) -> Dict[str, str]:
```
Nathanjp91, line 109: Mostly so they can be swapped in and out of QueryStrings for Paginated endpoints so that .where() syntax still works, for now ignores modifiers. I think this is not the most elegant solution though as eventually I think we'll want to probably do both server side query filter and client side once collected

darwin/future/core/types/query.py
```diff
@@ -154,17 +162,16 @@ def __init__(
         self.meta_params: dict = meta_params or {}
         self.client = client
         self.filters = filters or []
-        self.results: Optional[List[T]] = None
-        self._changed_since_last: bool = True
+        self.results: dict[int, T] = {}
```
Nathanjp91, line 165: This is the core change that allows for Pagination, but all existing functions have been swapped over to preserve the outputs

darwin/future/meta/queries/item_id.py
```diff
@@ -0,0 +1,41 @@
...
+            and "dataset_id" not in self.meta_params
+        ):
+            raise ValueError("Must specify dataset_id to query item ids")
+        team_slug: str = self.meta_params["team_slug"]
+        dataset_ids: int = (
+            self.meta_params["dataset_ids"]
+            if "dataset_ids" in self.meta_params
+            else self.meta_params["dataset_id"]
+        )
+        params: QueryString = reduce(
```
Nathanjp91, line 25: How I use the aforementioned QueryString + QueryString syntax, map reduce them with the page object and all the filters

owencjones: I like that python has `itertools`, but I'd prefer a fluent interface like any sensible language ~like Rust~

Anyway, good stuff!

owencjones (approved):
Some comments to address before merge but good stuff

darwin/future/core/types/query.py
```diff
@@ -205,12 +212,12 @@ def __next__(self) -> T:
 
     def __getitem__(self, index: int) -> T:
         if not self.results:
```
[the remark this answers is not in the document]
Nathanjp91: self.collect() already accounts for it, which is a wrapper that runs the _collect function that has to be written per meta class but does all the _changed_since_last and other logic

darwin/future/core/types/query.py
```diff
@@ -242,12 +252,71 @@ def collect_one(self) -> T:
...
         return [
             m for m in objects if filter.filter_attr(getattr(m._element, filter.name))
         ]
+
+
+class PaginatedQuery(Query[T]):
+    def __init__(
+        self,
+        client: ClientCore,
+        filters: List[QueryFilter] | None = None,
```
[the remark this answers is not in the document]
Nathanjp91: I believe we can if we use the from __future__ import annotations at the top, which I do anyway for some imports around classes that return themselves.

darwin/future/core/types/query.py
```diff
@@ -154,17 +162,16 @@ def __init__(
...
     def filter(self, filter: QueryFilter) -> Query[T]:
         return self + filter
 
     def __add__(self, filter: QueryFilter) -> Query[T]:
         self._changed_since_last = True
-        return self.__class__(
-            self.client, filters=[*self.filters, filter], meta_params=self.meta_params
-        )
+        self.filters.append(filter)
+        return self
```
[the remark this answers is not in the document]
Nathanjp91: it is slightly different yeah, it is modification in place rather than re-allocation, but I think that's a much more sensible approach here as most people won't be using the __add__ dunder functionality except us anyway as internal functions, and it the modification in place saves time and memory.

PR #708, opened 2023-11-07
[PY-398] Polling on Item Push

saurbhc:
# Problem
1. Before pushing an item to the next stage, we need a way to make sure those items should have complted the processing.
2. dataset-stage might not be the 1st staged returned from the API, which raises an error when pushed to next stage.

# Solution
1. Add Pooling on Item Push (before pushing item to next stage), with an optional `wait` parameter
2. Additionally, fix dataset stage logic in Workflow obj

# Changelog
Pooling on Item Push + Fix dataset stage logic


darwin/future/meta/objects/workflow.py
```diff
@@ -55,6 +55,12 @@ def stages(self) -> StageQuery:
             meta_params["dataset_name"] = self.datasets[0].name
         return StageQuery(self.client, meta_params=meta_params)
 
+    def _get_dataset_stage(self) -> Stage | None:
```
Nathanjp91, line 58: I'm torn between an optional None return here or raising some MissingDataset Exception or similar. I think most functions that would call this would expect it to return something useful and if not then it should abort process and be captured somewhere.

Nathanjp91 (changes requested):

darwin/future/meta/objects/workflow.py
saurbhc: Yep that makes sense. Let me raise `MissingDataset` Exception when there's no Dataset Stage in a workflow (this will replace the returning `None` from this method)

darwin/future/meta/objects/stage.py
```diff
@@ -75,6 +82,18 @@ def move_attached_files_to_stage(self, new_stage_id: UUID) -> Stage:
             self.meta_params["dataset_id"],
         )
         ids = [x.id for x in self.item_ids.collect_all()]
+
+        if wait:
+            while True:
```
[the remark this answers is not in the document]
saurbhc: Adding retries makes total sense! Although, I couldn't get it working with [urllib3.util.Retry](https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html) as it dosen't have a `retry.wait()` - looks like it works best with `http.client`. For now, I've implemented something similar to this in a new commit.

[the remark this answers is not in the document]
owencjones: My take on this is that I like the NASA code rules approach:
> All loops must have fixed bounds. This prevents runaway code.

So, basically, I don't hate `while True` if the `while` block contains a max timeout or loop.

@Nathanjp91's approach is very sound, it could be recreated with a safe version using `while True`, but I'd want to see a maximum bound.

darwin/future/meta/objects/stage.py
```diff
@@ -75,6 +83,30 @@ def move_attached_files_to_stage(self, new_stage_id: UUID) -> Stage:
...
+
+        def all_items_complete(item_ids: list[UUID]) -> bool:
+            for item_id in item_ids:
+                if get_item(self.client, slug, item_id).processing_status != "complete":
+                    return False
+            return True
+
+        if wait:
+            max_attempts = 5
+            wait_time = 0.5  # seconds
```
owencjones, line 95: Magic numbers might be a good idea to abstract into optional function parameters with defaults.

darwin/future/meta/objects/stage.py
```diff
@@ -75,6 +83,30 @@ def move_attached_files_to_stage(self, new_stage_id: UUID) -> Stage:
             self.meta_params["dataset_id"],
         )
         ids = [x.id for x in self.item_ids.collect_all()]
+
+        def all_items_complete(item_ids: list[UUID]) -> bool:
+            for item_id in item_ids:
```
owencjones, line 88: I'd probably use `all` with a list comp here, as it's more performant - I don't know if we really expect this to be enough of a factor to matter, but given that it's an easy change, I'd suggest it.

```python
def all_items_complete(item_ids: list[UUID]) -> bool:
    return all(
        get_item(
            self.client,
            slug,
            item_id
        ).processing_status == "complete"
        for item_id in item_ids
    )
```

☝🏻 Check the above, it's untested, but it's a decent gist of the idea.

owencjones (approved):
I like it, with some suggestions - review comments before merge

darwin/future/meta/objects/stage.py
Nathanjp91: Small thing to consider here as well is that each time this gets called, it's making len(item_ids) requests, regardless of how many are still processing. 

I would propose a check that pops off from the list ones that are completed so they're not continually checked.

owencjones: Yeah, decent idea, could make the difference at scale.

darwin/future/meta/objects/stage.py
saurbhc: I agree - adding these as default arguments in the top level(s) of this function calls in a new commit.

darwin/future/meta/objects/stage.py
```diff
@@ -59,7 +61,52 @@ def item_ids(self) -> ItemIDQuery:
...
+            max_attempts (int, optional): Max number of attempts. Defaults to 5.
+            wait_time (float, optional): Wait time between attempts. Defaults to 0.5.
+        """
+        completed = []
+        for attempt in range(1, wait_max_attempts + 1):
+            # check if all items are complete
+            for item_id in item_ids:
+                if get_item(self.client, slug, item_id).processing_status != "complete":
+                    # if not complete, wait and try again with the in-complete items
+                    time.sleep(wait_time * attempt)
```
Nathanjp91, line 86: this time.sleep should be outside of the item_ids loop, we want it to wait a certain amount of time after checking the entire loop rather than per item. Even a small wait per item is going to add up when looking at 1000+ items and each check is independent of the others in the list anyway. Also can just do item_ids.remove(id) rather than the list comprehension every time. Either works though.

Nathanjp91 (changes requested):
Minor changes on where the sleep happens.

darwin/future/meta/objects/stage.py
saurbhc: Good catch! Fixed in a new commit.

>  can just do item_ids.remove(id) rather than the list comprehension every time

This is done as well - It is now iterating over a copy of `item_ids[:]` and removing the item from `item_ids`.

PR #709, opened 2023-11-07
[PY-386][external] Core method to move items between folders & tests

JBWilkie:
# Problem
Currently no way to move items between folders in `darwin.future`

# Solution
Define the core method to move a list of items by their UUIDs by using the path endpoint

# Changelog
Added core method to move items between folders

### DIFF
--- a/darwin/future/core/client.py
+++ b/darwin/future/core/client.py
@@ -10,7 +10,12 @@ from pydantic import BaseModel, root_validator, validator
 from requests.adapters import HTTPAdapter, Retry
 
 from darwin.future.core.types.common import JSONType, QueryString
-from darwin.future.exceptions import NotFound, Unauthorized, UnprocessibleEntity
+from darwin.future.exceptions import (
+    BadRequest,
+    NotFound,
+    Unauthorized,
+    UnprocessibleEntity,
+)
 
 
 class TeamsConfig(BaseModel):
@@ -235,6 +240,8 @@ def raise_for_darwin_exception(response: requests.Response) -> None:
     """
     if response.status_code == 200:
         return
+    if response.status_code == 400:
+        raise BadRequest(response)
     if response.status_code == 401:
         raise Unauthorized(response)
     if response.status_code == 404:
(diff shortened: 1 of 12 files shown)


darwin/future/core/client.py
```diff
@@ -235,6 +240,8 @@ def raise_for_darwin_exception(response: requests.Response) -> None:
     """
     if response.status_code == 200:
         return
+    if response.status_code == 400:
+        raise BadRequest(response)
```
JBWilkie, line 244: As covered in my comment in [PY-386 here](https://linear.app/v7labs/issue/PY-386/core-edit-items-move-to-folder#comment-4929616a), we probably want to catch the 400 errors this endpoint can throw in Core

To do this, I introduced the `BadRequest` class. This broke some existing tests that returned 400s, as previously they were expecting `HTTPError`, so updated each breaking test to expect BadRequest instead

darwin/future/tests/core/items/test_move_items_to_folder.py
```diff
@@ -0,0 +1,58 @@
...
+        },
+        path="/test/path",
+    )
+
+    # Check that the response mathces what we expect
+    assert response == {"affected_item_count": 2}
+
+
+def test_move_list_of_items_to_folder_with_error_response() -> None:
+    api_client = Mock(spec=ClientCore)
```
Nathanjp91, line 45: They both achieve the same thing but for consistency can you move this to a responses.RequestMock approach

Nathanjp91 (changes requested):

darwin/future/tests/core/items/test_move_items_to_folder.py
JBWilkie: Makes sense - There are a few other tests written this way, I'll update them as well

owencjones: @Nathanjp91 introduced the use of `responses` on this stuff, but it has become the paradigm overall, and it's a good example of socialised testing when used well.

owencjones (approved):

Nathanjp91 (approved):

PR #712, opened 2023-11-10
[PY-560][external] Changes to core API methods to improve consistency

JBWilkie:
# Problem
- 1: There were some inconsistencies with how our core API methods were written. They should be written to totally reflect the API endpoint interfaces. For example, a few methods were written to only accept a single `dataset_id`, when the endpoints can accept multiple `dataset_ids`.
- 2: The majority of our tests use the `responses.RequestsMock` syntax, but some were using a mock API client with `api_client = Mock(spec=ClientCore)`. It is better to have a consistent approach.

# Solution
- 1: Adjusted the methods to accept multiple dataset_ids
- 2: Adjusted tests so that all core tests use the former style for consistency

# Changelog
Changes to Core API methods to improve API endpoint parameter alignment & style consistency


darwin/future/core/items/delete_items.py
```diff
@@ -19,26 +19,28 @@ def delete_list_of_items(
...
-        The client to use for the request
+        The client to use for the request.
     team_slug: str
-        The slug of the team containing the items
-    dataset_id: int
-        The ID of the dataset containing the items
-    item_ids: List[UUID]
-        The IDs of the items to be deleted
+        The slug of the team containing the items.
+    dataset_ids: (int | List[int])
```
owencjones, line 25: Same comment as before, this looks like it suggests a tuple
```suggestion
    dataset_ids: int | List[int]
```

darwin/future/core/items/move_items.py
```diff
@@ -1,47 +1,56 @@
...
-    dataset_id: str
-        The id or slug of the dataset to move items for
-    stage_id: str
-        The id or slug of the stage to move items to
-    item_ids: List[UUID]
-        A list of item ids to move to the stage
+        The slug of the team to move items for.
+    workflow_id: UUID
+        The id of the workflow to move items for.
+    dataset_ids: (int | List[int])
```
owencjones, line 30:
```suggestion
    dataset_ids: int | List[int]
```

darwin/future/core/items/move_items_to_folder.py
```diff
@@ -20,29 +20,31 @@ def move_list_of_items_to_folder(
...
-        The client to use for the request
+        The client to use for the request.
     team_slug: str
-        The slug of the team containing the items
-    dataset_id: int
-        The ID of the dataset containing the items
-    item_ids: List[UUID]
-        The IDs of the items to be moved
+        The slug of the team containing the items.
+    dataset_ids: (int | List[int])
```
owencjones, line 26:
```suggestion
    dataset_ids: int | List[int]
```

darwin/future/core/items/restore_items.py
```diff
@@ -19,26 +19,29 @@ def restore_list_of_items(
...
-        The client to use for the request
+        The client to use for the request.
     team_slug: str
-        The slug of the team containing the items
-    dataset_id: int
-        The ID of the dataset containing the items
-    item_ids: List[UUID]
-        The IDs of the items to be restored
+        The slug of the team containing the items.
+    dataset_ids: (int | List[int])
```
owencjones, line 25:
```suggestion
    dataset_ids: int | List[int]
```

darwin/future/core/items/set_item_priority.py
```diff
@@ -20,30 +20,32 @@ def set_item_priority(
...
-        The client to use for the request
+        The client to use for the request.
     team_slug: str
-        The slug of the team to set the priority for
-    dataset_id: int
-        The dataset to set the priority for
-    item_ids: List[UUID]
-        The item ids to set the priority for
+        The slug of the team containing the items.
+    dataset_id: (int | List[int])
```
owencjones, line 26:
```suggestion
    dataset_id: int | List[int]
```

darwin/future/meta/objects/stage.py
```diff
@@ -74,8 +74,10 @@ def move_attached_files_to_stage(self, new_stage_id: UUID) -> Stage:
             self.meta_params["workflow_id"],
             self.meta_params["dataset_id"],
         )
-        ids = [x.id for x in self.item_ids.collect_all()]
-        move_items_to_stage(self.client, slug, w_id, d_id, new_stage_id, ids)
+        ids = [str(x.id) for x in self.item_ids.collect_all()]
+        move_items_to_stage(
+            self.client, slug, w_id, d_id, new_stage_id, {"item_ids": ids}
```
owencjones, line 79: I know these aren't yours, but non-descriptive variable names could be renamed.

owencjones (changes requested):
Looks good - I only requested doc block changes, and they're all in as suggestions so you could do them in GUI and ping me for approval

darwin/future/meta/objects/stage.py
JBWilkie: Good catch, have amended

owencjones (approved):

PR #715, opened 2023-11-13
[PY-407][PY-534][PY-404] Item + ItemQuery objects

Nathanjp91:
# Problem
Item + ItemQuery doesn't exist, need to be added in so can manage and download these items.  Also add in associated methods and features to objects that have Items as well as methods that items should have

# Solution
Added in these objects
- Item Meta now exists along with ItemQuery
-  delete() method on both item and itemquery that requests the item/s be deleted on the server
- objects that interact with items have an items property that maps to a ItemQuery
- Small changes to QueryString for easier typing

# Changelog
Items now queryable with paginated support


darwin/future/core/types/common.py
```diff
@@ -73,18 +80,31 @@ class QueryString:
...
     def __str__(self) -> str:
-        return "?" + "&".join(f"{k}={v}" for k, v in self.value.items())
+        output: str = "?" if self.value else ""
+        for k, v in self.value.items():
+            if isinstance(v, list):
+                for x in v:
+                    output += f"{k}={x}&"
+            else:
+                output += f"{k}={v}&"
+        return output[:-1]  # remove trailing &
```
Nathanjp91, line 108: This had to be changed and added to facilitate that a key can map to multiple values, eg dataset_id=0000&dataset_id=0001

darwin/future/data_objects/item.py
```diff
@@ -75,7 +75,7 @@ def validate_fps(cls, values: dict) -> dict:
         elif isinstance(value, (int, float)):
             type = values.get("type")
             if type == "image":
-                assert value == 0, "fps must be 0 for images"
+                assert value == 0 or value == 1.0, "fps must be '0' or '1.0' for images"
```
Nathanjp91, line 78: turns out on our api, 1.0 also exists for items that are images. Found in dogfooding.

darwin/future/core/items/delete_items.py
```diff
@@ -10,7 +11,8 @@
 def delete_list_of_items(
     client: ClientCore,
     team_slug: str,
-    dataset_ids: int | List[int],
+    dataset_ids: int | list[int] | Literal["all"],
+    item_ids: List[UUID],
```
Nathanjp91, line 15: changed how delete_items works, item_ids is required so it doesn't make sense to have that as part of the optional filters rather than a required component. Also added in Literal "all" which should delete everything in a dataset_id. I think this still works with item_ids being an empty list but unsure on this corner case as I only just thought of it and haven't tested.

darwin/future/core/types/common.py
```diff
@@ -1,13 +1,20 @@
...
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, List, Mapping, Protocol, Union
 
 from darwin.future.data_objects import validators as darwin_validators
-from darwin.future.data_objects.typing import UnknownType
 
 JSONType = Union[Dict[str, Any], List[Dict[str, Any]]]  # type: ignore
 
 
+class Implements_str(Protocol):
```
Nathanjp91, line 10: so this protocol should mean it'll work with anything that can be converted to a string, so things like ints, floats etc will work.

darwin/future/core/types/common.py
```diff
@@ -73,18 +80,31 @@ class QueryString:
...
+        mapped: dict[str, list[str] | str] = {}
+        for k, v in value.items():
+            if isinstance(v, list):
+                mapped[k] = [str(x) for x in v]
+            else:
+                mapped[k] = str(v)
+        return mapped
 
-    def __init__(self, value: Dict[str, str]) -> None:
+    def __init__(self, value: Mapping[str, list[Stringable] | Stringable]) -> None:
```
Nathanjp91, line 96: Mapping here is a more free-form Dict. As per the mypy docs, it's better to use the constructs like Mapping/Sequence as input parameters, and more concrete ones like dict/list as return types.

darwin/future/core/types/common.py
```diff
@@ -1,13 +1,20 @@
...
 
 JSONType = Union[Dict[str, Any], List[Dict[str, Any]]]  # type: ignore
 
 
+class Implements_str(Protocol):
+    def __str__(self) -> str:
+        ...
+
+
+Stringable = Union[str, Implements_str]
```
Nathanjp91, line 15: I'm not entirely sure why this union has to exist, but I couldn't get the type system to properly work without it. One would expect implements_str to encompass str natively, but it doesn't.

darwin/future/core/items/delete_items.py
JBWilkie: I've been able to use the [delete endpoint](https://docs.v7labs.com/reference/items-delete) successfully while not passing in any `item_ids`

In PY-560, I removed item_ids from the core API methods because it wasn't required. Am I missing something?

Nathanjp91: Hmm, I swear in my testing the other day when adding in delete functionality that I got error responses from the endpoint when item_ids weren't included. I think they might have been 200 returns, but the body had an error code in it.

JBWilkie: If you only provide dataset_ids, you'll get something like:

```
{
  "errors": {
    "code": "MISSING_SELECT_ALL_FILTER",
    "detail": null,
    "message": "Filter object has to contain at least one non-required filter or 'select_all: true'.",
    "status": 422
  }
}
```

But all you need to do is add any other filter, it can be item_ids or anything else

Nathanjp91: ah ok, I'll make that change then. I must have just inferred that an ID had to be present

darwin/future/tests/core/types/test_querystring.py
```diff
@@ -13,12 +11,23 @@ def test_querystring_happy_path() -> None:
     assert str(query_string_2) == "?foo=bar&baz=qux"
 
     query_string_3 = QueryString({})
-    assert str(query_string_3) == "?"
+    assert str(query_string_3) == ""
```
owencjones, line 14: Yeah, protocols do work much better for this eh.

owencjones (approved):
Some comments for consideration, but overall LGTM

darwin/future/core/items/delete_items.py
```diff
@@ -32,14 +34,12 @@ def delete_list_of_items(
     JSONType
         The response data.
     """
-    assert (
-        filters
-    ), "No parameters provided, please provide at least one non-dataset id filter"
     payload = {
         "filters": {
-            "dataset_ids": dataset_ids
+            "dataset_ids": [str(item) for item in dataset_ids]
```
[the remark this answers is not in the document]
Nathanjp91: all our endpoints do, although I have just removed that because I do think it's dangerous. Also it makes typing a bit of a nightmare anyway.

darwin/future/data_objects/item.py
```diff
@@ -15,7 +15,7 @@ def validate_no_slashes(v: UnknownType) -> str:
     assert isinstance(v, str), "Must be a string"
     assert len(v) > 0, "cannot be empty"
     assert "/" not in v, "cannot contain slashes"
-    assert " " not in v, "cannot contain spaces"
+    # assert " " not in v, "cannot contain spaces"
```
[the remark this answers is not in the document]
Nathanjp91: it breaks on an item pulled from server. I commented it out to get it working but forgot about it. It should be removed though

PR #716, opened 2023-11-13
[PY-559][internal] E2E test: darwin convert CLI function with YOLOv8

owencjones:
# Problem
`darwin convert` function was not tested for any format

# Solution
A parameterised E2E test was introduced that tests the functionality for `darwin convert` for the `yolo_segmented` (YoloV8) version.

This test performs a byte-for-byte comparison of output against a stored expectation.

This will be expanded in future tickets to allow for testing of other formats.

# Changelog
Introduced new E2E test for `darwin convert`, and a directory comparison helper.


e2e_tests/cli/convert/test_convert.py
```diff
@@ -0,0 +1,72 @@
...
+
+                with Path(expected_path / file.name).open() as f:
+                    expected_content = f.read()
+
+                if content != expected_content:
+                    return (False, f"File {file} does not match expected file")
+
+        return (True, "")
+
+    @pytest.mark.parametrize(
```
Nathanjp91, line 49: I'm a bit confused with this paramterization, is the intention that we'd extend this to other formats? In which case should the class be something more like TestConvertersCLI? or is the intention that the parameters would indicate different cases of yolov8?

owencjones: Yeah, that is the intention - see ticket for the full overview - but good call on the class name, I have just overlooked that rename.  Will action.

Nathanjp91 (approved):
Good approach, very modular.

PR #718, opened 2023-11-14
[PY-401][external] Restore Meta Item & ItemQuery functions

JBWilkie:
# Problem
Currently no Meta methods for restoring items

# Solution

# Changelog
Meta methods to restore Items & ItemQuery objects


darwin/future/meta/objects/item.py
```diff
@@ -53,6 +54,18 @@ def delete(self) -> None:
...
 
+    def restore(self) -> None:
+        team_slug, dataset_id = (
+            self.meta_params["team_slug"],
+            self.meta_params["dataset_id"]
+            if "dataset_id" in self.meta_params
+            else self.meta_params["dataset_ids"],
+        )
+        assert isinstance(team_slug, str)
+        dataset_id = cast(Union[int, List[int]], dataset_id)
```
owencjones, line 65: I'm a fan of your use of `cast` in these tickets John, assuming we're confident of the typing in `meta_params`

darwin/future/tests/meta/objects/test_itemmeta.py
```diff
@@ -44,3 +44,26 @@ def test_delete(item: Item) -> None:
...
+                        "filters": {
+                            "item_ids": [str(item.id)],
+                            "dataset_ids": [dataset_id],
+                        }
+                    }
+                )
+            ],
+            json={},
+        )
+        item.restore()
```
owencjones, line 69: As with the other one, I would sad path these, but in these, they are small enough, that it isn't a massive deal.  Red/Green/Refactor is a good approach.

darwin/future/tests/meta/queries/test_item.py
```diff
@@ -70,3 +70,38 @@ def test_delete(
...
+                        "filters": {
+                            "item_ids": [str(item.id) for item in items],
+                            "dataset_ids": [dataset_id],
+                        }
+                    }
+                )
+            ],
+            json={},
+        )
+        item_query.restore()
```
owencjones, line 107: And with this one ideally, but again, small ticket, not the end of the world.

The thing I would aim to be testing is each path through the function

owencjones (approved):
Comments for consideration as before, but ok with you making your own changes afterwards, and merging.

darwin/future/tests/meta/objects/test_itemmeta.py
JBWilkie: Quick Q on this - Since neither `item.restore()` or `item_query.restore() `take any parameters nor do they return anything, what could we consider the sad path?

JBWilkie: @owencjones I discussed these tests with Nathan - Since they don't return anything or take any arguments, there should only be one path these functions can actually take

Error responses from the API are covered in the underlying Core method tests. Given this, I'll proceed to merge but let me know if you disagree or have any other thoughts

darwin/future/tests/meta/queries/test_item.py
JBWilkie: @owencjones Similar to my comment on the restore meta ticket and after discussion w/ Nathan, there is only one path this function can take due to no argument & no return

Will proceed with this test alone as error responses are tested by the underlying Core method, but let me know if you disagree

PR #719, opened 2023-11-14
[PY-402][external] Archive Meta Item & ItemQuery functions

JBWilkie:
# Problem
Currently no Meta methods for archiving items

# Solution

# Changelog
Meta methods to archive Items & ItemQuery objects


darwin/future/tests/meta/objects/test_itemmeta.py
```diff
@@ -44,3 +44,26 @@ def test_delete(item: Item) -> None:
...
+                        "filters": {
+                            "item_ids": [str(item.id)],
+                            "dataset_ids": [dataset_id],
+                        }
+                    }
+                )
+            ],
+            json={},
+        )
+        item.archive()
```
owencjones, line 69: Test the assert in the `archive` function works too.

darwin/future/tests/meta/queries/test_item.py
```diff
@@ -70,3 +70,38 @@ def test_delete(
...
+                        "filters": {
+                            "item_ids": [str(item.id) for item in items],
+                            "dataset_ids": [dataset_id],
+                        }
+                    }
+                )
+            ],
+            json={},
+        )
+        item_query.archive()
```
owencjones, line 107: Sad path opp here too.

owencjones (approved):
Some more sad path opportunities in these tests, but the code is sound and getting really consistent.  Good stuff.  I'll approve so you can make changes at discretion.

darwin/future/tests/meta/queries/test_item.py
JBWilkie: @owencjones Similar to my comment on the restore meta ticket and after discussion w/ Nathan, there is only one path this function can take due to no argument & no return

Will proceed with this test alone as error responses are tested by the underlying Core method, but let me know if you disagree

PR #720, opened 2023-11-14
[PY-403][external] Set Priority Meta Item & ItemQuery functions

JBWilkie:
# Problem
Currently no Meta methods for setting item priority

# Solution

# Changelog
Meta methods to set priority for Items & ItemQuery objects


owencjones (approved):
Yeah, good stuff

PR #721, opened 2023-11-14
[Ai-1477]-darwin-v2-export-fix

ChristofferEdlund:
# Problem
Recent changes to darwin introduced local datasets loading of only `darwin_v2` format.

When converting video datasets to images by using the split function `split_video_annotations`
https://github.com/v7labs/darwin-py/blob/79fd606063dff08c1e22d08c452b75f663cc25a2/darwin/dataset/remote_dataset.py#L143

there are usage of a darwin export will not generate v2 annotations, instead it generates this json (assumably v1)
https://github.com/v7labs/darwin-py/blob/79fd606063dff08c1e22d08c452b75f663cc25a2/darwin/exporter/formats/darwin.py#L17
```
    return {
        "annotations": annotations,
        "image": {
            "filename": annotation_file.filename,
            "height": annotation_file.image_height,
            "width": annotation_file.image_width,
            "url": annotation_file.image_url,
        },
    }

```

Leading to our dataloaders not being able to load any data that has been converted from video now that `v2` format is enforced.

# Solution
Updating the darwin v2 export function `build_image_annotation` to generate valid `v2` image annotations.
It's only used in that one place above currently so changing it should fix our problem and not introduce any unwanted issues.

# Changelog
- Updated `split_video_annotations` convertion so it generates valid per-image darwin `v2` annotations.
- Added option to select if one wants to load empty annotations in LocalDataset with `keep_empty_annotations` arg (default is False)


almazan (approved):

darwin/utils/utils.py
```diff
@@ -1005,6 +1005,10 @@ def split_video_annotation(annotation: dt.AnnotationFile) -> List[dt.AnnotationF
             for a in annotation.annotations
             if isinstance(a, dt.VideoAnnotation) and i in a.frames
         ]
+
+        if len(annotations) < 1:
```
whilefalse, line 1009: Are we sure this isn't going to break anything for people using darwin-py? This is a breaking change in behaviour for anyone calling this function themselves, right?

whilefalse (approved):

darwin/utils/utils.py
ChristofferEdlund: This is a breaking change indeed, this will only generate annotations for frames that has them. Removing this we need to change the logic in other places to (perhaps local_dataset) to only load and parse the annotation files that has annotations in them. 

In practice, I think that this is only used in our own training infrastructure, and only for videos. But, happy to discuss other ways to address this instead.

darwin/torch/dataset.py
```diff
@@ -634,12 +639,17 @@ def get_target(self, index: int) -> Dict[str, Tensor]:
...
+        try:
+            stacked_targets = {
+                "boxes": torch.stack([v["bbox"] for v in targets]),
+                "area": torch.stack([v["area"] for v in targets]),
+                "labels": torch.stack([v["label"] for v in targets]),
+                "image_id": torch.tensor([index]),
+            }
+        except Exception as e:
+            print(target)
+            raise e
```
whilefalse, line 652: Is it intended to leave this in?

whilefalse (approved):

darwin/datatypes.py
```diff
@@ -565,9 +566,18 @@ def make_polygon(
...
     """
+
+    if darwin_v1:
+        polygon_data = {"path": point_path}
+    else:
+        # Lets handle darwin V2 datasets
+        if not isinstance(point_path[0], list):
+            point_path = [point_path]
+        polygon_data = {"paths": point_path}
+
```
ChristofferEdlund, line 577: Do we still want to keep support for darwin_v1 ?

Nathanjp91: so v1 is now deprecated, as I understand it's just some latent stuff in AI that require it. If you're moving away from v1 here, then I'm happy to ignore v1 support.

Nathanjp91: so v1 is now deprecated, as I understand it's just some latent stuff in AI that require it. If you're moving away from v1 here, then I'm happy to ignore v1 support.

ChristofferEdlund: This PR is part of us moving away from darwin v1, so we won't need it anymore. But tests might need to be re-written, in particular these:

- https://github.com/v7labs/darwin-py/blob/36ca721102507048319c8ffcf9f9373137653460/darwin/exporter/formats/coco.py#L576

- https://github.com/v7labs/darwin-py/blob/36ca721102507048319c8ffcf9f9373137653460/darwin/utils/utils.py#L749

- https://github.com/v7labs/darwin-py/blob/36ca721102507048319c8ffcf9f9373137653460/tests/darwin/datatypes_test.py#L10

- https://github.com/v7labs/darwin-py/blob/36ca721102507048319c8ffcf9f9373137653460/tests/darwin/datatypes_test.py#L21

ChristofferEdlund: According to slack conv, we will keep the option for darwin-v1 format for now. Team darwin-py can remove it later.

https://vseven.slack.com/archives/C016XJEPU78/p1700578652840489

darwin/exporter/formats/darwin.py
```diff
@@ -16,65 +16,144 @@
...
+        "h": data.get("h"),
+        "w": data.get("w"),
+        "x": data.get("x"),
+        "y": data.get("y"),
+    }
+
+
+def _build_polygon_data(
+    data: Dict[str, Any]
+) -> Dict[str, List[List[Dict[str, float]]]]:
```
Nathanjp91, line 76: This looks kind of nebulous and not immediately understandable. Can we create a datatype here instead
`Path = List[List[Dict[str, float]]]` or something similar.

darwin/torch/dataset.py
```diff
@@ -328,6 +328,11 @@ def get_target(self, index: int) -> Dict[str, Any]:
         for annotation in target["annotations"]:
             annotation_type: str = annotation.annotation_class.annotation_type
             path_key = "paths" if annotation_type == "complex_polygon" else "path"
+
+            # Darwin V2 only has paths (TODO it might be more robust fixes)
```
Nathanjp91, line 332: is this still TODO?

darwin/importer/importer.py
```diff
@@ -697,22 +723,22 @@ def _handle_annotators(
 def _get_annotation_data(
     annotation: dt.AnnotationLike, annotation_class_id: str, attributes: dt.DictFreeForm
 ) -> dt.DictFreeForm:
+    # Todo Convert to import Darwin format
```
Nathanjp91, line 726: Is this still a TODO?

Nathanjp91 (changes requested):
Some stuff to look at.

darwin/exporter/formats/darwin.py
Nathanjp91: if it didn't break a bunch of other typing I'd even suggest more structure here with namedtuples, like
```python
def Point(NamedTuple):
    x: float
    y: float
    
Path = List[Point]
Paths = List[Path]
```

We're exploring stronger typing for darwin future so something like that'll probably make it in.

darwin/datatypes.py
```diff
@@ -565,9 +566,18 @@ def make_polygon(
     Annotation
         A polygon ``Annotation``.
     """
+
+    if darwin_v1:
```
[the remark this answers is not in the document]
ChristofferEdlund: @Nathanjp91 , we did not. That is why things crashed when trying to generate json v2 image annotations.

darwin/importer/importer.py
ChristofferEdlund: Not anymore, since I did that now (hopefully)

darwin/torch/dataset.py
ChristofferEdlund: I suppose it might be more robust ways of doign this, but happy to keep it this way too.

darwin/exporter/formats/darwin.py
ChristofferEdlund: I am in favor of this, you want to add it into this PR or should it be a ticket for darwin-py after this is merged?

darwin/dataset/remote_dataset_v2.py
```diff
@@ -543,3 +544,8 @@ def _fetch_stages(self, stage_type):
             workflow_id,
             [stage for stage in workflow["stages"] if stage["type"] == stage_type],
         )
+
+    def _build_image_annotation(
```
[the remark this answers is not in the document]
ChristofferEdlund: `build_image_annotation` is either imported from export/formats/ darwin.py or darwin_1_0.py depending if it is a v1 or v2 remote dataset.

I made this class to handle each case dynamically. 
`darwin.py` `build_image_annotation` (v2 format)
https://github.com/v7labs/darwin-py/blob/31a831412f60b7d2d96da3da55c8ed612d6b54e0/darwin/exporter/formats/darwin.py#L17

`darwin_1_0.py` `build_image_annotation` (v1 format)
https://github.com/v7labs/darwin-py/blob/31a831412f60b7d2d96da3da55c8ed612d6b54e0/darwin/exporter/formats/darwin_1_0.py#L218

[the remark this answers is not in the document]
ChristofferEdlund: Any suggestions on how to solve this in another way is welcome. :+1:

darwin/exporter/formats/darwin.py
Nathanjp91: For now can you add this to the PR just as 
```python
darwin/datatypes.py
Path = List[Dict[str, float]]
Paths = List[Path]

def _build_polygon_data(...) -> Dict[str, Paths]
```

I'll do the stronger typing stuff with darwin future code, but the changes above will be backwards compatible with existing typing, it just explicitly names them for easier understanding.

darwin/dataset/remote_dataset_v2.py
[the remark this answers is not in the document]
Nathanjp91: oh right I see, the function name is the same but the import points to a different format. That makes more sense then.

darwin/datatypes.py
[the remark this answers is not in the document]
Nathanjp91: but is this make_polygon function not shared with images? How did they work but videos didn't?

darwin/exporter/formats/darwin.py
ChristofferEdlund: @Nathanjp91 , how about PolygonPath and PolygonPaths, since "Path" collides with the commonly used from pathlib import Path

darwin/datatypes.py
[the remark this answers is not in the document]
ChristofferEdlund: Was it ever used for images? I think using it for images made them into darwin_v1 format. The annotations are made on the platform, not sure what they used. For the `convert` cli functionality I do not think there is an option to convert to darwin-v2 ?

darwin/datatypes.py
```diff
@@ -565,9 +560,18 @@ def make_polygon(
...
+    else:
+        # Lets handle darwin V2 datasets
+        if not isinstance(point_path[0], list):
+            point_path = [point_path]
+        polygon_data = {"paths": point_path}
+
     return Annotation(
         AnnotationClass(class_name, "polygon"),
-        _maybe_add_bounding_box_data({"path": point_path}, bounding_box),
+        _maybe_add_bounding_box_data(polygon_data, bounding_box),
```
rslota, line 574: This doesn't look correct to me. You seem to be putting v1 vs v2 differences into abstract model that doesn't care about versions. 
Can we use the `datatypes.make_polygon` or `datatypes.make_complex_polygon` instead? I'm 99% sure that putting `paths` (plural) on a class of type `polygon` is invalid in our codebase. 

For reference you can take a look at the logic that reads JSON files directly here: https://github.com/v7labs/darwin-py/blob/79fd606063dff08c1e22d08c452b75f663cc25a2/darwin/utils/utils.py#L647
As you can see this code uses both `polygon` and `complex_polygon` classes for internal structures and this is how it should stay.

ChristofferEdlund: Thank you @rslota , it's I that misunderstood the  internal darwin format and assume that it should adhere to the darwin-v1 and v2 formats. Will make changes to the PR that keeps the internal format intact, thanks!

darwin/datatypes.py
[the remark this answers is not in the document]
ChristofferEdlund: Removed this now based on Rafaels comment.

Nathanjp91 (approved):

PR #722, opened 2023-11-15
[A!-1424] darwin py black conv

ChristofferEdlund:
# Problem
Reformatting to black linewidth 88

# Solution
Reformatting to black linewidth 88

# Changelog
- Reformatting to black linewidth 88


Nathanjp91 (approved):
Sensible PR and brings entire codebase in line.

PR #723, opened 2023-11-15
automatic ruff --fix changes

Nathanjp91:
# Problem
Existing issues in older code can be automatically applied via ruff tool

# Solution
run `ruff --fix` on existing code, applying changes

# Changelog
ruff based automatic changes applied to existing code

### DIFF
--- a/darwin/dataset/remote_dataset_v1.py
+++ b/darwin/dataset/remote_dataset_v1.py
@@ -345,8 +345,10 @@ class RemoteDatasetV1(RemoteDataset):
         items : Iterator[DatasetItem]
             The ``DatasetItem``\\s to be completed.
         """
+
         def wf_template_id_mapper(item):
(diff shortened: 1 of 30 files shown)


darwin/dataset/remote_dataset_v1.py
```diff
@@ -346,9 +345,10 @@ def complete(self, items: Iterator[DatasetItem]) -> None:
         items : Iterator[DatasetItem]
             The ``DatasetItem``\\s to be completed.
         """
-        wf_template_id_mapper = lambda item: item.current_workflow[
-            "workflow_template_id"
-        ]
+
+        def wf_template_id_mapper(item):
```
Nathanjp91, line 349: interesting here that ruff hates lambdas. I don't disagree, the syntax sucks in python

owencjones (approved):

PR #724, opened 2023-11-16
[PY-605][external]Added support for folder structures when importing CSV video tags

JBWilkie:
# Problem
Currently, CSV video tag imports assume items are in a flat folder structure, meaning the importer ignores files in folder structures.

# Solution
Added logic to pull out and consider the remote path when importing

# Changelog
Added support for folder structures when importing CSV video tags

### DIFF
--- a/darwin/importer/formats/csv_tags_video.py
+++ b/darwin/importer/formats/csv_tags_video.py
@@ -54,7 +54,6 @@ def parse_path(path: Path) -> Optional[List[dt.AnnotationFile]]:
         annotation_classes = set(
             [annotation.annotation_class for annotation in annotations]
         )
-
         filename_path = Path(filename)
         remote_path = str(filename_path.parent)
         if not remote_path.startswith("/"):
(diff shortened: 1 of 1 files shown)


owencjones (approved):
Yep, seems fine to me.  I don't know if this is heavily used, but useful change nonetheless.

Bit of a worry that it didn't break a test, so we could ticket introducing one it would break.

JBWilkie:
> Yep, seems fine to me. I don't know if this is heavily used, but useful change nonetheless.
> 
> Bit of a worry that it didn't break a test, so we could ticket introducing one it would break.

Good idea, I've ticketed the test in PY-606

darwin/importer/formats/csv_tags_video.py
```diff
@@ -51,15 +51,22 @@ def parse_path(path: Path) -> Optional[List[dt.AnnotationFile]]:
             file_annotation_map[filename].append(annotation)
     for filename in file_annotation_map:
         annotations = file_annotation_map[filename]
-        annotation_classes = {annotation.annotation_class for annotation in annotations}
+        annotation_classes = set(
+            [annotation.annotation_class for annotation in annotations]
+        )
```
saurbhc, line 56: This could be simplified a bit by directly using set-comprehension:
```suggestion
        annotation_classes = set(
            annotation.annotation_class for annotation in annotations
        )
```

NPatel1980:
QA passed, pair tested with John, imports to files within folders are now working.

PR #725, opened 2023-11-21
[MED-22][external] import nifti to raster layer mask annotations

tomvars:
# Problem
We previously supported nifti imports as polygons, this PR introduces support for nifti imports into the raster layer.
# Solution
Users now have a new mode available at upload. In addition to `video` and `instances` which both write polygons into Darwin, users now have `mask` mode which writes to the raster layer and creates `mask` annotations.

# Changelog
We've introduced the ability to import nifti labels directly into the raster layer! This means users can perform inference off-platform or import existing annotations into a project and then annotate them using our Masks feature.


darwin/importer/formats/nifti.py
```diff
@@ -177,8 +196,96 @@ def get_video_annotation(
         raise Exception("If is_mpr is True, slot_names must be of length 3")
 
 
-def nifti_to_video_annotation(
-    volume, class_name, class_idxs, slot_names, view_idx=2, pixdims=(1, 1, 1)
+def get_mask_video_annotations(volume, processed_class_map, slot_names):
```
Nathanjp91, line 199: Can you type annotate this? Inputs and outputs.

darwin/importer/formats/nifti.py
```diff
@@ -177,8 +196,96 @@ def get_video_annotation(
...
+            interpolated=False,
+            slot_names=slot_names,
+        )
+        mask_video_annotation.id = mask_annotation_ids[class_name]
+        mask_video_annotations.append(mask_video_annotation)
+
+    return [raster_video_annotation] + mask_video_annotations
+
+
+def nifti_to_video_polygon_annotation(
```
Nathanjp91, line 282: Same as above.

darwin/importer/formats/nifti.py
```diff
@@ -376,3 +483,16 @@ def process_nifti(
         data_array = nib.orientations.apply_orientation(img.get_fdata(), ornt)
         pixdims = img.header.get_zooms()
         return data_array, pixdims
+
+
+def convert_to_dense_rle(raster):
```
Nathanjp91, line 488: as above.

Nathanjp91 (changes requested):
Would also love to see some tests added if possible.

tomvars:
> Would also love to see some tests added if possible.

Added a test

tests/darwin/importer/formats/import_nifti_test.py
```diff
@@ -1,3 +1,4 @@
+import ast
```
Nathanjp91, line 1: I'm guessing this a stray import from vscode? if so remove it
```suggestion
```

Nathanjp91 (approved):

PR #731, opened 2023-11-22
[PY-627] Long Video support for split video annotations

Nathanjp91:
# Problem
Long videos currently does not get split correctly for formats like coco that require treating each image seperately. 

# Solution
Added in logic for generation of split video annotations if frame_urls is not present, relying instead on frame_count, which is a field to be added in conjunction with changes to slots

# Changelog
split video annotations should now work correctly with long video files

### DIFF
--- a/darwin/utils/utils.py
+++ b/darwin/utils/utils.py
@@ -999,7 +999,7 @@ def split_video_annotation(annotation: dt.AnnotationFile) -> List[dt.AnnotationF
     if not annotation.frame_count and not annotation.frame_urls:
         raise AttributeError("This Annotation has no frames")
     
-    urls = annotation.frame_urls if annotation.frame_urls else [''] * (annotation.frame_count or 1)
+    urls = annotation.frame_urls if annotation.frame_urls else [None] * (annotation.frame_count or 1)
     frame_annotations = []
     for i, frame_url in enumerate(urls):
         annotations = [
--- a/darwin/utils/utils.py
+++ b/darwin/utils/utils.py
@@ -573,7 +573,7 @@ def _parse_darwin_v2(path: Path, data: Dict[str, Any]) -> dt.AnnotationFile:
             .get("name", None),
             annotation_classes=annotation_classes,
             annotations=annotations,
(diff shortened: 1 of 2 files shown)


darwin/utils/utils.py
```diff
@@ -993,13 +994,17 @@ def split_video_annotation(annotation: dt.AnnotationFile) -> List[dt.AnnotationF
...
-        raise AttributeError("This Annotation has no frame urls")
-
+        raise AttributeError("This is not a video annotation")
+
+    if not annotation.frame_count and not annotation.frame_urls:
+        raise AttributeError("This Annotation has no frames")
+    urls = (
+        annotation.frame_urls
+        if annotation.frame_urls
+        else [None] * (annotation.frame_count or 1)
```
Nathanjp91, line 1004: the `or 1` here is just to satisfy type check, the prior assert should already catch that either frame_count or frame_urls should exist but for some reason mypy is not smart enough to pick that up.

darwin/utils/utils.py
```diff
@@ -993,13 +994,17 @@ def split_video_annotation(annotation: dt.AnnotationFile) -> List[dt.AnnotationF
...
+    )
     frame_annotations = []
-    for i, frame_url in enumerate(annotation.frame_urls):
+    for i, frame_url in enumerate(urls):
```
Nathanjp91, line 1007: In the case of a long video frame_url get's passed into the AnnotationFile.image_url field as None. Perhaps this is not the correct assumption but I couldn't reason about what an image_url would be in this case, and none of our tests fail.

darwin/utils/utils.py
rslota: Can we add a comment somewhere here on why we're chaining this? Long Videos are pretty new, so this code may be a bit confusing without context.

saurbhc, line 1004: Just a bit pythonic this way with ternary operator:
```suggestion
    urls = annotation.frame_urls or [None] * (annotation.frame_count or 1)
```

saurbhc (approved):

rslota (approved):

PR #732, opened 2023-11-23
Fix `RecursionError` in `Item` class

saurbhc:
# Problem
Printing the `Item` class raises RecursionError

```python
from darwin.future.meta.client import Client

dataset = client.team.datasets.collect()[-1]
print(dataset.items.where().collect())
```
```console
  File "t4.py", line 15, in <module>
    raise SystemExit(main())
  File "t4.py", line 8, in main
    print(dataset.items.where().collect())
  File "/Users/saurabhchopra/github/v7labs/darwin-py/darwin/future/meta/objects/base.py", line 50, in __repr__
    return str(self)
  File "/Users/saurabhchopra/github/v7labs/darwin-py/darwin/future/meta/objects/base.py", line 50, in __repr__
    return str(self)
  File "/Users/saurabhchopra/github/v7labs/darwin-py/darwin/future/meta/objects/base.py", line 50, in __repr__
    return str(self)
  [Previous line repeated 329 more times]
RecursionError: maximum recursion depth exceeded while getting the str of an object
```

# Solution
Implement the missing `__str__` method in `Item` class, which solves the issue.

# Changelog
Fix `RecursionError` while printing an `Item` class.

### DIFF
--- a/darwin/future/meta/objects/base.py
+++ b/darwin/future/meta/objects/base.py
@@ -48,3 +48,6 @@ class MetaBase(Generic[R]):
 
     def __repr__(self) -> str:
         return str(self)
+
+    def __str__(self) -> str:
+        return f"{self.__class__.__name__}({self._element})"
--- a/darwin/future/meta/objects/item.py
+++ b/darwin/future/meta/objects/item.py
@@ -6,8 +6,8 @@ from uuid import UUID
 from darwin.future.core.items.archive_items import archive_list_of_items
 from darwin.future.core.items.delete_items import delete_list_of_items
 from darwin.future.core.items.move_items_to_folder import move_list_of_items_to_folder
-from darwin.future.core.items.set_item_priority import set_item_priority
 from darwin.future.core.items.restore_items import restore_list_of_items
+from darwin.future.core.items.set_item_priority import set_item_priority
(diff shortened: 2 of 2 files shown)


Nathanjp91 (approved):
Interested to understand why that causes a recursion error, and if it's general to all inherited meta objects, seeking to fix that (or enforce __str__ as an abstractmethod), but good to approve this small fix.

saurbhc:
> Interested to understand why that causes a recursion error

From what I could see, it may be coming from the `Base` class -> `__repr__` method [[link](https://github.com/v7labs/darwin-py/blob/master/darwin/future/meta/objects/base.py#L49-L50)] which was calling the `str` implementation - as there was no `__str__` defined in `Item` class, it re-tried in a loop calling itself in recursion?

Nathanjp91:
> > Interested to understand why that causes a recursion error
> 
> From what I could see, it may be coming from the `Base` class -> `__repr__` method [[link](https://github.com/v7labs/darwin-py/blob/master/darwin/future/meta/objects/base.py#L49-L50)] which was calling the `str` implementation - as there was no `__str__` defined in `Item` class, it re-tried in a loop calling itself in recursion?

Can you add in a default implementation to `__str__` of that `Base` class for this PR, just print the underlying element it manages

PR #734, opened 2023-11-24
[PY-558] add e2e-test for `darwin convert yolo ...`

saurbhc:
# Problem
No e2e-test for `darwin convert yolo ...`

# Solution
Add the parameterised `YOLO` e2e-test, which can be tested by:
```console
$ pytest e2e_tests/cli/convert/test_convert.py::TestExportCli::test_darwin_convert[yolo-input_path1-expectation_path1]
```

# Changelog
Add `darwin convert` e2e-test for `yolo` format


e2e_tests/data/yolo/to/darknet.labels
```diff
@@ -0,0 +1,2 @@
+bbox
```
Nathanjp91, line 1: does yolo really only have a list of classes as output? if so can we just slim down the example json? It would seem not to need anything complex.

Nathanjp91 (approved):

PR #740, opened 2023-12-08
[PY-573] Darwin Future Properties base objects

Nathanjp91:
# Problem
No support exists for new Properties features loading into annotations

# Solution
Write pydantic objects to handle the upcoming properties features. Includes seperate objects to parse the metadata.json files + some base DarwinV2 annotations that these properties will be listed in. 

# Changelog
Basic Support for Property fields added


saurbhc (approved):
Looks good to me!

PR #741, opened 2023-12-08
[PY-611] Silently ignore "empty" masks during import

saurbhc:
# Problem
Silently ignore "empty" masks (i.e. masks that do not have a corresponding raster layer) during import

# Solution
During import, checking if the mask annotation is present in `raster_layer` - then further checking the same in the whole `dense_rle` flat array, if not present, don't import the mask (i.e., remove the mask annotation from import request data and also remove it from raster_layer mask-mapping)

# Changelog
Silently ignore "empty" masks (i.e. masks that do not have a corresponding raster layer) during import

### DIFF
--- a/darwin/importer/importer.py
+++ b/darwin/importer/importer.py
@@ -773,6 +773,30 @@ def _import_annotations(
 
         data = _get_annotation_data(annotation, annotation_class_id, attributes)
 
+        # check if the mask is empty (i.e. masks that do not have a corresponding raster layer) if so, skip import of the mask
+        if annotation_type == "mask":
+            rl = next((a for a in annotations if a.annotation_class.annotation_type == "raster_layer"), None)
+
+            # check if the 'annotation_class_id' is in raster_layer's mask_annotation_ids_mapping dict
+            if rl is not None:
+                mask_annotation_ids_mapping = rl.data.get("mask_annotation_ids_mapping")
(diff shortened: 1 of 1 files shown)


darwin/importer/importer.py
```diff
@@ -773,6 +773,30 @@ def _import_annotations(
...
+
+            # check if the 'annotation_class_id' is in raster_layer's mask_annotation_ids_mapping dict
+            if rl is not None:
+                mask_annotation_ids_mapping = rl.data.get("mask_annotation_ids_mapping")
+                dense_rle = rl.data.get("dense_rle")
+                _annotation_id = annotation.id
+                _annotation_mask_reference_i = mask_annotation_ids_mapping[_annotation_id]
+
+                if dense_rle is not None:
+                    for i in range(0, len(dense_rle), 2):
```
rslota, line 788: So this loop may take few seconds, in the end this in extreme case is as long as many pixels we have on the image. Meanwhile, for each mask we're doing this search, so quite high time complexity. Maybe I'm a bit too paranoid, but wdyt about building first some kind of hash lookup table for those IDs instead of looping though same `dense_rle` so many times?

Nathanjp91 (approved):
I think this is something we should test. Not here, but e2e should include an import raster mask with/without empty layers. Can you add tickets to linear under Py Grooming and Backlog for this.

PR #742, opened 2023-12-08
[PY-597] Darwin-py extra authenticate options

Nathanjp91:
# Problem
Darwin Authentication currently has no way to programmatically set authenticate options without engaging the CLI

# Solution
Set extra options, allowing to pass in api-key, default-team and dataset directory as CLI flags, as well as setting env variables
.env
```
DARWIN_API_KEY=' '
DARWIN_TEAM=' '
DARWIN_DATASETS_DIR=' '
```
or via CLI commands
`darwin authenticate --api_key {key} --default_team {team} --datasets_dir {dir}`

# Changelog
darwin-py can now be authenticated via env variables and CLI flags without requiring input


darwin/cli.py
```diff
@@ -63,7 +64,9 @@ def _run(args: Namespace, parser: ArgumentParser) -> None:
 
     # Authenticate user
     if args.command == "authenticate":
-        api_key = os.getenv("DARWIN_API_KEY")
+        api_key = os.getenv("DARWIN_API_KEY") or args.api_key
+        default_team = os.getenv("DARWIN_TEAM") or args.default_team
+        datasets_dir = os.getenv("DARWIN_DATASETS_DIR") or args.datasets_dir
         if api_key:
             print("Using API key from DARWIN_API_KEY")
```
JBWilkie, line 71: When I pass an API key with `--api_key`, I see `Using API key from DARWIN_API_KEY`

Is this technically incorrect, since it's actually from a CLI arg?

JBWilkie (approved):
Tested all options, all working on my end, LGTM

Just one small point on message output correctness

darwin/cli.py
JBWilkie: Same goes for the `--default_team` and `--dataset_dir` flags

Nathanjp91: updated for that

PR #744, opened 2023-12-12
[PY-618] Add properties to existing Annotation Objects

saurbhc:
# Problem
No support exists for new Properties features loading into annotation objects

# Solution
Add properties to existing Annotation Objects like `Annotation`, `VideoAnnotation`

TODO:
* [x] Merge #740 to master before merging this PR (as this is branched from the 740 ^).

# Changelog
Add properties to existing Annotation Objects


darwin/utils/utils.py
```diff
@@ -941,6 +945,21 @@ def _parse_annotators(annotators: List[Dict[str, Any]]) -> List[dt.AnnotationAut
     return [dt.AnnotationAuthor(annotator["full_name"], annotator["email"]) for annotator in annotators]
 
 
+def _parse_properties(properties: List[Dict[str, Any]]) -> Optional[List[dt.SelectedProperty]]:
```
Nathanjp91, line 948: does dt.SelectedProperty exist? I don't see it in datatypes so shouldn't this point to darwin.future...SelectedProperty?

Nathanjp91 (approved):
Mostly good. I think maybe a small typing issue.

darwin/utils/utils.py
saurbhc: You're right, the reason my testing worked because darwin.datatypes was already importing SelectedProperty from future..
Fixed in a new commit!

PR #745, opened 2023-12-12
[PY-632][external] Fix for LocalDatasets ignoring dataset partitions

JBWilkie:
# Problem
A recent PR introduced a bug where `LocalDataset` is ignoring the split and partition information and it's always loading all images of the dataset when passing [train, val, test] params.

# Solution
Re-introduce partitions to LocalDataset construction

# Changelog
Changes to support LocalDataset partitions & releases pulled in folders simultaneously


Nathanjp91 (approved):
Happy to approve after discussion on call.

PR #746, opened 2023-12-12
[PY-645][externa] Improved tolerance for dots in filenames & test linting

JBWilkie:
# Problem
If the filename of either:
- Any source file in any slot, or:
- The filename of the dataset item itself contains dots "." that aren't file extensions, the NifTI exporter will break

# Solution
Introduce logic that explicitly looks for `.dcm`, `.nii`, and `.nii.gz` extensions instead of pulling out suffixes

# Changelog
- Improved tolerance for dots in medical filenames
- Some test linting


darwin/exporter/formats/nifti.py
```diff
@@ -137,50 +137,28 @@ def check_for_error_and_return_imageid(
...
 
     filename = Path(video_annotation.filename)
-    try:
-        suffixes = filename.suffixes[-2:]
-    except IndexError:
-        suffixes = filename.suffixes
-    if len(suffixes) == 2:
-        image_id = str(filename).rstrip("".join(suffixes))
-    elif len(suffixes) == 1:
-        image_id = str(filename.stem)
```
JBWilkie, line 183: I'm a little unsure about this. Currently, if `.nii.gz` then we include this in the `image_id`

Otherwise, if `.dcm` or `.nii`, we don't include this in the `image_id`. Why is this?

rslota: > Currently, if .nii.gz then we include this in the image_id

Is that actually true? `image_id = str(filename).rstrip("".join(suffixes))` would strip out `.nii.gz` 🤔

darwin/exporter/formats/nifti.py
```diff
@@ -137,50 +137,28 @@ def check_for_error_and_return_imageid(
...
-                    image_id = filename.stem
-                else:
-                    return create_error_message_json(
-                        "Misconfigured filename, not ending in .nii or .dcm. Are you sure this is medical data?",
-                        output_dir,
-                        str(filename),
-                    )
-            else:
+            if not (
+                filename.name.endswith(".nii.gz")
```
rslota, line 145: Can we also make sure we check that case-insensitive? That's gonna be next reported bug.

darwin/exporter/formats/nifti.py
JBWilkie: Actually yes that is correct

Just tested and the behaviour now is:
- If any slot has any ending other than `.nii`, `.dcm`, or `.nii.gz`, the exporter fails (we want this)
- The filename itself can be anything, and if it ends with any of the 3 above we strip it away

I believe this is correct

darwin/exporter/formats/nifti.py
JBWilkie: Updated, all 6 checks are now case-insensitive

darwin/exporter/formats/nifti.py
```diff
@@ -137,50 +137,28 @@ def check_for_error_and_return_imageid(
...
-    except IndexError:
-        suffixes = filename.suffixes
-    if len(suffixes) == 2:
-        image_id = str(filename).rstrip("".join(suffixes))
-    elif len(suffixes) == 1:
-        image_id = str(filename.stem)
+    if filename.name.lower().endswith(".nii.gz"):
+        image_id = str(filename).rstrip(".nii.gz")
+    elif filename.name.lower().endswith(".nii"):
+        image_id = str(filename).rstrip(".nii")
```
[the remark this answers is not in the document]
JBWilkie: God I need to multitask less...

Updated. Now we check the lowered version of the filename so we're guaranteed to pick up any of the 3 matches

darwin/exporter/formats/nifti.py
```diff
@@ -137,50 +137,28 @@ def check_for_error_and_return_imageid(
...
-    try:
-        suffixes = filename.suffixes[-2:]
-    except IndexError:
-        suffixes = filename.suffixes
-    if len(suffixes) == 2:
-        image_id = str(filename).rstrip("".join(suffixes))
-    elif len(suffixes) == 1:
-        image_id = str(filename.stem)
+    if filename.name.lower().endswith(".nii.gz"):
+        image_id = str(filename).lower().rstrip(".nii.gz")
```
rslota, line 157: Are you sure `image_id` can be a lower case of the filename? I'm not sure how it's used, I would personally prefer to just strip the extension in case-insensitive manner (I guess you may need regex for that with `re.IGNORECASE`). But if you're sure `image_id` can be modified this way, that's fine by me :)

JBWilkie: I don't believe it's an issue, but I cannot be not certain. Therefore, I've updated this so that we only strip the extension in a case-insensitive manner and leave the rest of the `image_id` untouched

rslota (approved):

PR #747, opened 2023-12-12
[py-578][py-580] get properties endpoints

Nathanjp91:
# Problem
Lacking support for properties endpoints

# Solution
adds in functions for requesting properties objects
`def get_property_by_id(...)`
`def get_team_properties(...)`
`def get_team_full_properties(...)` -> wrapper around `get_team_properties` that includes options by default

# Changelog
properties endpoints supported in darwin.future.core


darwin/future/core/properties/get.py
```diff
@@ -0,0 +1,72 @@
...
+
+def get_team_full_properties(
+    client: ClientCore,
+    team_slug: Optional[str] = None,
+    params: Optional[QueryString] = None,
+) -> List[FullProperty]:
+    params = (
+        params + QueryString({"include_values": True})
+        if params and not params.get("include_values")
+        else QueryString({"include_values": True})
```
saurbhc, line 45: to avoid `darwin.future.exceptions.UnprocessibleEntity: UnprocessibleEntity: <Response [422]>` error when we call:
```python
client = Client.local()
properties = get_team_full_properties(client)
```
Updating `True` -> `"true"`

```suggestion
        params + QueryString({"include_values": "true"})
        if params and not params.get("include_values")
        else QueryString({"include_values": "true"})
```

saurbhc (changes requested):
QueryParams change, looks like the API can't process `True` as query param, but works with `"true"`.

darwin/future/core/properties/get.py
Nathanjp91: good catch. Fixed with a change to querystring though.

saurbhc (approved):
Looks good! Thank you 💯

PR #749, opened 2023-12-13
[PY-623] Get team properties backported to client

saurbhc:
# Problem
Will need to backport some api calls to client class, getting team properties which will be defined in future code.

# Solution
- add `client.get_team_properties` method
- backport future client

TODO:
* [x] Merge #747 to master before merging this PR.

# Changelog
Get team properties backported to client


darwin/config.py
```diff
@@ -263,3 +263,27 @@ def get_all_teams(self) -> List[Team]:
                 teams_data.append(the_team_data)
 
         return teams_data
+
+    def create_future_config(self) -> Dict[str, Any]:
```
Nathanjp91, line 267: I think I would prefer a futureConfig.from_old(old_config) approach than a oldConfig.to_future(...).

darwin/client.py
```diff
@@ -1474,3 +1482,19 @@ def api_v2(self) -> BackendV2:
         if not team:
             raise ValueError("No team was found.")
         return BackendV2(self, team.slug)
+
+    def get_team_properties(self, team_slug: Optional[str] = None, full: bool = False) -> List[FullProperty]:
```
Nathanjp91, line 1486: Not big, but I think the bool here should be default True, I think in most instances people will want the options and certainly for our use of it, it'll be required. Also I would probably rename it so that it's more clear what it does, like include_options.

Nathanjp91 (changes requested):
Would like to see a test (in future tests module) re config from old as well as tests in the older code in the style of responses. We're introducing api calls for properties that clients will be using so need to make reasonably sure that they don't break if we change things in darwin.future now.

Nathanjp91 (approved):
Looks good: I renamed include_options -> include_property_values because of discussion on project_properties about removing 'options' from the objects in favour of property_values instead. So just to keep consistency

PR #750, opened 2023-12-13
[MED-23][external] Support for export from Raster Layer to NifTI

tomvars:
# Problem

Users of Darwin can currently annotate as masks but not export these results to NifTI

# Solution

We create a new code path within the same NifTI export logic. This means that for darwin JSON that contains polygons we will still parse these as before and output a NifTI volume. If the user has K mask classes they will receive K output files called `SeriesInstanceUID_classname_m.nii.gz` (NOTE: for nifti uploads SeriesInstanceUID is the filename)

# Changelog
Introduced initial support for NifTI export from Masks

### DIFF
--- a/darwin/exporter/formats/nifti.py
+++ b/darwin/exporter/formats/nifti.py
@@ -33,6 +33,7 @@ class Volume:
     pixdims: List
     class_name: str
     series_instance_uid: str
+    from_raster_layer: bool
 
 
 def export(annotation_files: Iterable[dt.AnnotationFile], output_dir: Path) -> None:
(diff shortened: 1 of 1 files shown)


darwin/exporter/formats/nifti.py
```diff
@@ -228,95 +274,152 @@ def populate_output_volumes(
...
+                polygons = shift_polygon_coords(
+                    frames[frame_idx].data["path"],
+                    volume[annotation.annotation_class.name].pixdims,
+                )
+            else:
+                continue
+            frames[frame_idx].annotation_class.name
+            im_mask = convert_polygons_to_mask(polygons, height=height, width=width)
+            volume = output_volumes[series_instance_uid]
+            if view_idx == 0:
```
Nathanjp91, line 340: is there a reason you use '0' here but XYPLANE above? Are these totally different or is the magic number here meant to be XYPLANE? Same for below

darwin/exporter/formats/nifti.py
```diff
@@ -228,95 +274,152 @@ def populate_output_volumes(
...
+        slot_name = annotation.slot_names[0]
+        slot = slot_map[slot_name]
+        series_instance_uid = slot.metadata.get(
+            "SeriesInstanceUID", "SeriesIntanceUIDNotProvided"
+        )
+        volume = output_volumes.get(series_instance_uid)
+        frames = annotation.frames
+
+        # define the different planes
+        XYPLANE = 0
```
Nathanjp91, line 298: Are these intrinsinc to nifti and potentially used by other functions? If so can we make this a type

Something like
```python
class Plane(Enum):
    XY = 0
    XZ = 1
    YZ = 2

if view_idx == Plane.XY.value
```

darwin/exporter/formats/nifti.py
```diff
@@ -228,95 +274,152 @@ def populate_output_volumes(
...
+                    frames[frame_idx].data["path"],
+                    volume[annotation.annotation_class.name].pixdims,
+                )
+            else:
+                continue
+            frames[frame_idx].annotation_class.name
+            im_mask = convert_polygons_to_mask(polygons, height=height, width=width)
+            volume = output_volumes[series_instance_uid]
+            if view_idx == 0:
+                volume[annotation.annotation_class.name].pixel_array[
```
Nathanjp91, line 341: It would be great to generalize this function, if only so that it's named and has a docstring. I can't actually tell by inspection what it's meant to be doing too easily

```python

def transform_plane(plane, volume, ...): #or some other more relevant name
    """ docstrings explaining high level so that numpy logic is easier to reason about """
    if plane == XYZ:
        transform_indexes = [np.s_[:], frame_idx, np.s[:]]
    ...
    volume[annotation.annotation_class.name].pixel_array[transform_indexes] = np.logical_or(...)
```

Nathanjp91: and I know this is probably some nifti format specific thing but jumping into the code for these specific data formats from the outside is a lot easier if these sorts of transforms are easier to reason about, with names and descriptions

Nathanjp91 (changes requested):

darwin/exporter/formats/nifti.py
shernshiou: Implemented in the new commit

darwin/exporter/formats/nifti.py
shernshiou: Implemented in the new commit

JBWilkie (approved):
LGTM - Tested this commit in a deploybot to ensure we can export multiple mask & polygon classes as NifTI volumes

PR #756, opened 2023-12-16
[PY-576][PY-581] Create + Update of properties

Nathanjp91:
# Problem
No Support exists for creation or updating of properties

# Solution
Added functions for supporting creation and updating
```python
def update_property(...) -> FullProperty:
    ...
def update_property_value(...) -> PropertyValue:
    ...
def create_property(...) -> FullProperty:
    ...
```
Also added in e2e tests to support these new features in preparation for support of imports

To Do
- [x] unit tests for new validators, pydantic functions and endpoint core functions

# Changelog
Creation and updating of properties now possible in darwin.future

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,7 +4,7 @@
     "editor.formatOnSave": true,
     "editor.tabSize": 4,
     "editor.codeActionsOnSave": {
-      "source.organizeImports": true,
+      "source.organizeImports": "explicit"
     },
     "editor.defaultFormatter": "ms-python.black-formatter"
   },
--- a/darwin/future/data_objects/properties.py
+++ b/darwin/future/data_objects/properties.py
@@ -3,11 +3,20 @@ from __future__ import annotations
 import json
 import os
 from pathlib import Path
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Literal, Optional, Union
 
 from pydantic import validator
 
 from darwin.future.data_objects.pydantic_base import DefaultDarwin
+
+PropertyType = Literal[
+    "multi_select",
+    "single_select",
+    "text",
+    "attributes",
+    "instance_id",
+    "directional_vector",
+]
 
 
 class PropertyOption(DefaultDarwin):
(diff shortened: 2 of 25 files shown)


e2e_tests/data/coco/to/output.json
```diff
@@ -85,13 +85,13 @@
   ],
```
Nathanjp91, line 85: e2e tests broke again because of the ordering. Will put in a ticket for a fix on the coco exporter to always sort before saving.

darwin/future/data_objects/properties.py
```diff
@@ -49,14 +71,37 @@ class FullProperty(DefaultDarwin):
...
     annotation_class_id: Optional[int]
-    property_values: Optional[List[PropertyOption]]
-    options: Optional[List[PropertyOption]]
+    property_values: Optional[List[PropertyValue]]
+    options: Optional[List[PropertyValue]]
+
+    def to_create_endpoint(
+        self,
+    ) -> dict:
+        if not getattr(self, "annotation_class_id") or self.annotation_class_id is None:
```
Nathanjp91, line 86: rest of the fields are non-optional so they'll already be captured by ValidationErrors

e2e_tests/data/coco/to/output.json
Nathanjp91: and it reverted itself and had to be redone.

.vscode/settings.json
```diff
@@ -4,7 +4,7 @@
     "editor.formatOnSave": true,
     "editor.tabSize": 4,
     "editor.codeActionsOnSave": {
-      "source.organizeImports": true,
+      "source.organizeImports": "explicit"
```
Nathanjp91, line 7: vscode keeps saving the file like this even if I revert it. So I assume it's vscode deprecating or changing a naming convention

saurbhc: +1
I can confirm that my vscode updated the same change!

saurbhc (approved):
Great Work!! The E2E tests seem very helpful. :)

PR #759, opened 2023-12-19
[PY-622][PY-626][PY-665] Import Properties (create/update) while importing annotations

saurbhc:
# Problem
- [PY-622] Import Annotations is missing new Properties import feature (create and update properties)
- [PY-626] Updates to import endpoint - annotation_properties
- [PY-665] Update property-value object structure to support un-nested value field

# Solution
Implement `_import_properties` before `_import_annotations` & add those properties in import annotations API payload.

# Changelog
Implement Import Properties (create/update) while importing annotations


owencjones (changes requested):
Intricate work, good stuff.  Some concern that there are not extra tests for the extra functionality.  Reach out to me if there's a reason for that, or if you need backup, and we can sort it together.

darwin/importer/importer.py
```diff
@@ -267,6 +274,159 @@ def _resolve_annotation_classes(
...
+
+    create_properties: List[FullProperty] = []
+    update_properties: List[FullProperty] = []
+    for annotation in annotations:
+        annotation_name = annotation.annotation_class.name
+        annotation_type = annotation_type = (
+            annotation.annotation_class.annotation_internal_type
+            or annotation.annotation_class.annotation_type
+        )
+        annotation_class_id = int(annotation_class_ids_map[annotation_name])
```
Nathanjp91, line 317: I think this method might have an issue with conflicting annotation_names, which can happen. It's possible to have both a polygon and a line annotation be the same name. Only (annotation_name, annotation_type/class) is unique.

darwin/importer/importer.py
```diff
@@ -267,6 +274,159 @@ def _resolve_annotation_classes(
...
+            # check if property value is missing for a property that requires a value
+            if t_prop.required and not a_prop.value:
+                _msg = f"Annotation: '{annotation_name}' -> Property '{a_prop.name}' requires a value!"
+                raise ValueError(_msg)
+
+            # check if property value/type is different in t_prop (team) options
+            for t_prop_val in t_prop.property_values or []:
+                if t_prop_val.value == {"value": a_prop.value} and t_prop_val.type == a_prop.type:
+                    break
+            else:
```
Nathanjp91, line 394: I actually had to look up the semantics of for-else, I so rarely see them in python.

darwin/importer/importer.py
```diff
@@ -267,6 +274,159 @@ def _resolve_annotation_classes(
     return local_classes_not_in_dataset, local_classes_not_in_team
 
 
+def _import_properties(
+    metadata_path: Union[Path, bool],
+    client: "Client",
+    annotations: List[dt.Annotation],
+    annotation_class_ids_map: Dict[str, str],
+):
```
Nathanjp91, line 282: can you docblock this, detail it mostly for internal as it's not really client facing code. But it would be good to have a semblance of the intendid flow, of what conditions cause it to raise and exit early. Ie when properties don't exist, mismatched etc.

Nathanjp91 (changes requested):
Overall good code, but I think there's one issue that might come into play and maybe requires changing.

darwin/future/data_objects/properties.py
```diff
@@ -157,5 +157,5 @@ class SelectedProperty(DefaultDarwin):
 
     frame_index: int
     name: str
-    type: str
+    type: Literal["string"] = "string"
```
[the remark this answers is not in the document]
Nathanjp91: yeah, they had intentions of maybe supporting more types later, but as it stands everything is a string and then typecast on the backend.

darwin/datatypes.py
```diff
@@ -475,10 +475,10 @@ def split_paths_by_metadata(
     tuple[Path, Optional[list[PropertyClass]]]
         A tuple containing the path to the metadata file and the list of property classes.
     """
-    if not is_properties_enabled(path, dir, filename):
+    metadata_path = is_properties_enabled(path, dir, filename)
+    if isinstance(metadata_path, bool):
```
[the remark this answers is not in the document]
saurbhc: `is_properties_enabled` returns `

darwin/datatypes.py
```diff
@@ -405,13 +405,13 @@ class Property:
     name: str
 
     # Type of the property
-    type: str
+    type: PropertyType
```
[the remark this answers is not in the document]
saurbhc: Do you mean something like this:
```
    type: """PropertyType"""
```

darwin/importer/importer.py
```diff
@@ -267,6 +274,159 @@ def _resolve_annotation_classes(
     return local_classes_not_in_dataset, local_classes_not_in_team
 
 
+def _import_properties(
+    metadata_path: Union[Path, bool],
```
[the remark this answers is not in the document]
saurbhc: I do agree on the increased complexity of this function - I have also branched out something similar to this [PR](https://github.com/v7labs/darwin-py/pull/760/files).

Should I merge that ^ PR to this one and then do another attempt on creating sub-functions and reduce complexity?

darwin/importer/importer.py
saurbhc: Good catch! Updated the variable to have `(annotation_class.name, annotation_type)` as keys now!

darwin/importer/importer.py
saurbhc: They are quite helpful when we need to know if a for loop had an break statement or not - an alternative to this would be using a extra variable to hold the state :)

> The else clause executes after the loop completes normally. This means that the loop did not encounter a break statement

darwin/importer/importer.py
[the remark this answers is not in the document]
saurbhc: I believe we have some tickets on E2E test on properties import - so testing this function properly should be covered in that.

For the function names, what would you prefer? some examples:
- `_import_properties` -> `_create_update_properties`? - as it Creates/Updates missing/mismatched properties from annotation & metadata.json file to team-properties.
- `_update_payload_with_properties` -> `_update_annotation_payload_with_properties`? - as it updates the annotation import API payload data with annotation-properties data

What are your thoughts on it?

darwin/importer/importer.py
```diff
@@ -267,6 +275,260 @@ def _resolve_annotation_classes(
...
+
+                        property_values = full_property.property_values
+                        # find property value in m_prop (.v7/metadata.json) options
+                        for m_prop_option in m_prop_options:
+                            if m_prop_option.get("value") == a_prop.value:
+                                # update property_values with new value
+                                full_property.property_values.append(
+                                    PropertyValue(
+                                        position=m_prop_option.get("position"),  # type: ignore
+                                        type=m_prop_option.get("type"),  # type: ignore
```
AndriiKlymchuk, line 408: We don't have this field for property value in the manifest anymore. Please make fresh export and check. We don't need to pass it in the API either.

darwin/importer/importer.py
```diff
@@ -267,6 +275,260 @@ def _resolve_annotation_classes(
...
+                                        value={"value": m_prop_option.get("value")},  # type: ignore
```
AndriiKlymchuk, line 409: Value must not be wrapped in the `{"value": ...}` object anymore.

darwin/importer/importer.py
```diff
@@ -267,6 +275,260 @@ def _resolve_annotation_classes(
...
+                                    color=m_prop_option.get("color"),  # type: ignore
+                                )
+                            )
+                            break
+                    # if it doesn't exist, create it
+                    full_property = FullProperty(
+                        name=a_prop.name,
+                        type=m_prop_type,  # type from .v7/metadata.json
+                        required=m_prop.required,  # required from .v7/metadata.json
+                        description="property-created-during-annotation-import",
```
AndriiKlymchuk, line 434: Don't we export description in the manifest? As far as I remember we do, and must take it from there instead of puting custom string

AndriiKlymchuk (changes requested):
Left couple of comments

darwin/importer/importer.py
saurbhc: Ah yes, there's another PR which i'll merge now - https://github.com/v7labs/darwin-py/pull/769

darwin/importer/importer.py
saurbhc: description field is added in a new commit, and removed the type field too, also made sure the code dosen't break

AndriiKlymchuk (approved):

PR #761, opened 2024-01-02
add semantic_mask e2e test - data and cli

saurbhc:
# Problem
(describe the problem here)

# Solution

# Changelog
(Will appear in release docs)


darwin/exporter/formats/mask.py
```diff
@@ -512,7 +512,7 @@ def export(
             image = Image.fromarray(mask)
         image.save(outfile)
 
-    with open(output_dir / "class_mapping.csv", "w") as f:
+    with open(output_dir / "class_mapping.csv", "w", newline="") as f:
```
saurbhc, line 515: In windows, files generated has an extra line in b/w every row, which fails the e2e test as it fails to compare the output and expected files. Found the fix from these links:
- https://stackoverflow.com/a/3191811
- https://stackoverflow.com/a/3348664

owencjones: I think that's a difference between line endings for Windows and Linux/Posix being interpreted different by the writer (`\r\n` vs `\n`), but couldn't swear to it.

Either way, you've solved it.  Nice one.

owencjones (approved):
Yep, straightforward approach, like it.

PR #762, opened 2024-01-04
[PY-664][PY-662] BUG: Rasters import for videos fails on darwin-py 0.8.54

saurbhc:
# Problem
A customer hit an error with importing rasters 'VideoAnnotation' object has no attribute 'data' that I was able to reproduce on darwin-py 0.8.54 but imported successfully on 0.

# Solution
Handle Raster Layer checks for `VideoAnnotation` object.

# Changelog
Support Raster Layer checks for `VideoAnnotation` object.


darwin/importer/importer.py
```diff
@@ -729,6 +729,58 @@ def _get_overwrite_value(append: bool) -> str:
     return "false" if append else "true"
 
 
+def _parse_empty_masks(
+        annotation: dt.Annotation,
+        rl: dt.Annotation,
+        rl_dense_rle_ids: Optional[Set[str]] = None,
+        rl_dense_rle_ids_frames: Optional[Dict[int, Set[str]]] = None,
+    ):
+    if rl_dense_rle_ids_frames is None and isinstance(annotation, dt.VideoAnnotation):
```
owencjones, line 738: The structure of this is sound, and largely speaking I wouldn't change much, but some overall comments:
* It would be good to break the complexity down into smaller functions for the various sub clauses it performs
* Whilst I actually don't mind, PEP8 and `pythonism` would have us prefer `elif` over multiple `if` to ensure multiple blocks don't accidentally get run - where applicable obviously.
* Prefer verbose names over short ones - `rl` just as well be `raster_layer` for readability etc.

darwin/importer/importer.py
```diff
@@ -729,6 +729,58 @@ def _get_overwrite_value(append: bool) -> str:
...
+            del rl.data["mask_annotation_ids_mapping"][_annotation_id]
+            return rl_dense_rle_ids, rl_dense_rle_ids_frames
+
+    if rl_dense_rle_ids_frames is not None and isinstance(annotation, dt.VideoAnnotation):
+        # check every frame - if the 'annotation_class_id' is in raster_layer's mask_annotation_ids_mapping dict
+        assert isinstance(rl, dt.VideoAnnotation)
+        for frame_index, _annotation in annotation.frames.items():
+            _annotation_id = _annotation.id
+            if (
+                frame_index in rl_dense_rle_ids_frames and
```
owencjones, line 771: I'd be tempted to break this down into a meaningfully named boolean

owencjones (changes requested):
Some comments to consider.

darwin/importer/importer.py
saurbhc: I have updated the code to only have: if with elif. which simplifies the code. Variable names are more verbose too - so anyone can understand what's happening.

darwin/importer/importer.py
```diff
@@ -729,6 +729,58 @@ def _get_overwrite_value(append: bool) -> str:
...
+        for frame_index, _rl in rl.frames.items():
+            rl_dense_rle_ids_frames[frame_index] = set(_rl.data["dense_rle"][::2])
+
+    if rl_dense_rle_ids is None and isinstance(annotation, dt.Annotation):
+        # build a set of dense_rle_ids (for each frame in Annotation object)
+        assert isinstance(rl, dt.Annotation)
+        rl_dense_rle_ids = set(rl.data["dense_rle"][::2])
+
+    if rl_dense_rle_ids is not None and isinstance(annotation, dt.Annotation):
+        # check if the 'annotation_class_id' is in raster_layer's mask_annotation_ids_mapping dict
```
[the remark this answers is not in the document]
saurbhc: there's now only 1 if-elif branch in this function - for VideoAnnotation and Annotation objects.

darwin/importer/importer.py
saurbhc: Um, as there's only 1 unique if check like this now - does this still apply? happy to create a variable and write `if variable` block there instead something like:
```
variable = (if statement)
if variable:
    ...
```
What do you think?

darwin/importer/importer.py
```diff
@@ -729,6 +729,58 @@ def _get_overwrite_value(append: bool) -> str:
...
+        _annotation_id = annotation.id
+        if (
+            rl.data["mask_annotation_ids_mapping"][_annotation_id]
+            not in rl_dense_rle_ids
+        ):
+            # skip import of the mask, and remove it from mask_annotation_ids_mapping
+            logger.warning(
+                f"Skipping import of mask annotation '{annotation.annotation_class.name}' as it does not have a corresponding raster layer"
+            )
+            del rl.data["mask_annotation_ids_mapping"][_annotation_id]
```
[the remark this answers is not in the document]
saurbhc: The idea here is to check if there are any empty masks (which dosen't exist in dense_rle block of flat array) and if there is, then remove that from annotation mapping from the raster_layer.data["mask_annotation_ids_mapping"] object.

rslota (approved):

PR #763, opened 2024-01-04
[PY-646][external] Resolved precision bug

JBWilkie:
# Problem
As described in PY-646, sufficiently small polygons can result in an assertion error due to rounding 4 decimal place floats to integers. This can cause the minimum & maximum x or y coordinates of the enclosing bounding box to be equal, resulting in a "flat" bbox with no area that is therefore invalid

# Solution
Pass the `rounding` arg as `False` to prevent this

# Changelog
Resolved rounding error for very small polygons in Pytorch dataloaders

### DIFF
--- a/darwin/client.py
+++ b/darwin/client.py
@@ -1503,7 +1503,9 @@ class Client:
             team_slug=team_slug or self.default_team,
         )
 
-    def create_property(self, team_slug: Optional[str], params: Union[FullProperty, JSONDict]) -> FullProperty:
+    def create_property(
+        self, team_slug: Optional[str], params: Union[FullProperty, JSONDict]
+    ) -> FullProperty:
         darwin_config = DarwinConfig.from_old(self.config)
         future_client = ClientCore(darwin_config)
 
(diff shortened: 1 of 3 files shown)


darwin/client.py
```diff
@@ -1513,12 +1515,14 @@ def create_property(self, team_slug: Optional[str], params: Union[FullProperty,
...
         )
 
-    def update_property(self, team_slug: Optional[str], params: Union[FullProperty, JSONDict]) -> FullProperty:
+    def update_property(
+        self, team_slug: Optional[str], params: Union[FullProperty, JSONDict]
+    ) -> FullProperty:
         darwin_config = DarwinConfig.from_old(self.config)
         future_client = ClientCore(darwin_config)
 
         return update_property_future(
```
owencjones, line 1524: Slight naming issue here, which is probably ok to leave, but a Future is a thing in itself, similar to a Javascript Promise, so worth bearing in mind in erm... _future_

owencjones (approved):
All good.

PR #764, opened 2024-01-06
[PY-544][external] Item query & Item ID query sorting functionality + some linting

JBWilkie:
# Problem
Currently no way to implement server-side API request sorting

# Solution
Implemented server-side sorting logic got ItemQuery & ItemIDQuery objects as these are currently the only two objects that work with sorting

# Changelog
Implemented server-side sorting logic for items

### DIFF
--- a/darwin/future/meta/queries/item.py
+++ b/darwin/future/meta/queries/item.py
@@ -14,8 +14,9 @@ from darwin.future.core.items.set_stage_to_items import set_stage_to_items
 from darwin.future.core.items.tag_items import tag_items
 from darwin.future.core.items.untag_items import untag_items
 from darwin.future.core.types.common import QueryString
-from darwin.future.core.types.query import PaginatedQuery
+from darwin.future.core.types.query import PaginatedQuery, QueryFilter
 from darwin.future.data_objects.item import ItemLayout
+from darwin.future.data_objects.sorting import SortingMethods
 from darwin.future.data_objects.workflow import WFStageCore
 from darwin.future.exceptions import BadRequest
 from darwin.future.meta.objects.item import Item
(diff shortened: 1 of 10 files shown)


darwin/future/meta/queries/item.py
```diff
@@ -60,6 +61,14 @@ def _collect(self) -> Dict[int, Item]:
         }
         return items
 
+    def sort(self, **kwargs: str) -> None:
+        sorting_methods = SortingMethods(**kwargs)
```
JBWilkie, line 65: `(**kwargs)` here throws a type warning I've had trouble getting to the bottom of

```
Argument of type "str" cannot be assigned to parameter "accuracy" of type "Literal['asc', 'desc'] | None" in function "__init__"
  Type "str" cannot be assigned to type "Literal['asc', 'desc'] | None"
    Type cannot be assigned to type "None"
    "str" cannot be assigned to type "Literal['asc']"
    "str" cannot be assigned to type "Literal['desc']"Pylance[reportGeneralTypeIssues](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportGeneralTypeIssues)
```

darwin/future/meta/queries/item.py
```diff
@@ -60,6 +61,14 @@ def _collect(self) -> Dict[int, Item]:
...
+        for key, value in sorting_methods.dict().items():
```
JBWilkie, line 66: I don't think using `.dict()` here is completely correct

darwin/future/meta/queries/item_id.py
```diff
@@ -39,3 +40,11 @@ def _collect(self) -> Dict[int, V7ID]:
             for i, uuid in enumerate(uuids)
         }
         return results
+
+    def sort(self, **kwargs: str) -> "ItemIDQuery":
+        sorting_methods = SortingMethods(**kwargs)
+        for key, value in sorting_methods.dict().items():
```
JBWilkie, line 46: Same issueswith `(**kwargs)` and `dict()` as above

owencjones (approved):
Looks good to me, you'll need to address that typing issue, but I'm happy with a type ignore in the scenario.

darwin/future/meta/queries/item.py
JBWilkie: Aha, that makes sense, cheers

PR #765, opened 2024-01-09
Add assign batch action SDK

rslota:
This PR adds "assign" batch action. It's largely copy-pasted "set_state" as it works the same way, just with users, not stages. It doesn't support "User" object as input, as I'm pretty sure we don't have one yet.
Example test:

```
from darwin.future.meta.client import Client

client = Client.local()
workflow = client.team.workflows.where(name="property_testing").collect_one()

workflow.items.assign(assignee_id=190)

```

### DIFF
--- a/darwin/future/core/items/assign_items.py
+++ b/darwin/future/core/items/assign_items.py
@@ -1,0 +1,44 @@
+from __future__ import annotations
+
+from darwin.future.core.client import ClientCore
+from darwin.future.core.types.common import JSONType
+from darwin.future.data_objects.typing import UnknownType
+
+
+def assign_items(
+    client: ClientCore,
+    team_slug: str,
+    dataset_ids: int | list[int],
+    assignee_id: int,
+    workflow_id: str,
+    filters: dict[str, UnknownType],
+) -> JSONType:
+    """
+    Assign a user to all items matched by filters.
+
+    Args:
+        client (ClientCore): The Darwin Core client.
+        team_slug (str): The team slug.
+        dataset_ids (int | list[int]): The dataset ids.
+        assignee_id (int): The user id to assign.
+        workflow_id (str): The workflow id that selected items have to belong to.
+        filters Dict[str, UnknownType]: The parameters of the filter.
+
+    Returns:
+        JSONType: The response data.
+    """
+    assert (
+        filters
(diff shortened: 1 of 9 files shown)


darwin/future/core/items/assign_items.py
```diff
@@ -0,0 +1,44 @@
...
+    team_slug: str,
+    dataset_ids: int | list[int],
+    assignee_id: int,
+    workflow_id: str,
+    filters: dict[str, UnknownType],
+) -> JSONType:
+    """
+    Assign a user to all items matched by filters.
+
+    Args:
```
owencjones, line 18: Check this parses with the doc generation tool.

darwin/future/core/items/assign_items.py
```diff
@@ -0,0 +1,44 @@
...
+from darwin.future.data_objects.typing import UnknownType
+
+
+def assign_items(
+    client: ClientCore,
+    team_slug: str,
+    dataset_ids: int | list[int],
+    assignee_id: int,
+    workflow_id: str,
+    filters: dict[str, UnknownType],
```
Nathanjp91, line 14: we have a datatype for this, JSONDict, which is basically a wrapper for dict[str, Any]. Nothing huge but good to keep consistent. It's in future data_objects I think?

Nathanjp91 (approved):
LGTM. Shouldn't be any issues.

darwin/future/core/items/assign_items.py
rslota: This is copy-pasted from the set_stage, but sure, I'll change it here.

owencjones (approved):
Some comments

darwin/future/meta/objects/item.py
```diff
@@ -129,6 +130,31 @@ def set_layout(self, layout: ItemLayout) -> None:
         filters = {"item_ids": [str(self.id)]}
         set_item_layout(self.client, team_slug, dataset_id, layout, filters)
 
+    def assign(self, assignee_id: int, workflow_id: str | None = None) -> None:
+        if not assignee_id:
+            raise ValueError("Must specify assignee to assign items to")
```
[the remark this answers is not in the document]
rslota: Urgh, I just merged this before I got email about your comments :( 
But this code is copied from other similar functions, so that keeps it consistent with them anyway.

PR #767, opened 2024-01-10
[Ai-1544] update darwin py image load path

ChristofferEdlund:
# Problem

In the export of darwin json annotations there are an `[item][name] `that viewed name of the file on the darwin platform. 
But there are also a` [item][slots][0][source_files][0][file_name] `that specifies the original name of the file, this name is also used for downloading images using darwin.

We had a use-case where a customer has re-named their files on the platform and the darwin-name is different from the original filename. This crashes our loading of local datasets since it is looking at images with the darwin-name.

# Solution
We changes the loading of local datasets so it looks at the original filename instead of the darwin-name. With this change we also had to update the original filename when converting the annotations from video to image so the new filename is a .png and not .mp4 or something that causes a miss-match.

Further did a bugfix in compute_distributions (used to split data into train, val, and test).

# Changelog
- Updated loading of local dataset to ensure consistency when files changes name on the platform
- Fixed a bug in compute_distributions used to split data into train, val and test.


darwin/utils/utils.py
```diff
@@ -522,13 +522,17 @@ def get_image_path_from_stream(
     """
     try:
         if json_version == "2.0":
+            # filename = Path(darwin_json["item"]["name"])
```
owencjones, line 525: Could we remove the comment

darwin/utils/utils.py
```diff
@@ -562,7 +566,12 @@ def get_darwin_json_version(annotations_dir: Path) -> str:
     str
         A str representing the Darwin JSON version.
     """
-    with open(next(annotations_dir.glob("*.json")), "r") as file:
+
+    # Search for JSON files both directly in the directory and in subdirectories
+    json_files = list(annotations_dir.glob("**/*.json"))
+
+    anno = json_files[0]
```
owencjones, line 573: Only referenced once, so could replace with just
```python
with open(json_files[0] "r") as file:
```

owencjones (commented):
Some to consider

darwin/dataset/utils.py
```diff
@@ -293,7 +286,7 @@ def update(*a):
 
 def get_coco_format_record(
     annotation_path: Path,
-    annotation_type: str = "polygon",
+    annotation_type: Union[str, None] = "polygon",
```
almazan, line 289: this cannot be `None`, right? cf. assert in line 294

owencjones (approved):
Lgtm, seeya all!

almazan (approved):

PR #768, opened 2024-01-10
[PY-601][PY-644] Advanced Filters support for ItemQuery

Nathanjp91:
# Problem
Advanced Filter objects not supported for ItemQuery Objects

# Solution
Build in support for AF On ItemQuery
Includes:
- Core endpoint: unstable api /items/list
- Changes to ItemQuery where syntax to allow GroupFilter | SubjectFilter
- Detection of Simple vs Advanced Filter backend api requirements
- Changes to support serialization of nested Pydantic objects in V2

# Changelog
Advanced Filters endpoint now supported for item querying


darwin/future/data_objects/advanced_filters.py
```diff
@@ -88,12 +88,16 @@ class NotContains(BaseMatcher):
 
 class SubjectFilter(BaseModel):
     subject: str
-    matcher: BaseMatcher
+    matcher: SerializeAsAny[BaseMatcher]
```
Nathanjp91, line 91: This is an important change and should be noted for anyone else doing pydantic object inheritance. BaseMatcher is basically never meant to be used singularly, it acts like an ABC. When pydantic model_dump is used, it wonly only ever serialize fields that exist on BaseMatcher, so if you inherit from that and add fields, those added fields won't be serialized. The solution is to add in pydantics SerializeAsAny wrapper

owencjones (approved):

rslota (approved):

PR #769, opened 2024-01-11
[PY-665] - Update property-value object structure to support un-nested value field

saurbhc:
# Problem
PropertyValue.value was `{'value': str}` before - which is being changed in the BE API.

# Solution
Update the value from dict to `str` and make sure we're passing the correct str structure wherever required.
TODO:
* [ ] Merge #759 to master before merging this PR. (as it's branched out from it).
* [ ] only merge this once the BE API changes are released to prod.

# Changelog
Update property-value object structure to support un-nested value field


Nathanjp91 (approved):
LGTM

PR #771, opened 2024-01-15
[PY-658][external] Identified and correctly named all multi-processed functions

JBWilkie:
# Problem
darwin-py uses multi-threading and multi-processing, but currently everything is labelled as multi-threading

# Solution
Identified and re-named multi-processing functions from `multi_threaded` to `multi_processed`

# Changelog
- Correctly labelled multi-processed functions as multi-processed instead of multi-threaded


saurbhc (approved):
LGTM :)

PR #773, opened 2024-01-23
[DAR-214][external Fixed unhandled annotation types in 1.0 conversion

JBWilkie:
# Problem
A previous PR introduced in 0.8.45 broke the following annotation types when converting from 2.0 to 1.0 JSON:
- Cuboid
- Ellipse
- Keypoint
- Skeleton

# Solution
- Introduced logic to handle these types during conversion
- Added a test for this logic

# Changelog
Fixed a bug that prevented certain annotation types from being converted from 2.0 to 1.0 JSON

### DIFF
--- a/darwin/exporter/formats/darwin_1_0.py
+++ b/darwin/exporter/formats/darwin_1_0.py
@@ -205,6 +205,9 @@ def _build_legacy_annotation_data(
         # Poygons and complex polygons usually have attached bounding_box annotations
         v1_data["bounding_box"] = data["bounding_box"]
 
+    if not v1_data:
+        v1_data[annotation_class.annotation_type] = data
+
     return v1_data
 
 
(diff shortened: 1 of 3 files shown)


darwin/exporter/formats/darwin_1_0.py
```diff
@@ -205,6 +205,9 @@ def _build_legacy_annotation_data(
         # Poygons and complex polygons usually have attached bounding_box annotations
         v1_data["bounding_box"] = data["bounding_box"]
 
+    if not v1_data:
+        v1_data[annotation_class.annotation_type] = data
+
```
JBWilkie, line 210: This is how the entire function used to work. Basically if it's not a tag, bbox, or polygon, simply return the structure that we used to

saurbhc (approved):
LGTM!

PR #776, opened 2024-01-30
[PLA-548][external] Resolved issue with item path filtering

JBWilkie:
# Problem
The `fetch_remote_files()` method does not accept `item_paths` as a filter

This led to some confusion as this was shown as being possible as per the following recipe: https://docs.v7labs.com/recipes/list-datasetitems-from-remotedataset-sdk

# Solution
Allow `item_paths` to be passed as a filter to `fetch_remote_files()`

# Changelog
Fixed bug preventing filtering by item path for an SDK function


saurbhc (approved):
Thank you, I can confirm it works :)

PR #778, opened 2024-02-12
[PLA-591][external] Fix for pushing subfolder structures on Windows

JBWilkie:
# Problem
When uploading subfolder structures on Windows, we were partly using local paths involving backslashes, causing situations like these:
06868fc877461b009f6df2b5f71dc66b

This was happening because we were converting `WindowsPath` objects with `str()`

# Solution
Make sure that all local paths are converted to `PosixPath` objects before `str()` conversion. This allows subfolder structures to be uploaded on Windows

# Changelog
Fixed bug preventing the upload of subfolder structures using darwin-py on Windows


saurbhc (approved):
Thank you 🙏

PR #779, opened 2024-02-19
[PLA-595] Don't create sub-annotation keyframe for every keyframe when importing annotations

saurbhc:
# Problem
As of now, when importing annotations, we automatically make every keyframe a "sub-annotation keyframe" by marking that the sub-annotation has changed.

# Solution
> during import, between D-Py and backend we still use this format which include both annotation and sub-annotation data at the same time, for each "keyframe". It would be kinda-possible to drop sub-annotation data from the request whenever we know it didn't change since last one, which likely wouldn't create on backend side sub-annotation keyframe.

# Changelog
Don't create sub-annotation keyframe for every keyframe when importing annotations

### DIFF
--- a/darwin/importer/importer.py
+++ b/darwin/importer/importer.py
@@ -277,7 +277,7 @@ def _resolve_annotation_classes(
 
 def _update_payload_with_properties(
     annotations: List[Dict[str, Unknown]],
-    annotation_id_property_map: Dict[str, Dict[str, Dict[str, Set[str]]]]
+    annotation_id_property_map: Dict[str, Dict[str, Dict[str, Set[str]]]],
 ) -> None:
     """
     Updates the annotations with the properties that were created/updated during the import.
@@ -290,13 +290,16 @@ def _update_payload_with_properties(
 
         if annotation_id_property_map.get(annotation_id):
             _map = {}
-            for _frame_index, _property_map in annotation_id_property_map[annotation_id].items():
+            for _frame_index, _property_map in annotation_id_property_map[
+                annotation_id
+            ].items():
                 _map[_frame_index] = {}
                 for prop_id, prop_val_set in dict(_property_map).items():
                     prop_val_list = list(prop_val_set)
                     _map[_frame_index][prop_id] = prop_val_list
 
             annotation["annotation_properties"] = dict(_map)
+
 
 def _import_properties(
     metadata_path: Union[Path, bool],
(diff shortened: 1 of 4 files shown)


JBWilkie (approved):
Tested and LGTM - Duplicate sub-annotations (those that haven't changed since the last change) are removed the import payload correctly. However, this doesn't appear to accomplish the overall goal of PLA-595, which makes me think it may require backend work

darwin/importer/importer.py
```diff
@@ -996,11 +1019,38 @@ def _handle_annotators(
...
+            ):
+                # drop sub-annotation whenever we know it didn't change since last one
+                # which likely wouldn't create on backend side sub-annotation keyframe.
+                # this is a workaround for the backend not handling duplicate sub-annotations.
+                continue
+            subs.append(sub)
+        last_subs = _annotation.subs
+        _annotation.subs = subs
+
+
```
JBWilkie, line 1047: @saurbhc I know I approved this, but I found an issue in the case that there is >1 sub-annotation per annotation. I made this change to support those cases

saurbhc: @JBWilkie thank you for updating the code 💯 

The code is already quite efficient, but there's a small optimisation you can make, I understand the intent to use `all()` and `any()` which both iterate over **all** elements -- you can use a set to store the previous sub-annotations and check if the current sub-annotation is in the set? something like:

```python
last_subs = set()
for _, _annotation in annotation.frames.items():
    _annotation: dt.Annotation
    subs = []
    for sub in _annotation.subs:
        sub_tuple = (sub.annotation_type, sub.data)
        if sub_tuple in last_subs:
            # drop sub-annotation whenever we know it didn't change since last one
            # which likely wouldn't create on backend side sub-annotation keyframe.
            # this is a workaround for the backend not handling duplicate sub-annotations.
            continue
        subs.append(sub)
        last_subs.add(sub_tuple)
    _annotation.subs = subs
```

-  `last_subs` is a set that stores tuples of `(sub.annotation_type, sub.data)`
-  This should be faster because checking if an element is in a set is an O(1) operation, while all() and any() are O(n) operations.


What do you think?

saurbhc: Ah, wait!

I just learned that if the sub-annotation (`sub.data`) is a list of attributes, then sub.data will be a **list** which is un-hashable. we can convert it to tuple first like

```python
sub_tuple = (sub.annotation_type, tuple(sub.data))
```

but it already seems a bit complicated and we might need to handle it for other datatypes too.

Let's Ignore this optimization bit! Thank you!!

JBWilkie: But thank you for your work here @saurbhc! Please merge whenever :)

PR #780, opened 2024-02-28
[PLA-640] Updating metadata property values while importing annotations

saurbhc:
# Problem
While importing annotations with properties, property-values which are not in annotation file but present in metadata.json are not created/updated in annotation-property.

# Solution

Adding additional logic after annotation-properties are created/updated on the server. Logic is:
1. get the latest team-properties
2. only go ahead in the case when annotation-properties are present in team-properties
3. _build metadata-property-values set_
4. _build team-property-values set_
5. extra values: _diff b/w metadata & team property-values set_
6. update property with extra-property-values!

# Changelog
Updating additional metadata property values while importing annotations

### DIFF
--- a/darwin/importer/importer.py
+++ b/darwin/importer/importer.py
@@ -557,6 +557,44 @@ def _import_properties(
             )
             updated_properties.append(prop)
 
+    for fp in created_properties + updated_properties:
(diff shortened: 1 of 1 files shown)


JBWilkie (approved):
Tested a few different scenarios importing single / multi select property values that didn't yet exist server-side, and it worked perfectly every time - LGTM!

Particularly like that it prints to console, nice visual feedback!

PR #781, opened 2024-02-29
[PLA-672][external] Support for importing simple_table annotations

JBWilkie:
# Problem
The optional `simple_table` annotation type can be exported, but not imported with darwin-py. This is because darwin-py doesn't have an annotation constructor for this type

# Solution
Added a `make_simple_table()` function 

# Changelog
Added support to darwin-py for importing simple_table annotations

### DIFF
--- a/darwin/datatypes.py
+++ b/darwin/datatypes.py
@@ -998,6 +998,51 @@ def make_table(
     )
 
 
+def make_simple_table(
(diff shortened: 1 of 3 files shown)


saurbhc (approved):
Thank you 🙏 LGTM 💯

PR #783, opened 2024-03-06
[PLA-683][external] 2 Minor fixes for importing of PASCAL VOC annotations

JBWilkie:
# Problem
2 issues:
- 1:  `get_remote_files()` was calling `fetch_remote_files()` with an incorrect filter parameter `filenames` instead of `item_names`
- 2: PASCAL VOC annotations could not be imported to dataset items outside of the root folder

# Solution
- Fix for problem 1: Updated `get_remote_files()` so that it passes `item_names` instead of `filenames` to `fetch_remote_files()`
- Fix for problem 2: Added logic to split the filename into `remote_path` and an updated filename if a remote path is included in the filename

# Changelog
- Fixed a misnamed querystring filter parameter in `fetch_remote_files()`
- Added support for importing PASCAL VOC annotations to dataset items outside of the root folder '`/`'


saurbhc (approved):
Thank you 🙏

PR #785, opened 2024-03-09
[PLA-696][external] Allow registration of single-slotted read-write files from external storage

JBWilkie:
# Problem
Currently, the REST API is the only way to register externally stored items. The ability to register items using darwin-py is simpler in some scenarios and can prevent difficult-to-spot payload errors

# Solution
Added user-facing objects & methods to register single-slotted read-write items from external storage

# Changelog
Added ability to register single-slotted read-write files from external storage in darwin-py


darwin/dataset/utils.py
```diff
@@ -840,3 +845,51 @@ def sanitize_filename(filename: str) -> str:
...
+        Whether to preserve folders or place the file in the Dataset root.
+
+    Returns
+    -------
+    str
+        The parsed external file path.
+    """
+    if not preserve_folders:
+        return "/"
+    return "/" + "/".join(storage_key.split("/")[:-1])
```
saurbhc, line 895:
```suggestion
    return "/".join(storage_key.split("/")[:-1]) + "/"
```

Is the `/` on the end of it? as this could give us double `/` in the start of the string, for example:
```python
$ python
>>> storage_key = "/Users/saurabhchopra/.darwin/image.jpg"
>>> "/" + "/".join(storage_key.split("/")[:-1])
'//Users/saurabhchopra/.darwin'
```

darwin/dataset/remote_dataset_v2.py
```diff
@@ -552,3 +558,64 @@ def _build_image_annotation(
...
+            }
+            if fps and file_type == "video":
+                item["fps"] = fps
+            if multi_planar_view and file_type == "dicom":
+                item["extract_views"] = "true"
+            items.append(item)
+
+        # Do not register more than 500 items in a single request
+        chunked_items = [items[i : i + 500] for i in range(0, len(items), 500)]
+        print(f"Registering {len(items)} items in {len(chunked_items)} batch(es)")
```
saurbhc, line 593: we could use a generator expression instead of a list comprehension for `chunked_items`. This will create the chunks on-the-fly during the iteration, which can save memory if items is very large:

```suggestion
        # Do not register more than 500 items in a single request
        chunk_size = 500
        chunked_items = (items[i : i + chunk_size] for i in range(0, len(items), chunk_size))
        print(f"Registering {len(items)} items in chunks of {chunk_size} items...")
```

_Python also introduced a [itertools.batched](https://docs.python.org/3.12/library/itertools.html#itertools.batched) function in python3.12 (they also added an 'equivalent function' in the same docs as well for older python versions - maybe we can use that? - just an optimisation to our chunking approach)_

darwin/objectstore.py
saurbhc, line 1: maybe we can move this to `/darwin/datatypes.py`?

tests/darwin/cli_functions_test.py
```diff
@@ -47,7 +47,7 @@ def test_default_non_verbose(
         remote_dataset: RemoteDataset,
         request_upload_endpoint: str,
     ):
-        request_upload_response = response = {
+        request_upload_response = {
```
saurbhc, line 50: Good Spot 💯

saurbhc (approved):
Thank you 🙏

darwin/objectstore.py
JBWilkie: This looks like a much more natural place for it, nice spot

darwin/dataset/remote_dataset_v2.py
JBWilkie: Great suggestion! Sadly at the moment we're support `[>=3.8.0, <3.12](https://pypi.org/project/darwin-py/)`, I think moving to 3.12 for this change would be too disruptive. What do you say to sticking with the generator expression?

darwin/dataset/utils.py
JBWilkie: II don't believe item paths are allowed to end with a `/`

I put a [/](url) at the front for two reasons:
- 1: Storage keys themselves [cannot start with a](https://docs.v7labs.com/docs/registering-items-from-external-storage#the-basics-1) `/`, therefore we shouldn't ever run into doubles
image
- 2: If the storage key is in the root folder of the bucket, it'll just be filename.jpg (for example), and to preserve the folder structure, path needs to be `/filename`
- 
However, I can imagine users passing storage keys starting with `/` doing this anyway. What do you think about raising an error if this happens?

saurbhc: > This file path is case sensitive, cannot start with a forward slash

My bad, I missed reading this part - That makes sense!! 🙏

PR #786, opened 2024-03-10
Update utils.py line 576

gerdac:
update to line 576

image_path = get_image_path_from_stream(darwin_json, images_dir, with_folders = 0, annotation_filepath = annotation_path)

# Problem
Unable to use trainer because the following error is thrown:
TypeError: get_image_path_from_stream() missing 2 required positional arguments: 'with_folders' and 'annotation_filepath'

get_image_path_from_stream() command requires all 4 inputs. It does not work otherwise. 

# Solution
included the arguments 'with_folders' and 'annotation_filepath'

# Changelog
(Will appear in release docs)

### DIFF
--- a/darwin/dataset/utils.py
+++ b/darwin/dataset/utils.py
@@ -573,7 +573,7 @@ def _map_annotations_to_images(
     invalid_annotation_paths = []
     for annotation_path in annotations_dir.glob("**/*.json"):
         darwin_json = stream_darwin_json(annotation_path)
-        image_path = get_image_path_from_stream(darwin_json, images_dir)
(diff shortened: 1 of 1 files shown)


JBWilkie:
Hi @gerdac, I apologise for the significant delay here! However, I'm happy to report that the latest version of darwin-py (1.0.0) made some significant changes to the way we name and read local files. You can read about them [here](https://github.com/v7labs/darwin-py/releases/tag/v1.0.0). Part of this work resolved this issue with the missing arguments in this call of the `get_image_path_from_stream()` function.

PR #787, opened 2024-03-14
[PLA-702] Add dependabot for pip and GitHub Actions

saurbhc:
# Problem
running tests for #786, I noticed out of date github-actions.

# Solution
I've added a dependabot config(s) to make sure we stay relatively up to date.
  - for github-actions to run weekly
  - for pip to run weekly and ignore major version updates (as it could be breaking changes)

# Changelog
Add dependabot for pip and GitHub Actions


JBWilkie (approved):
This is a great idea, nice one!

PR #788, opened 2024-03-14
[PLA-585][external] Allow imports where some annotation frames are missing a main type

JBWilkie:
# Problem
This PR solves one main problem (1) but includes lots of clean up and refactoring of old V1 code and concepts as well (2 & 3). It's a very large PR, so I've annotated it heavily to help with reviewing it

- 1: In IO-1163, we introduced logic to exports to compress the resulting JSON **if** we determine that the export size was sufficiently large that it would cause it to OOM. This compression took the form of only exporting keyframes, and adding an `only_keyframes` field in the JSON to reflect this. Unfortunately, darwin-py import logic was built on the assumption that every frame has a main annotation type, which this change violates, breaking these imports
- 2: Removal of the concept of a `complex_polygon` annotation type. In V1, we stored polygons as complex polygons if they had >1 path. We removed this distinction in V2, instead referring to all polygons as type `polygon`. Therefore we should update darwin-py to reflect this
- 3: Removal of old functions & erroneous deprecation messages. Darwin-py contains a lot of unused functions from the V1 version of the platform. These have already been re-written for V2 and will thus never be used again. There were also many deprecation messages that were not needed, out of date or wrong (functions were actually in use)

# Solution
- 1: For each annotation with `only_keyframes` set to True, get the first frame with a main type. Store this type & annotation data. Then, iterate through each frame of the annotation. If the main type is missing, use the stored value to populate it for import. Otherwise if the main type is present, update the stored annotation data. Repeat this for the entire annotation
- 2: Removed the concept of complex polygons from darwin-py, instead referring to all polygons simply as type `polygon` to reflect how we store polygons in our DB
- 3: Removed the unused functions, and removed incorrect deprecation messages

# Changelog
- Updated import logic to support importing of annotations that were compressed during export to avoid memory cap errors
- Removed the distinction between complex polygons & polygons
- Removed unused functions & erroneous deprecation warnings


darwin/dataset/download_manager.py
```diff
@@ -28,15 +27,8 @@
...
-from darwin.version import __version__
 
 
-@deprecation.deprecated(
-    deprecated_in="0.7.5",
-    removed_in="0.8.0",
-    current_version=__version__,
-    details="The api_url parameter will be removed.",
-)
 def download_all_images_from_annotations(
```
JBWilkie, line 32: This function is used by `pull()`

darwin/exporter/formats/coco.py
```diff
@@ -401,7 +82,7 @@ def _calculate_tag_categories(
                 categories[annotation_class.name] = _calculate_category_id(
                     annotation_class
                 )
-    return categories
+    return dict(sorted(categories.items(), key=lambda item: item[1]))
```
JBWilkie, line 85: These changes are to Ensure COCO categories are always in the same, ascending order to avoid E2E flakiness

darwin/importer/importer.py
```diff
@@ -71,12 +69,6 @@
 """
 
 
-@deprecation.deprecated(  # type:ignore
-    deprecated_in="0.7.12",
-    removed_in="0.8.0",
-    current_version=__version__,
-    details=DEPRECATION_MESSAGE,
-)
 def build_main_annotations_lookup_table(
```
JBWilkie, line 72: This is used when importing annotations

darwin/importer/importer.py
```diff
@@ -108,12 +100,6 @@ def build_main_annotations_lookup_table(
     return lookup
 
 
-@deprecation.deprecated(  # type:ignore
-    deprecated_in="0.7.12",
-    removed_in="0.8.0",
-    current_version=__version__,
-    details=DEPRECATION_MESSAGE,
-)
 def find_and_parse(  # noqa: C901
```
JBWilkie, line 103: This is used when importing annotations

darwin/importer/importer.py
```diff
@@ -199,12 +179,6 @@ def build_attribute_lookup(dataset: "RemoteDataset") -> Dict[str, Unknown]:
     return lookup
 
 
-@deprecation.deprecated(  # type:ignore
-    deprecated_in="0.7.12",
-    removed_in="0.8.0",
-    current_version=__version__,
-    details=DEPRECATION_MESSAGE,
-)
 def get_remote_files(
```
JBWilkie, line 182: This is used when importing annotations

darwin/importer/importer.py
```diff
@@ -1035,15 +1009,17 @@ def _handle_subs(
...
-    if "complex_polygon" in data:
-        del data["complex_polygon"]
-        data["polygon"] = {
-            "path": annotation.data["paths"][0],
-            "additional_paths": annotation.data["paths"][1:],
-        }
+    if "polygon" in data:
+        if len(annotation.data["paths"]) > 1:
+            data["polygon"] = {
+                "path": annotation.data["paths"][0],
```
JBWilkie, line 1018: When importing polygons, they have to be structured as a primary path, and any subsequent paths have to be passed in `additional_paths` in the payload

darwin/utils/utils.py
```diff
@@ -128,7 +106,6 @@ def is_image_extension_allowed_by_filename(filename: str) -> bool:
     return any(filename.lower().endswith(ext) for ext in SUPPORTED_IMAGE_EXTENSIONS)
 
 
-@deprecation.deprecated(deprecated_in="0.8.4", current_version=__version__)
 def is_image_extension_allowed(extension: str) -> bool:
```
JBWilkie, line 109: This is used when pulling exports

JBWilkie:
@saurbhc I've actioned all points of feedback

saurbhc (approved):
Thank you 🙏

PR #789, opened 2024-03-15
[DAR-1104][External] Fix colours mismatch between files on semantic mask export

AndriiKlymchuk:
# Problem
The same semantic mask could be represented by different colors in the files of the same export.

# Solution
The root cause of the issue is an assumption that labels in the `raster_layer.data.mask_annotation_ids_mapping` are the same across different files which is not true. Because of that assumption labels were used as colors ids causing the mismatch.
I changed the logic to have static color ids for all files and convert labels into color ids for every file separately. 

# Changelog
Fix colours mismatch between files on semantic mask export

### DIFF
--- a/darwin/datatypes.py
+++ b/darwin/datatypes.py
@@ -1400,7 +1400,6 @@ class MaskTypes:
     CategoryList = List[str]
     ExceptionList = List[Exception]
     UndecodedRLE = List[int]
-    DecodedRLE = List[List[int]]
     ColoursDict = Dict[str, int]
     RgbColors = List[int]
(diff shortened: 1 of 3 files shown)


JBWilkie (approved):
Tested by comparing output masks of master branch to this one, always had consistent colour mappings on this branch

LGTM!

shernshiou (approved):

saurbhc (approved):
Thank you 🙏

AndriiKlymchuk:
@umbertoDifa you were to fast. I found one issue during testing. Opening another PR.

PR #791, opened 2024-03-15
[PLA-256] Existing properties are now imported without a `.v7/manifest.json file 

saurbhc:
# Problem
When importing annotations with property values, if the  .v7/manifest.json file isn't present, properties will be entirely skipped.

# Solution
Team properties (if they exist) are passed along with annotations while importing them.

# Changelog
Existing properties are now imported without a .v7/manifest.json file


JBWilkie (approved):
Tested importing existing property values (single & multi-select) without the manifest and it worked in all cases

Imports also correctly raised ValueError when trying to import property values that don't exist in the team

Nice one!

PR #803, opened 2024-03-19
[PLA-639][External] Allow properties to be imported for annotations without IDs

JBWilkie:
# Problem
Currently, annotations with properties can only be imported if the annotation has an ID in the JSON file. Annotation IDs are not required for any other import functionality, so we should remove this requirement

# Solution
When importing annotations, if an annotation does not have an ID, give it one with `uuid.uuid4()`. This works because we don't actually respect the ID that's assigned during import (to avoid clashes), it just needs to be a unique value per annotation

# Changelog
Removed the requirement for annotation IDs to be present in imported JSON to successfully import properties

### DIFF
--- a/darwin/importer/importer.py
+++ b/darwin/importer/importer.py
@@ -1,3 +1,4 @@
+import uuid
 from collections import defaultdict
 from logging import getLogger
 from multiprocessing import cpu_count
@@ -281,9 +282,9 @@ def _get_team_properties_annotation_lookup(client):
     team_properties = client.get_team_properties()
 
     # (property-name, annotation_class_id): FullProperty object
-    team_properties_annotation_lookup: Dict[
-        Tuple[str, Optional[int]], FullProperty
-    ] = {}
+    team_properties_annotation_lookup: Dict[Tuple[str, Optional[int]], FullProperty] = (
+        {}
+    )
     for prop in team_properties:
         team_properties_annotation_lookup[(prop.name, prop.annotation_class_id)] = prop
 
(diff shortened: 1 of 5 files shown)


darwin/importer/importer.py
```diff
@@ -1308,6 +1309,10 @@ def _import_annotations(
 
         if annotation.id:
             serial_obj["id"] = annotation.id
+        else:
+            annotation_id = str(uuid.uuid4())
+            annotation.id = annotation_id
+            serial_obj["id"] = annotation_id
```
JBWilkie, line 1315: This gives every annotation an ID if one isn't present, even those that we're not importing properties to. Because of this, I'm not extremely happy with it as a solution, but it is by far the simplest solution I could think of. IDs have to be added here before `_import_properties()` & `_update_payload_with_properties()`

I don't think inefficiency this will have a significant impact. However, I'm open to feedback on alternative approaches!

tests/darwin/importer/importer_test.py
```diff
@@ -572,7 +572,18 @@ def test__import_annotations() -> None:
...
+            == assertion["annotations"][0]["annotation_class_id"]
+        )
+        assert output["annotations"][0]["data"] == assertion["annotations"][0]["data"]
+        assert (
+            output["annotations"][0]["actors"] == assertion["annotations"][0]["actors"]
+        )
+        assert (
+            output["annotations"][0]["context_keys"]
+            == assertion["annotations"][0]["context_keys"]
+        )
```
JBWilkie, line 586: Because the importer assigns a random UUID, comparing the entire `annotations` field for each no longer works. Instead, we need to compare the individual fields inside `annotations`

darwin/importer/importer.py
saurbhc, line 1315: simplifying this one:

```suggestion
        annotation.id = annotation.id or str(uuid.uuid4())
        serial_obj["id"] = annotation.id
```

saurbhc (approved):
Thank you 🙏

PR #809, opened 2024-04-04
[PLA-801][external] fetch_remote_files() is called with an old, V1 argument

JBWilkie:
# Problem
- Changes made to darwin-py in 0.8.58 to remove V1 code removed a V1 backwards compatibility "handler" for an older version of a filter parameter when retrieving a list of remote files in `fetch_remote_files()`

# Solution
- Re-add this backward compatibility and update all invocations of `fetch_remote_files()` to use `item_names` instead of `filenames`

# Changelog
- Re-added backward compatibility for a V1 filter parameter when fetching remote files


saurbhc (approved):
LGTM 🙏

PR #813, opened 2024-04-12
Fixed bug when importing annotations with properties to non-default teams

JBWilkie:
# Problem
When importing annotations with properties to a non-default team, a codepath is invoked that creates a client object that is always for the default team as per the `config.yaml` file. This results in an authorization error for non-default teams

# Solution
Create a client object that is for the team containing the target dataset

# Changelog
 Fixed a bug when importing annotations to non-default teams


saurbhc (approved):
Thank you for fixing this 🙏

PR #816, opened 2024-04-15
[DAR-1609][External] pull() never creates the properties metadata file in the annotations directory

JBWilkie:
# Problem
When downloading annotation ZIPs when using `pull()`, the properties manifest file is never created in the annotations directory. This is because despite the fact that we download the same ZIP file that the frontend provides, we first unzip the files in a temporary directory, then only move `.json` files to the annotations directory. This leaves the `.v7` folder behind

# Solution
If the properties metadata file is present, move it to the annotations directory

# Changelog
Fixed a bug where downloading annotations with darwin-py would not include the properties metadata file


darwin/dataset/remote_dataset.py
```diff
@@ -278,6 +278,13 @@ def pull(
                 annotations_dir.mkdir(parents=True, exist_ok=False)
                 stems: dict = {}
 
+                # If properties were exported, move the manifest file to the annotations folder
+                if (tmp_dir / ".v7").exists():
+                    metadata_file = tmp_dir / ".v7" / "metadata.json"
+                    metadata_dir = annotations_dir / ".v7"
+                    metadata_dir.mkdir(parents=True, exist_ok=False)
```
saurbhc, line 285: - changing the `exist_ok=False` -> `True` - to make sure it dosen't give an error. Docs: https://docs.python.org/3/library/os.html#os.makedirs

```suggestion
                    metadata_dir.mkdir(parents=True, exist_ok=True)
```

saurbhc:
Let's also add a test case, maybe [here](https://github.com/v7labs/darwin-py/blob/master/tests/darwin/dataset/remote_dataset_test.py#L651)? - which makes sure it moves the metadata.json file

saurbhc (approved):
Thank you 🙏

PR #818, opened 2024-04-18
update process_nifti to take custom ornt

dorfmanrobert:
# Problem
(describe the problem here)

# Solution

# Changelog
(Will appear in release docs)

### DIFF
--- a/darwin/importer/formats/nifti.py
+++ b/darwin/importer/formats/nifti.py
@@ -470,21 +470,18 @@ def correct_nifti_header_if_necessary(img_nii):
 
 
 def process_nifti(
-    input_data: Union[Sequence[nib.nifti1.Nifti1Image], nib.nifti1.Nifti1Image]
+    input_data: nib.nifti1.Nifti1Image,
+    ornt: Optional[List[List[float]]] = [[0.0, -1.0], [1.0, -1.0], [2.0, -1.0]]
 ):
     """
-    Function which takes in a single nifti path or a list of nifti paths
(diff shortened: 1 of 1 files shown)


shernshiou (approved):

darwin/importer/formats/nifti.py
```diff
@@ -470,21 +470,18 @@ def correct_nifti_header_if_necessary(img_nii):
...
 def process_nifti(
-    input_data: Union[Sequence[nib.nifti1.Nifti1Image], nib.nifti1.Nifti1Image]
+    input_data: nib.nifti1.Nifti1Image,
+    ornt: Optional[List[List[float]]] = [[0.0, -1.0], [1.0, -1.0], [2.0, -1.0]]
 ):
     """
-    Function which takes in a single nifti path or a list of nifti paths
-    and returns the pixel_array, affine and pixdim
+    Function which takes in a single nibabel nifti object 
+    and returns the pixel_array and pixdims
```
JBWilkie, line 478: Might it be a good idea to add a little more detail to the docstring? From a non-medical perspective it's not immediately obvious what this function does, perhaps a string in the style of the `affine_to_spacing()` function above? i.e:

- Description
- Args
- Returns

JBWilkie (approved):

darwin/importer/formats/nifti.py
dorfmanrobert: definitely agree, updated this with details

PR #825, opened 2024-04-22
[DAR-1562][External] Allow pulling of unpopulated required property annotations

JBWilkie:
# Problem
When pulling annotations with unpopulated required properties, the `frame_index` field will be set to `null`. This caused a Pydantic `ValidationError` because it needs to be type `int`

# Solution
It's required that darwin-py can pull unpopulated required properties (but not import them), so allow the `SelectedProperty` `frame_index` attribute to be `None`

# Changelog
Fixed bug preventing darwin-py from being able to pull annotations with unpopulated required properties


darwin/future/data_objects/properties.py
```diff
@@ -147,7 +147,7 @@ class SelectedProperty(DefaultDarwin):
         value (str): Value of the property
     """
 
-    frame_index: int
+    frame_index: Optional[int] = None
```
JBWilkie, line 150: This is the only meaningful change in this PR, everything else is black or ruff formatting / linting

tests/darwin/exporter/formats/export_darwin_test.py
```diff
@@ -128,3 +129,8 @@ def test_complete_annotation_file_with_bounding_box_and_tag_v2():
     }
 
     assert build_image_annotation(annotation_file, "Test team") == expected_output
+
+
+def test_parse_annotations_with_missing_required_properties():
+    path = Path("tests/darwin/data/annotation_with_missing_required_properties.json")
+    annotaions = parse_darwin_json(path=path)
```
JBWilkie, line 136: This tests that `SelectedProperty` objects can be created even if `frame_index` & `value` fields are `None`

darwin/future/tests/data_objects/test_properties.py
```diff
@@ -35,3 +35,10 @@ def test_properties_metadata_fails() -> None:
     path = Path("darwin/future/tests/data/does_not_exist")
     with pytest.raises(FileNotFoundError):
         MetaDataClass.from_path(path)
+
+
+def test_can_parse_unpopulated_required_properties() -> None:
+    selected_property = SelectedProperty(
+        frame_index=None, name="name", type="type", value=None
+    )
+    assert selected_property is not None
```
JBWilkie, line 44: This test ensures `SelectedProperty` objects can be instantiated even when `frame_index` & `value` are of type  `None`

umbertoDifa (approved):

PR #827, opened 2024-04-25
[PY-668][Internal] Exclude macos from testing on 3.8 and 3.9

shernshiou:
# Problem
From https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json, Python 3.8 and Python 3.9 is not available for Arm64 / MacOS.

# Solution
Exclude Python 3.8 and 3.9 test for MacOS.

### DIFF
--- a/.github/workflows/JOB_tests.yml
+++ b/.github/workflows/JOB_tests.yml
@@ -12,6 +12,12 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: ["3.8", "3.9", "3.10", "3.11"]
+        exclude:
+          - os: macos-latest
+            version: "3.8"
+          - os: macos-latest
+            version: "3.9"
+
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/JOB_tests.yml
+++ b/.github/workflows/JOB_tests.yml
@@ -14,14 +14,14 @@ jobs:
         python-version: ["3.8", "3.9", "3.10", "3.11"]
         exclude:
           - os: macos-latest
-            version: "3.8"
+            python-version: "3.8"
           - os: macos-latest
-            version: "3.9"
+            python-version: "3.9"
         include:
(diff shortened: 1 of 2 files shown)


.github/workflows/JOB_tests.yml
```diff
@@ -12,6 +12,12 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: ["3.8", "3.9", "3.10", "3.11"]
+        exclude:
+          - os: macos-latest
+            version: "3.8"
+          - os: macos-latest
+            version: "3.9"
+
```
saurbhc, line 20: You're right that 3.8 and 3.9 versions are not available on `macos-latest` runners (which now points to M1), but they're available in the pre-M1 runners (`macos-13`).

python/peps has a nice workaround to it - https://github.com/python/peps/pull/3763/files

Adding include blocks for 3.8 and 3.9 (which might do the trick) 
```suggestion
        exclude:
          - os: macos-latest
            version: "3.8"
          - os: macos-latest
            version: "3.9"
        include:
          - os: macos-13
            version: "3.8"
          - os: macos-13
            version: "3.9"
```

.github/workflows/JOB_tests.yml
```diff
@@ -12,6 +12,16 @@ jobs:
...
+        exclude:
+          - os: macos-latest
+            version: "3.8"
+          - os: macos-latest
+            version: "3.9"
+        include:
+          - os: macos-13
+            version: "3.8"
+          - os: macos-13
+            version: "3.9"
```
saurbhc, line 24: This could solve this issue - https://github.com/v7labs/darwin-py/actions/runs/8830086377

```suggestion
        exclude:
          - os: macos-latest
            python-version: "3.8"
          - os: macos-latest
            python-version: "3.9"
        include:
          - os: macos-13
            python-version: "3.8"
          - os: macos-13
            python-version: "3.9"
```

saurbhc (approved):
Thank you! 🥇

PR #828, opened 2024-04-25
[DAR-1165][DAR-1734][External] Polygons from other plane export correctly

shernshiou:
# Problem
The polygons at Coronal and Sagittal are not exported.

# Solution
Go through all the slots to ensure all the polygons are exported. 

Fixes DAR-1165
Fixes DAR-1734

# Changelog
- Polygon from Coronal and Sagittal plane are exported correctly.


shernshiou:
Result

image

JBWilkie (approved):
Tested fix on problematic data provided to us by the client, and NifTI exports of those files are no longer missing any annotations

PR #832, opened 2024-04-30
[DAR-1734][External] Re-add functions used by the backend

JBWilkie:
# Problem
The Darwin backend makes use of two functions that were previously removed from darwin-py

# Solution
Re-add those functions

# Changelog
Re-added functions used by the Darwin backend


saurbhc (approved):
LGTM

PR #838, opened 2024-05-09
[DAR-1639][External] Unit tests & bug fixes for in-platform model training

JBWilkie:
# Problem
Darwin's in-platform model training uses darwin-py to pull and load data into workable structures. A lack of unit tests in darwin-py has prevented us from updating the darwin-py versions that are used. These versions contain bugs

# Solution
To allow in-platform model training to use the latest versions of darwin-py, this PR:
- 1: Introduces tests for loading and splitting image & video datasets as `ClassificationDataset`, `InstanceSegmentationDataset`, and `ObjectDetectionDataset` objects
- 2: Fixes a few bugs with saving and loading dataset releases into data loaders

Some of these fixes contain changes to the way darwin-py names files when saving and loading them. The PR also contains the following:
- Fix for converting Darwin JSON 2.0 polygons to the YOLOv8 format (This was causing E2E failures)
- Fix for E2E COCO conversion test

# Changelog
- Unit tests for in-platform model training
- Fixed a handful of small bugs related to saving and loading datasets into dataloaders
- Fixed 2 JSON conversion issues


darwin/dataset/download_manager.py
```diff
@@ -95,11 +95,13 @@ def download_all_images_from_annotations(
...
-                / Path(annotation.remote_path.lstrip("/\\")).resolve().absolute()
-                / Path(annotation.filename)
-            )
+            filename = Path(annotation.slots[0].source_files[0]["file_name"])
+            if use_folders and annotation.remote_path != "/":
+                planned_image_path = (
+                    images_path / Path(annotation.remote_path.lstrip("/\\")) / filename
+                )
+            else:
+                planned_image_path = images_path / filename
```
JBWilkie, line 104: When downloading a release, darwin-py skips files that have already been downloaded in previous `pull()` operations by checking `planned_image_path` against the local filesystem. These changes fix a bug whereby `planned_image_path` only worked for files that were downloaded with folders

darwin/dataset/download_manager.py
```diff
@@ -339,13 +340,7 @@ def _download_single_slot_from_json_annotation(
...
             image_url = image["url"]
             image_filename = image["file_name"]
-
-            if not use_folders:
-                suffix = Path(image_filename).suffix
-                stem = annotation_path.stem
-                filename = str(Path(stem + suffix))
-            else:
-                filename = slot.source_files[0]["file_name"]
+            filename = slot.source_files[0]["file_name"]
```
JBWilkie, line 343: Currently when pulling single-slotted items:
- If pulling without folders, we name the resulting files based on `item.name`
- If pulling with folders, we use `source_files[0].file_name`

`item.name` is the alias that the item has in Darwin. This can be changed by users, but `source_files[0].file_name` cannot. This change is a part of a push to unify how we name, save, & load single-slotted items

One thing to note about this change is that it's possible to register the same file from external storage (same `source_files[0].file_name`) as two differently named (`item.name`) dataset items in the same dataset folder. With these changes, pulling a dataset with these files would result in one overwriting the other **if** it was pulled without folder

darwin/dataset/utils.py
```diff
@@ -498,49 +502,61 @@ def get_annotations(
...
+def _get_annotation_filepaths_from_split(
+    release_path: Path,
+    annotation_type: str,
+    partition: str,
+    split_type: str,
+    split: Optional[str] = "default",
+) -> Generator[str, None, None]:
     """
-    Determines the file stems based on the dataset split and other parameters.
+    Determines the filepaths based on the dataset split and other parameters.
```
JBWilkie, line 537: This function used to work with file path stems. However, to support datasets with identically named items in different folders, splits have to record the full file path to each annotation file

darwin/dataset/utils.py
```diff
@@ -934,3 +961,28 @@ def chunk_items(items: List[Any], chunk_size: int = 500) -> Iterator[List[Any]]:
...
+    Returns
+    -------
+    dt.AnnotationFile
+        The corrected annotation.
+    """
+    frame_path = annotation_file.stem / Path(annotation["item"]["name"].split("/")[-1])
+    annotation["item"]["slots"][0]["source_files"][0][
+        "file_name"
+    ] = frame_path.as_posix()
+    return annotation
```
JBWilkie, line 988: This function is only ever invoked with `split_video_annotations()` is used. Because:
- 1: This PR means that image files are always read from `source_files[0].file_name` (changing from `item.name`)
- 2: Running `split_video_annotations()` creates another layer of directories containing the annotation file for each video frame

It's necessary to include this directory in this field so that each frame's image can be found when loading the dataset

darwin/importer/importer.py
```diff
@@ -1399,8 +1399,9 @@ def _overwrite_warning(
     """
     files_to_overwrite = []
     for local_file in local_files:
+        item_id = remote_files.get(local_file.full_path)[0]
         remote_annotations = client.api_v2._get_remote_annotations(
-            local_file.item_id,
+            item_id,
             dataset.team,
```
JBWilkie, line 1405: `item_id` is not a requirement to import annotation files. Therefore, read it from the corresponding remote file, where we are guaranteed to find it

darwin/utils/utils.py
```diff
@@ -1463,13 +1466,14 @@ def get_annotation_files_from_dir(path: Path) -> Iterator[str]:
...
-    Iterator[str]
+    Generator[str]
         An iterator of all the JSON annotation files in the given directory.
     """
+
     return (
         str(filepath)
         for filepath in sorted(path.glob("**/*.json"))
-        if "/.v7/" not in str(filepath)
+        if "/.v7/" not in str(filepath) and "\\.v7\\" not in str(filepath)
```
JBWilkie, line 1476: Necessary to support Windows filepaths

darwin/exporter/exporter.py
```diff
@@ -31,7 +31,7 @@ def darwin_to_dt_gen(
     count = 0
     for file_path in map(Path, file_paths):
         files = (
-            list(get_annotation_files_from_dir(file_path))
+            list(map(Path, get_annotation_files_from_dir(file_path)))
```
JBWilkie, line 34: Each object in `files` has to be of type `Path`

darwin/exporter/formats/yolo_segmented.py
```diff
@@ -180,13 +180,22 @@ def _handle_polygon(
...
-        if "path" in data:
+        if "paths" in data:
+            paths_data = data["paths"]
+            if len(paths_data) > 1:
+                raise DarwinException from ValueError(
+                    "Complex polygon detected. The YOLOV8 format only supports simple polygons with a single path."
+                )
+            path_data = paths_data[0]
+        elif "path" in data:
+            # Continuing to support old version, in case anything relies on it.
```
JBWilkie, line 191: This fixes a bug with polygons in this conversion format. Since all Darwin JSON 2.0 polygons are represented as `paths` (even simple polygons with a single path), we need to make sure we don't ignore them

e2e_tests/data/yolov8/to/test_input_with_bboxes_and_polys.txt
```diff
@@ -1,3 +1,4 @@
...
 0 0.59466 0.112 0.9053199999999999 0.112 0.9053199999999999 0.32265999999999995 0.59466 0.32265999999999995
```
JBWilkie, line 2: This change is necessary after fixing the polygon > YOLOv8 conversion issue above - This is the corrected version of the output file

e2e_tests/data/yolov8/from/test_input_with_bboxes_and_polys.json
```diff
@@ -61,9 +61,7 @@
             {
               "x": 158.18,
               "y": 242.01
-            }
-          ],
-          [
+            },
```
JBWilkie, line 64: YOLOv8 doesn't support complex polygons, so adjust the convert test input to be simple polygons

e2e_tests/data/coco/to/output.json
```diff
@@ -85,13 +85,13 @@
...
+      "id": 348813479,
+      "name": "test_bb",
       "supercategory": "root"
     },
     {
-      "id": 348813479,
-      "name": "test_bb",
+      "id": 3961009249,
+      "name": "test_poly",
       "supercategory": "root"
```
JBWilkie, line 95: This was causing another E2E test failure where conversions to COCO [now output consistent, sorted annotation categories](https://github.com/v7labs/darwin-py/pull/788/commits/6b1560acb4bc68eedc043ec7fc2db979785fbcd4). Previously, they could be generated in different orders from run to run

tests/model_training/test_object_detection_model_training.py
```diff
@@ -0,0 +1,92 @@
+import tempfile
+from pathlib import Path
+from zipfile import ZipFile
+
+from darwin.dataset.split_manager import split_dataset
+from darwin.torch.dataset import ObjectDetectionDataset
+
+
+class TestObjectDetectionDatasetLoading:
```
umbertoDifa, line 9: This test and the others are very similar, we might have somthing like this to reduce duplicated code. The same can be applied for videos.
```

def run_split_and_load_dataset_test(dataset_name, dataset_class, expected_splits, split_values):
    with tempfile.TemporaryDirectory() as tmpdir:
        with ZipFile("tests/model_training_data.zip") as zfile:
            zfile.extractall(tmpdir)
            dataset_path = Path(tmpdir) / "model_training_data" / dataset_name

            split_dataset(
                dataset_path=dataset_path,
                release_name="complete",
                val_percentage=0.1,
                test_percentage=0.2,
            )

            split_types = ["random", "stratified"]
            partitions = ["test", "train", "val"]
            for split_type in split_types:
                for partition in partitions:
                    dataset_partition = dataset_class(
                        dataset_path=dataset_path,
                        release_name="complete",
                        partition=partition,
                        split_type=split_type,
                        split=split_values,
                    )
                    assert (
                        len(dataset_partition.annotations_path)
                        == expected_splits[f"{split_type}_{partition}"]
                    )
                    assert (
                        len(dataset_partition.images_path)
                        == expected_splits[f"{split_type}_{partition}"]
                    )

def test_split_and_load_classification_dataset_images(self):
    expected_splits = {
        "random_test": 40,
        "random_train": 140,
        "random_val": 20,
        "stratified_test": 40,
        "stratified_train": 140,
        "stratified_val": 20,
    }
    run_split_and_load_dataset_test(
        "classification-test",
        ClassificationDataset,
        expected_splits,
        "140_20_40"
    )

def test_split_and_load_instance_segmentation_dataset_images(self):
    expected_splits = {
        "random_test": 3,
        "random_train": 7,
        "random_val": 2,
        "stratified_test": 3,
        "stratified_train": 7,
        "stratified_val": 2,
    }
    run_split_and_load_dataset_test(
        "instance-segmentation-test",
        InstanceSegmentationDataset,
        expected_splits,
        "7_2_3"
    )

def test_split_and_load_object_detection_dataset_images(self):
    expected_splits = {
        "random_test": 11,
        "random_train": 37,
        "random_val": 6,
        "stratified_test": 11,
        "stratified_train": 37,
        "stratified_val": 6,
    }
    run_split_and_load_dataset_test(
        "object-detection-test",
        ObjectDetectionDataset,
        expected_splits,
        "37_6_11"
    )

```

umbertoDifa (changes requested):
The code looks good! Yet, many changes in one PR. I think this can be 4 PRs, to de-risk and ease the code review:
- Introduce Tests for Loading and Splitting Datasets
- Fix Bugs with Saving and Loading Dataset Releases
- Fix for Converting Darwin JSON 2.0 Polygons to YOLOv8 Format
- Fix for E2E COCO Conversion Test

darwin/dataset/download_manager.py
```diff
@@ -308,7 +310,10 @@ def _download_single_slot_from_json_annotation(
     generator = []
 
     if video_frames and slot.type != "image":
-        video_path: Path = parent_path / annotation_path.stem
+        if not use_folders:
```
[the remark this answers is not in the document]
JBWilkie: Agreed, this is better style in my opinion

tests/darwin/dataset/remote_dataset_test.py
```diff
@@ -398,7 +398,7 @@ def test_works_on_videos(
                             "height": 1080,
                             "thumbnail_url": "",
                             "source_files": [
-                                {"file_name": "test_video.mp4", "url": ""}
+                                {"file_name": "test_video/0000000.png", "url": ""}
```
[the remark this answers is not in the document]
JBWilkie: This is necessary due to the change in this PR to always read images from `source_files.file name` instead of `item.name`. In this test on the master branch, you'll see that `item.name` [already points to the PNGs](https://github.com/v7labs/darwin-py/blob/384f1efd3f01862ecf621ebe110d6f8e30f20cd0/tests/darwin/dataset/remote_dataset_test.py#L379) (split video frames), so given the change we also need to point source_files.file_name to the PNGs

JBWilkie:
This is being closed in favour of splitting the work up into smaller PRs (much of which is already done). This work includes breaking changes to our naming convention, so they will be released in a 1.0.0 version

PR #839, opened 2024-05-11
[DAR-1910][External] Fixed rare TypeError bug when importing COCO annotations

JBWilkie:
# Problem
When importing a directory of COCO annotation files, the importer returns a `List[List[AnnotationFile]]` type. This causes darwin-py to [try flattening the list](https://github.com/v7labs/darwin-py/blob/master/darwin/importer/importer.py#L155). If there are any `NoneType`s in `parsed_files` at this point, a `TypeError` will be thrown: `'NoneType' object is not iterable.`

# Solution
These `NoneType`s in `parsed_files` come from files that the importer was unable to parse. Therefore, remove all `NoneType`s from `parsed_files` prior. Note that we can't simply filter for `.json` files, since some annotation files are in the `.xml` format

# Changelog
Fixed rare bug when importing directories of COCO annotations


darwin/importer/importer.py
```diff
@@ -147,6 +147,7 @@ def maybe_console(*args: Union[str, int, float]) -> None:
     else:
         maybe_console("Using single CPU")
         parsed_files = list(map(importer, tqdm(files) if is_console else files))
+    parsed_files = [f for f in parsed_files if f is not None]
```
JBWilkie, line 150: At this point, `parsed_files` will contain a `None` for every file that the relevant importer was incapable of parsing. Remove it from the list before proceeding to avoid the issue targeted by this change

shernshiou (approved):

PR #844, opened 2024-05-21
[DAR-2235][External] Fix failing E2Es & allow them to be run locally

JBWilkie:
# Problems
- 1: Bug in YOLOV8 converter preventing it from working with Darwin JSON 2.0
- 2: COCO conversion E2E test failing due to mis-ordered categories in expected result
- 3: Small typing issue when reading annotation files from a directory in the exporter
- 4: E2E tests involving the `e2e_tests/data/base_annotation.json` file couldn't be run due to a `FileNotFoundError`

# Solutions
- 1: Allow YOLOV8 converter to work with Darwin JSON 2.0 by checking `paths` before `path`
- 2: Re-order the COCO categories in the expected result to fix the test
- 3: Fixed typing issue by ensuring a list of `Path` objects instead of `str`
- 4: Allow E2E tests to be run locally by ensuring we can always find the relevant file

# Changelog
- Fixed a YOLOV8 conversion bug
- Fixed incorrect test data for COCO conversion
- Fixed small typing issue in the exporter
- Change to allow E2Es to be fully run locally


darwin/importer/importer.py
```diff
@@ -1400,8 +1400,9 @@ def _overwrite_warning(
     """
     files_to_overwrite = []
     for local_file in local_files:
+        item_id = remote_files.get(local_file.full_path)[0]
```
JBWilkie, line 1403: `item_id` is not a requirement to import annotation files. Therefore, read it from the corresponding remote file, where we are guaranteed to find it

shernshiou (approved):

PR #855, opened 2024-06-05
[DAR-2228][External] - Fix for NifTI exports containing no polygons

JBWilkie:
# Problem
A change made in DAR-1165 accidentally introduced a bug whereby NifTI exports of files containing no polygons would fail. 

# Solution
- 1: Only attempt to export polygons if any polygon annotations are present
- 2: Added unit tests to make sure we're calling the polygon to volume & mask to volume functions when appropriate

# Changelog
Fixed issue that caused NifTI exports without polygons to fail


darwin/exporter/formats/nifti.py
```diff
@@ -350,7 +351,6 @@ def populate_output_volumes_from_polygons(
                 plane,
                 frame_idx,
             )
-    return volume
```
JBWilkie, line 353: This return statement actually isn't needed since we don't use `volume` anywhere. The function instead modifies the value of `output_volumes`

shernshiou: @JBWilkie Can you update the docs for this function too? Otherwise, this PR is good.

shernshiou (approved):

PR #856, opened 2024-06-05
[DAR-2161][External] Exponential backoff for external storage registration

JBWilkie:
# Problem
Recent changes to our external storage registration endpoints mean that they can return HTTP 429 if rate limits are exceeded. This can break registration through darwin-py as it was built assuming this would not be the case

# Solution
Introduce exponential backoff using the [tenacity](https://tenacity.readthedocs.io/en/latest/) library

# Changelog
Added exponential backoff for external storage registration


darwin/dataset/remote_dataset_v2.py
```diff
@@ -832,3 +856,18 @@ def register_multi_slotted(
                 print(f"  - {item}")
         print(f"Reistration complete. Check your items in the dataset: {self.slug}")
         return results
+
+    @retry(
+        wait=wait_exponential_jitter(initial=30, max=240),
+        stop=stop_after_attempt(5),
+        retry=retry_if_exception_type(requests.exceptions.RequestException),
+        before_sleep=before_sleep,
```
JBWilkie, line 864: This is just a 1st draft policy, I'll need feedback on the optimal min/max times as well as the criteria we stop after

tests/darwin/dataset/remote_dataset_test.py
```diff
@@ -1075,3 +1077,44 @@ def test_register_files_with_blocked_items(self, remote_dataset: RemoteDatasetV2
             {"item1": ["test.jpg"]},
             multi_slotted=True,
         )
+        assert len(result["registered"]) == 0
+        assert len(result["blocked"]) == 1
```
JBWilkie, line 1081: Small fix to a previously existing test that was missing asserts

shernshiou (approved):

darwin/dataset/remote_dataset_v2.py
```diff
@@ -832,3 +856,18 @@ def register_multi_slotted(
                 print(f"  - {item}")
         print(f"Reistration complete. Check your items in the dataset: {self.slug}")
         return results
+
+    @retry(
+        wait=wait_exponential_jitter(initial=60, max=120),
+        stop=stop_after_attempt(5),
+        retry=retry_if_exception_type(requests.exceptions.RequestException),
+        before_sleep=before_sleep,
+    )
```
JBWilkie, line 865: I'll need some feedback on this. Currently, this basically says:

If we hit a 429, exponentially retry with a minimum wait time of 30s, and a maximum wait time of 240s. Introduce some [jitter](https://aws.amazon.com/builders-library/timeouts-retries-and-backoff-with-jitter/) into the wait times, and fail after a maximum of 5 retries.

darwin/dataset/remote_dataset_v2.py
```diff
@@ -718,13 +732,18 @@ def register_single_slotted(
...
             print(f"Registering {len(chunk)} items...")
-            response = self.client.api_v2.register_items(payload, team_slug=self.team)
-            for item in json.loads(response.text)["items"]:
-                item_info = f"Item {item['name']} registered with item ID {item['id']}"
-                results["registered"].append(item_info)
-            for item in json.loads(response.text)["blocked_items"]:
-                item_info = f"Item {item['name']} was blocked for the reason: {item['slots'][0]['reason']}"
-                results["blocked"].append(item_info)
+            try:
+                response = self.register_items_with_retry(payload, team_slug=self.team)
```
balysv, line 736: It would be more conventional for this line should be left unchanged and instead have the retry logic implementation in the client code. Changes are more places would use that client (potentially clients use it directly in their scripts) and then they would benefit from the retry mechanis,

darwin/dataset/remote_dataset_v2.py
balysv: Great idea introducing some jitter :+1: For the values themselves, I would have a start of 60 seconds, ranging to 5 minutes with 10 retries. This may also not work ideally for all uses as it depends what concurrency this function is called.

darwin/dataset/remote_dataset_v2.py
```diff
@@ -832,3 +856,18 @@ def register_multi_slotted(
                 print(f"  - {item}")
         print(f"Reistration complete. Check your items in the dataset: {self.slug}")
         return results
+
+    @retry(
+        wait=wait_exponential_jitter(initial=60, max=120),
+        stop=stop_after_attempt(5),
+        retry=retry_if_exception_type(requests.exceptions.RequestException),
```
balysv, line 863: just to verify: isn't `RequestException` thrown by any other client condition than the one you coded?

darwin/dataset/remote_dataset_v2.py
```diff
@@ -47,6 +48,19 @@
...
+from tenacity import (
+    RetryCallState,
+    retry,
+    retry_if_exception_type,
+    stop_after_attempt,
+    wait_exponential_jitter,
+)
+
+
+def before_sleep(retry_state: RetryCallState):
```
balysv, line 60: would call this something more exact to what it does: `log_rate_limit_exceeded` or similar

balysv (changes requested):

darwin/dataset/remote_dataset_v2.py
JBWilkie: Great point, have made this change

darwin/backend_v2.py
```diff
@@ -238,6 +255,12 @@ def import_annotation(
             f"v2/teams/{team_slug}/items/{item_id}/import", payload=payload
         )
 
+    @retry(
+        wait=wait_exponential_jitter(initial=60, max=300),
+        stop=stop_after_attempt(10),
+        retry=retry_if_status_code_429,
+        before_sleep=log_rate_limit_exceeded,
+    )
     @inject_default_team_slug
```
balysv, line 264: any chance we can get a test for this seeing the retries triggered? You probably need to mock the api responses returning `429`s

JBWilkie: Done

balysv (approved):

PR #859, opened 2024-06-06
[DAR-2487][External] Fixed long-standing ruff linter issues

JBWilkie:
# Problem
We use [ruff](https://docs.astral.sh/ruff/) to lint darwin-py. There've been some long standing non-functional errors highlighted by the linter

# Solution
Resolve the errors worth resolving, and suppress those that are not 

# Changelog
Fixed or suppressed remaining darwin-py errors highlighted by the ruff linter

### DIFF
--- a/darwin/exporter/formats/mask.py
+++ b/darwin/exporter/formats/mask.py
@@ -10,7 +10,7 @@ import numpy as np
 try:
     from numpy.typing import NDArray
 except ImportError:
-    NDArray = Any  # type:ignore
+    NDArray = Any  # type:ignore # noqa F821
 from PIL import Image
 from upolygon import draw_polygon
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,9 +51,9 @@ warn_required_dynamic_aliases = true
 warn_untyped_fields = true
 
 [tool.ruff]
-ignore = ["E203", "E402", "E501"]
+lint.ignore = ["E203", "E402", "E501", "C901"]
 line-length = 88
-select = ["E", "F", "C"]
+lint.select = ["E", "F", "C"]
 
 [tool.flake8]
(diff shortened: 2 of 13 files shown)


darwin/exporter/formats/mask.py
```diff
@@ -10,7 +10,7 @@
 try:
     from numpy.typing import NDArray
 except ImportError:
-    NDArray = Any  # type:ignore
+    NDArray = Any  # type:ignore # noqa F821
```
JBWilkie, line 13: We need this `NDArray = Any` to support darwin-py working with Databricks, which relies on a very old `NumPy` version

pyproject.toml
```diff
@@ -51,9 +51,9 @@ warn_required_dynamic_aliases = true
 warn_untyped_fields = true
 
 [tool.ruff]
-ignore = ["E203", "E402", "E501"]
+lint.ignore = ["E203", "E402", "E501", "C901"]
```
JBWilkie, line 54: [C901](https://docs.astral.sh/ruff/rules/complex-structure/) highlights all functions with a McCabe complexity > n. Given how far darwin-py is into it's development, I don't think we should consider this. It might be worth highlighting for `darwin.future`

tests/darwin/exporter/formats/export_mask_test.py
```diff
@@ -13,7 +13,7 @@
 try:
     from numpy.typing import NDArray
 except ImportError:
-    NDArray = Any  # type:ignore
+    NDArray = Any  # type:ignore # noqa F821
```
JBWilkie, line 16: [As above](https://github.com/v7labs/darwin-py/pull/859/files#r1629782306)

shernshiou (approved):
This is a good chance

PR #863, opened 2024-06-10
[DAR-2504][External] Generate empty NifTI volumes when no polygons are present

JBWilkie:
# Problem
Currently, if no polygons are present in a NifTI file at the point of export, no file will be generated. This is problematic for our backend exports which seem to hang in these scenarios

# Solution
If there are no polygons, export an empty NifTI file

# Changelog
Generate an empty NifTI file if no polygons are present for export


darwin/exporter/formats/nifti.py
```diff
@@ -160,6 +160,8 @@ def build_output_volumes(
         )
         # Builds output volumes per class
         volume_dims, pixdims, affine, original_affine = process_metadata(slot.metadata)
+        if not class_names_to_export:
+            class_names_to_export = ["empty"]
```
JBWilkie, line 164: This basically says: If there are no polygons for export, create an empty NifTI volume as though there was a polygon with class name `empty`

I'm not very happy with just referring to this as `empty`, so open for suggestions!

This also means that if a user exports NifTI files with mask annotations but no polygons, there will be two files output. This doesn't seem ideal. Should we first check if there are raster annotations and skip this if they are present?

darwin/importer/formats/nifti.py
```diff
@@ -16,8 +16,8 @@
     from scipy.ndimage import zoom
 except ImportError:
     import_fail_string = """
-    You must install darwin-py with pip install nibabel connected-components-3d
-    in order to import with using nifti format
+    You must install darwin-py with pip install darwin-py\[medical]
+    in order to export using nifti format
```
JBWilkie, line 20: This came up recently from a user - It's better to let the optional `[medical]` do the work instead

shernshiou (approved):

PR #864, opened 2024-06-12
feat: Implement no isotropic annotation

shernshiou:
# Problem
(describe the problem here)

# Solution

# Changelog
(Will appear in release docs)


shernshiou:
Setting this aside awaiting further discussion with @JBWilkie

JBWilkie:
> Setting this aside awaiting further discussion with @JBWilkie

@shernshiou Since this piece of work is turning out to be a bit more complex than first anticipated, would you like me to review & test it now, or wait until we've progressed it a bit more?

JBWilkie (approved):
Reviewed approach with Shern in a call

PR #868, opened 2024-06-17
[DAR-2246][External] Unify darwin-py naming convention

JBWilkie:
# Problem
Depending on the type of data and the method used, darwin-py does not adopt a unified approach for referring to local files when pulling export releases and loading files. This has historically:
- Been a significant source of client and employee confusion
- Resulted in a number of bugs

# Solution
Unify the logic based on internal discussions. See [this table in Notion](https://www.notion.so/v7labs/Darwin-py-file-naming-convention-18f43cbdf61443a5a408bc978b2bc513?pvs=4#6532ff354e734f6a91b4fe8a9d108594) for more detail. Changes are:
- 1: Make `use_folders` the default behaviour when pulling datasets
- 2: Standardise the use of item names: The `item_name` will be used as a default to save and read files, with three notable exceptions:
    - Downloading video files as video frames: Each video will result in a folder named after the item name, containing individual frames named after the frame index
    - Downloading multi-slotted items: Multi-slotted items have one item name but potentially many source names. This will result in a folder named after the item name, containing one file for each slot named after the source name
    - Downloading a slotted item consisting of multiple concatenated DICOM slices: These items can contain multiple source files in a single slot. Each item will result in a folder named after the item name, containing one file for each slice named after the source name

There is also a small bug fix for correctly skipping files when pulling a release if the release is pulled as a flat structure (covered in a comment below)

# Changelog
- Updated the way darwin-py names files when saving them in some scenarios
- Added tests against these changes
- Small fix for correctly skipping files when pulling a release if the release is pulled as a flat structure


darwin/dataset/download_manager.py
```diff
@@ -95,11 +95,13 @@ def download_all_images_from_annotations(
...
-                / Path(annotation.remote_path.lstrip("/\\")).resolve().absolute()
-                / Path(annotation.filename)
-            )
+            filename = annotation.filename
+            if use_folders and annotation.remote_path != "/":
+                planned_image_path = (
+                    images_path / Path(annotation.remote_path.lstrip("/\\"))
+                ) / filename
+            else:
+                planned_image_path = images_path / filename
```
JBWilkie, line 104: This allows `planned_image_path` to work with dataset releases pulled in a flat structure

darwin/dataset/download_manager.py
```diff
@@ -95,11 +95,13 @@ def download_all_images_from_annotations(
 
         if not force_replace:
             # Check the planned path for the image against the existing images
-            planned_image_path = (
-                images_path
-                / Path(annotation.remote_path.lstrip("/\\")).resolve().absolute()
-                / Path(annotation.filename)
-            )
+            filename = annotation.filename
```
dorfmanrobert, line 98: Is there a reason it was `Path(annotation.filename)` before?

JBWilkie: Good spot - I don't believe so. `annotation.filename` gives type `str` but it can still be used to construct `Path` objects in this way

However, there is no harm in making it type `Path`, so I have kept this behaviour

tests/darwin/dataset/remote_dataset_test.py
```diff
@@ -806,6 +809,170 @@ def fake_download_zip(self, path):
...
+        download_funcs = self._test_pull_naming_convention(
+            file_name,
+            use_folders=False,
+            video_frames=False,
+            force_slots=True,
+        )
+        assert download_funcs[0].args[2] == expected_paths[0]
+        assert download_funcs[1].args[2] == expected_paths[1]
+        assert download_funcs[2].args[2] == expected_paths[2]
+        assert 1 == 1
```
dorfmanrobert, line 893: can delete this

tests/darwin/dataset/remote_dataset_test.py
dorfmanrobert, line 1: just to double check, do these tests cover both short and long videos?

JBWilkie: Added 2 tests to cover long videos & long videos pulled as frames

dorfmanrobert (approved):
nice!

PR #872, opened 2024-06-19
[DAR-2639][External] Add data & tests for in-platform model training

JBWilkie:
# Problem
There are no tests for in-platform model training functionality

# Solution
Add data & tests to support testing of darwin-py functionality used in in-platform model training. In addition to this:
- 1: Re-add a function `convert_xyxy_to_bounding_box` used by model training
- 2: Force `get_annotations()` to read local items from the item name, not the annotation file name

# Changelog
(Will appear in release docs)


darwin/dataset/download_manager.py
```diff
@@ -339,7 +344,7 @@ def _download_single_slot_from_json_annotation(
             image_url = image["url"]
             image_filename = image["file_name"]
             suffix = Path(image_filename).suffix
-            stem = annotation_path.stem
+            stem = Path(annotation.filename).stem
```
JBWilkie, line 347: In cases where there are duplicate item names, the `annotation_path.stem` will have an `_n` appended, which will cause darwin-py to fail to find the corresponding item. Therefore, use the item name since this will not have a `_n` appended for uniqueness

darwin/dataset/utils.py
```diff
@@ -505,55 +505,57 @@ def _validate_inputs(
...
+            Filepaths for the dataset.
 
     Raises:
         ValueError: If the split_type is invalid.
         FileNotFoundError: If the dataset partition file is not found.
     """
-    if split_type is None:
-        split_file = f"{partition}.txt"
-    elif split_type == "random":
+    if split_type == "random":
```
dorfmanrobert, line 555: should we add a default if we don't handle None?/why was this there before?

JBWilkie: Before the code gets to this stage, it passes the `split_type` into `_validate_inputs()` above. This function used to accept `split_type` as `None`, but since the top-level function `get_annotations()` sets `split_type` to `random` in the definition, it will never be `None`

dorfmanrobert (approved):
resolves model training bugs, phew!

PR #876, opened 2024-06-28
[DAR-2707][External] Allow repeated polling of pending export releases

JBWilkie:
# Problem
Before a dataset release can be pulled, it needs to finish generating. The time taken for this can vary based on export size and current load on the export pipeline. If a release isn't ready for pulling, then darwin-py will throw an error

# Solution
Introduce the optional `retry` parameter (SDK & CLI) that allows polling of pending dataset releases. If the pending release becomes available within the allotted time, it will be automatically downloaded

# Changelog
Allow optional polling of pending dataset releases in case the release is not yet ready for download


darwin/cli_functions.py
```diff
@@ -428,8 +429,14 @@ def pull_dataset(
         Pulls video frames images instead of video files. Defaults to False.
     force_slots: bool
         Pulls all slots of items into deeper file structure ({prefix}/{item_name}/{slot_name}/{file_name})
+    retry: bool
+        If True, will repeatedly try to download the release if it is still processing up to a maximum of 5 minutes.
     """
     version: str = DatasetIdentifier.parse(dataset_slug).version or "latest"
+    if version == "latest" and retry:
+        raise ValueError(
+            "To retry downloading a release, a release name must be provided. This can be done as follows:\n\ndarwin dataset pull team-slug/dataset-slug:release-name"
```
JBWilkie, line 438: This is necessary because even if a pending release **is** the latest release, the API response will be `False` until it is available for download. Therefore if passing `retry` as `True` without specifying the release name, we have no way to know which release is actually being referred to

darwin/dataset/remote_dataset.py
```diff
@@ -251,16 +255,46 @@ def pull(
...
+            If the release is still processing after the maximum retry duration.
         """
 
         console = self.console or Console()
 
         if release is None:
-            release = self.get_release()
+            if retry:
+                raise ValueError(
+                    "To retry downloading a release, a release name must be provided. This can be done as follows:\n\nrelease = dataset.get_release(name='release_name')\ndataset.pull(release=release, retry=True)"
```
JBWilkie, line 267: [As above](https://github.com/v7labs/darwin-py/pull/876/files#r1658843008), with an alternative error message tailored to the SDK method of pulling

darwin/dataset/remote_dataset.py
```diff
@@ -251,16 +255,46 @@ def pull(
...
+                release = self.get_release(retry=retry)
 
         if release.format != "json" and release.format != "darwin_json_2":
             raise UnsupportedExportFormat(release.format)
 
+        if release.status == "pending":
+            if retry:
+                retry_duration = 300
+                retry_interval = 10
+                while release.status == "pending" and retry_duration > 0:
```
JBWilkie, line 279: Poll every 10 seconds, up to a maximum of 5 minutes. I'm open to feedback on what the duration of `retry_duration` should be!

darwin/dataset/remote_dataset.py
```diff
@@ -744,9 +785,13 @@ def get_release(self, name: str = "latest") -> "Release":
...
         """
-        releases = self.get_releases()
+        releases = self.get_releases(retry)
         if not releases:
-            raise NotFound(str(self.identifier))
+            raise NotFound(
+                str(
+                    f"No releases found for dataset '{self.name}'. Please create an export of this dataset first."
+                )
+            )
```
JBWilkie, line 794: This is a better, more specific error message

darwin/dataset/remote_dataset.py
```diff
@@ -715,24 +749,31 @@ def get_report(self, granularity: str = "day") -> str:
...
+            If True, return  all releases, including those that are not available.
+
         Returns
         -------
         List["Release"]
             Returns a sorted list of available ``Release``\\s with the most recent first.
         """
 
-    def get_release(self, name: str = "latest") -> "Release":
+    def get_release(self, name: str = "latest", retry: bool = True) -> "Release":
```
JBWilkie, line 767: This has to be `True` by default. Otherwise it cannot return releases that are not available for download. We need this to be able to display [this](https://github.com/v7labs/darwin-py/pull/876/files#diff-200952219fbbebac3a477ed4f073d173489e125467b83f5e15ef67105535e274R291-R292) message for pending releases

shernshiou (approved):
I do not detect any problem codewise. Haven't tested it tho. 👍🏼

darwin/dataset/remote_dataset.py
```diff
@@ -715,24 +749,31 @@ def get_report(self, granularity: str = "day") -> str:
...
         """
         Get a specific ``Release`` for this ``RemoteDataset``.
 
         Parameters
         ----------
         name : str, default: "latest"
             Name of the export.
+        retry : bool, default: True
```
balysv, line 775: the name of the argument `retry` doesn't match what it's doing "return all releases, if true". It should be something like `incude_pending` or similar that is a bit more self explanataory

darwin/dataset/remote_dataset_v2.py
```diff
@@ -135,11 +140,19 @@ def get_releases(self) -> List["Release"]:
...
+
+        if retry:
+            return sorted(
+                releases,
+                key=lambda x: x.version,
+                reverse=True,
+            )
+        else:
+            return sorted(
+                filter(lambda x: x.available, releases),
```
balysv, line 152: I'm sure you can have this nicer where only the first argument `releases` or `filter(lambda x: x.available, releases)` is chosen with an `if` and have a single return line:

```
return sorted(
                releases_fn,
                key=lambda x: x.version,
                reverse=True,
            )
```

darwin/cli_functions.py
```diff
@@ -428,8 +429,14 @@ def pull_dataset(
         Pulls video frames images instead of video files. Defaults to False.
     force_slots: bool
         Pulls all slots of items into deeper file structure ({prefix}/{item_name}/{slot_name}/{file_name})
+    retry: bool
+        If True, will repeatedly try to download the release if it is still processing up to a maximum of 5 minutes.
     """
     version: str = DatasetIdentifier.parse(dataset_slug).version or "latest"
+    if version == "latest" and retry:
```
balysv, line 436: Few questions here:
- I don't recall the details but is the name `latest` hardcoded by us? 
- What happens if a client deliberately passes the name `latest` with `retry=True`? 
- I don't think this restriction is necessary, can't we pick the name of the latest release ourselves before performing the download and then do the retry logic using it instead of `latest`. This would ensure we refer to the same export even if a new export would be created in the meantime

darwin/dataset/release.py
```diff
@@ -22,6 +23,8 @@ class Release:
         The version of the ``Release``.
     name : str
         The name of the ``Release``.
+    status : str
```
balysv, line 26: Unsure if it's common in darwin-py but this would  be better as an `enum` as it only has a select few values.

tests/darwin/dataset/release_test.py
```diff
@@ -16,6 +16,7 @@ def release(dataset_slug: str, team_slug_darwin_json_v2: str) -> Release:
         team_slug=team_slug_darwin_json_v2,
         version="latest",
         name="test",
+        status="test_status",
```
balysv, line 19: For documentation purposes, it'd be best to use actual values of export statuses here instead of stubs as they are enums and not arbitrary strings.

balysv (changes requested):

darwin/dataset/remote_dataset.py
```diff
@@ -251,16 +255,46 @@ def pull(
...
+            else:
+                release = self.get_release(retry=retry)
 
         if release.format != "json" and release.format != "darwin_json_2":
             raise UnsupportedExportFormat(release.format)
 
+        if release.status == "pending":
+            if retry:
+                retry_duration = 300
+                retry_interval = 10
```
balysv, line 278: It would be more conventional to have these configurable via CLI or some SDK settings

darwin/cli_functions.py
JBWilkie: > I don't recall the details but is the name latest hardcoded by us?

Yes, `latest` is a reserved release name. If you try to create an export named `latest`, the api responds with `{"errors":{"name":["is reserved"]}}`

> What happens if a client deliberately passes the name `latest` with `retry=True`?

[We will return the latest available release](https://github.com/v7labs/darwin-py/blob/DAR-2707/darwin/dataset/remote_dataset.py#L799). Since it is never possible to create an export with the name `latest`, should we disallow this when using `get_release()` to prevent confusion?

>I don't think this restriction is necessary, can't we pick the name of the latest release ourselves before performing the download and then do the retry logic using it instead of latest. This would ensure we refer to the same export even if a new export would be created in the meantime

Actually yes, I think we can. This is because each release has a `export_date` of type `datetime.datetime`. This allows us to select the most recent release incase `retry` is passed as `True`. I'll make this change now

darwin/dataset/remote_dataset.py
JBWilkie: @balysv This makes sense. I can see 2 options, both of which involve building in some validation:
- 1: Make these values configurable in the `~/.config.yaml` file, or:
- 2: Add two additional arguments: `retry_duration` and `retry_interval` with default values of ~10 minutes & ~10 seconds. These can be configured, but if they're passed without `retry=True` then we will throw an error

I'm leaning toward the additional arguments

balysv (approved):

PR #880, opened 2024-07-01
[DAR-2045][External] Multi-threaded annotation imports

JBWilkie:
# Problem
Importing annotations only works in a single-threaded fashion. This results in a significant bottleneck when sending import API requests in series

# Solution
Add optional multi-threading to significantly improve speed when importing annotations. Pass the `use_multi_cpu` flag as `True` to enable multi-threading

# Changelog
Added option to import annotations with multi-threading, resulting in significant speed gains

### DIFF
--- a/darwin/cli_functions.py
+++ b/darwin/cli_functions.py
@@ -899,6 +899,9 @@ def dataset_import(
             dataset_identifier=dataset_slug
         )
 
+        if cpu_limit is not None:
+            use_multi_cpu = True
+
         import_annotations(
             dataset,
             importer,
--- a/darwin/importer/importer.py
+++ b/darwin/importer/importer.py
@@ -928,7 +928,7 @@ def import_annotations(  # noqa: C901
                     for _ in tqdm(
                         concurrent.futures.as_completed(futures),
                         total=len(futures),
-                        desc="Importing files",
+                        desc="Importing annotations from local file",
                     ):
                         future = next(concurrent.futures.as_completed(futures))
                         try:
@@ -938,7 +938,9 @@ def import_annotations(  # noqa: C901
                                 f"Generated an exception: {exc}", style="error"
                             )
             else:
-                for file in tqdm(files_to_track, desc="Importing files"):
+                for file in tqdm(
+                    files_to_track, desc="Importing annotations from local file"
+                ):
                     import_annotation(file)
 
     if use_multi_cpu:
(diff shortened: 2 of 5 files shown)


darwin/cli_functions.py
```diff
@@ -899,6 +899,9 @@ def dataset_import(
             dataset_identifier=dataset_slug
         )
 
+        if cpu_limit is not None:
+            use_multi_cpu = True
+
```
JBWilkie, line 904: Without this change, the `--cpu_limt` CLI flag does nothing. This allows multi-threading to work for annotation imports initiated by the CLI

darwin/importer/importer.py
```diff
@@ -952,36 +973,55 @@ def import_annotations(  # noqa: C901
...
-                    for error in errors:
-                        console.print(f"\t{error}", style="error")
+            if use_multi_cpu:
+                with concurrent.futures.ThreadPoolExecutor(
+                    max_workers=cpu_limit
+                ) as executor:
+                    futures = [
+                        executor.submit(import_annotation, file)
+                        for file in files_to_track
+                    ]
```
JBWilkie, line 984: There are 2 multi-threading loops (1 inside the other). Both are necessary. This is the inner loop, and it parallelises the import of each `import_annotation` object

This is useful for annotation formats where 1 JSON file can represent multiple dataset items, such as COCO, where 1 file contains the entire dataset. In these cases, you could have a large number of `import_annotation` objects in `files_to_track`

darwin/importer/importer.py
```diff
@@ -952,36 +973,55 @@ def import_annotations(  # noqa: C901
...
+                    files_to_track, desc="Importing annotations from local file"
+                ):
+                    import_annotation(file)
+
+    if use_multi_cpu:
+        with concurrent.futures.ThreadPoolExecutor(max_workers=cpu_limit) as executor:
+            futures = [
+                executor.submit(process_local_file, local_path)
+                for local_path in {local_file.path for local_file in local_files}
+            ]
```
JBWilkie, line 1008: There are 2 multi-threading loops (1 inside the other). Both are necessary. This is the outer loop, and it parallelises the import of each local annotation file for import.

This is useful for annotation formats where there is a 1-to-1 mapping between annotation file and dataset item, such as Darwin JSON 2.0

shernshiou (approved):

PR #886, opened 2024-07-16
[DAR-2996][External] Fixed `remove_extra` logic and added removal of empty directories

JBWilkie:
# Problem
The optional `remove_extra` feature has been in a broken state for a while due to several changes so file naming conventions in the last year

# Solution
Adjusted the behaviour of the optional `remove_extra` `pull()` feature so that it removes any files and directories that **would not** be overwritten by the release currently being downloaded, essentially acting as a wipe of the entire images directory. However, it cannot be implemented this way in case there are difficulties downloading any file in the current release (that would result in missing files). Therefore, we:
- 1: Build a list of dataset file paths that would result from downloading the release
- 2: Remove any existing file that is not present in the above list
- 3: Since the above can result in empty local dataset directories, scan and remove empty directories from the `images` dataset directory

# Changelog
Fixed logic of the `remove_extra` feature when pulling dataset releases


darwin/dataset/download_manager.py
```diff
@@ -82,7 +82,7 @@ def download_all_images_from_annotations(
         raise ValueError(f"Annotation format {annotation_format} not supported")
 
     # Verify that there is not already image in the images folder
-    unfiltered_files = images_path.rglob("*") if use_folders else images_path.glob("*")
+    unfiltered_files = images_path.rglob("*") if use_folders else images_path.rglob("*")
```
JBWilkie, line 85: Users may download flat releases of a dataset **after** downloading releases with the folder structure

darwin/dataset/download_manager.py
```diff
@@ -82,9 +82,10 @@ def download_all_images_from_annotations(
         raise ValueError(f"Annotation format {annotation_format} not supported")
 
     # Verify that there is not already image in the images folder
-    unfiltered_files = images_path.rglob("*") if use_folders else images_path.glob("*")
     existing_images = {
-        image for image in unfiltered_files if is_image_extension_allowed(image.suffix)
+        image
+        for image in images_path.rglob("*")
+        if is_image_extension_allowed(image.suffix)
```
JBWilkie, line 88: We want `rglob()` in both cases, because users may download flat releases of a dataset after downloading releases with the folder structure

darwin/dataset/download_manager.py
```diff
@@ -82,9 +83,10 @@ def download_all_images_from_annotations(
         raise ValueError(f"Annotation format {annotation_format} not supported")
 
     # Verify that there is not already image in the images folder
-    unfiltered_files = images_path.rglob("*") if use_folders else images_path.glob("*")
     existing_images = {
-        image for image in unfiltered_files if is_image_extension_allowed(image.suffix)
+        image
+        for image in images_path.rglob("*")
+        if is_file_extension_allowed(image.name)
```
JBWilkie, line 89: We want `rglob()` in both cases, because users may download flat releases of a dataset after downloading releases with a folder structure

darwin/dataset/download_manager.py
```diff
@@ -114,17 +116,25 @@ def download_all_images_from_annotations(
...
+            for annotation_path in annotations_path.glob(f"*.{annotation_format}")
+        )
+        release_image_paths = [
+            _get_planned_image_paths(annotation, images_path, use_folders)
+            for annotation in annotations
         ]
+        if any(isinstance(i, list) for i in release_image_paths):
+            release_image_paths = [
+                item for sublist in release_image_paths for item in sublist
+            ]
```
JBWilkie, line 130: Nested list for slotted items, so flatten if needed

darwin/dataset/download_manager.py
```diff
@@ -343,8 +353,12 @@ def _download_single_slot_from_json_annotation(
...
             image_url = image["url"]
             image_filename = image["file_name"]
-            suffix = Path(image_filename).suffix
-            stem = Path(annotation.filename).stem
+            if image_filename.endswith(".nii.gz"):
+                suffix = ".nii.gz"
+                stem = image_filename[: -len(suffix)]
+            else:
+                suffix = Path(image_filename).suffix
+                stem = Path(annotation.filename).stem
```
JBWilkie, line 361: Small extra bug-fix when downloading single-slotted NifTI files (`.nii.gz`). Because of the way we handle suffixes when downloading single-slotted items, they are currentyl named `.gz`. NifTI files are a special case of being the only file type we support with 2 extensions. Therefore, we simple handle this case here

This is not a problem for the multi-slotted case, because there we take the entire source name as the filename (which includes `.nii.gz`)

darwin/dataset/download_manager.py
```diff
@@ -343,8 +353,12 @@ def _download_single_slot_from_json_annotation(
...
             image_url = image["url"]
             image_filename = image["file_name"]
-            suffix = Path(image_filename).suffix
-            stem = Path(annotation.filename).stem
+            if image_filename.endswith(".nii.gz"):
+                suffix = ".nii.gz"
+                stem = annotation.filename[: -len(suffix)]
+            else:
+                suffix = Path(image_filename).suffix
+                stem = Path(annotation.filename).stem
```
JBWilkie, line 361: Small extra bug-fix when downloading single-slotted NifTI files (.nii.gz) I noticed when working on this. Because of the way we handle suffixes when downloading single-slotted items, they are currentyl named .gz. NifTI files are a special case of being the only file type we support with 2 extensions. Therefore, we simply handle this case here

This is not a problem for the multi-slotted case, because there we take the entire source name as the filename (which includes .nii.gz)

darwin/dataset/download_manager.py
```diff
@@ -95,14 +97,14 @@ def download_all_images_from_annotations(
...
-                planned_image_path = images_path / filename
-            if planned_image_path in existing_images:
+            planned_image_paths = _get_planned_image_paths(
+                annotation, images_path, use_folders
+            )
+
+            if all(
+                planned_image_path in existing_images
+                for planned_image_path in planned_image_paths
+            ):
```
JBWilkie, line 107: We want `if all` here because we should re-download a multi-slotted file if any local source file is missing

darwin/utils/utils.py
```diff
@@ -110,21 +110,21 @@ def is_image_extension_allowed_by_filename(filename: str) -> bool:
...
+    filename : str
+        The name of the file.
 
     Returns
     -------
     bool
         Whether or not the given extension is allowed.
     """
-    return extension.lower() in SUPPORTED_IMAGE_EXTENSIONS
+    return any(filename.lower().endswith(ext) for ext in SUPPORTED_EXTENSIONS)
```
JBWilkie, line 127: Necessary to check for `.nii.gz`, which we support. Without this, we an only check the last suffix which will be `.gz`

shernshiou (approved):

PR #887, opened 2024-07-17
[DAR-2991][External] Set CLI to pull with folders by default & display overwrite warning

JBWilkie:
# Problem
In [DAR-2246](https://github.com/v7labs/darwin-py/pull/868), we made significant breaking changes to the way darwin-py names files when downloading them. One of those changes was changing the default behaviour of `pull()` so that it replicates remote folder structures locally by default. However, it was overlooked that CLI-initiated pulls had their own default behaviour, and this was not changed in that PR

This, combined with the naming changes themselves, has meant that CLI-initiated dataset pulls will result in overwritten local files **if** there are identically named files in different folders in the release

# Solution
- Change the default behaviour of `pull()` from the CLI to locally replicate remote folder structures. To allow users to still pull flat releases with the CLI, added a `--no-folders` arg
- Implemented logic to detect duplicate local file paths if pulling without folders. If duplicates are detected, display a warning

The warning is in the following style:
```
Warning: Identical filenames detected in your export release. 

You are pulling a flat release with identically named dataset items. The release will still be pulled, but to prevent overwriting your dataset files, please re-pull the release with the folder structure. This can be done as follows:
- CLI: darwin dataset pull team_slug/dataset_slug --folders
- SDK: dataset.pull(use_folders=True)

The following paths are duplicated:
- {path_1} is duplicated 2 times
- {path_2} is duplicated 2 times
```

# Changelog
- Dataset releases pulled with the CLI will now locally replicate the remote folder structure by default
- If pulling a dataset release without folders, display a warning if there are any local files that will be overwritten


darwin/options.py
```diff
@@ -528,6 +532,9 @@ def parse_args(self) -> Tuple[Namespace, ArgumentParser]:
         """
         args = self.parser.parse_args()
 
+        if args.action == "pull" and args.folders is not True:
+            args.folders = args.folders.lower() == "true"
+
```
dorfmanrobert, line 537: does this have any problems since `args.folders` is a bool?

darwin/dataset/download_manager.py
```diff
@@ -708,3 +712,39 @@ def _remove_empty_directories(images_path: Path) -> bool:
...
+) -> None:
+    """
+    If pulling a release without folders, check for duplicate filepaths in the download functions.
+    This can arise when pulling a flat release with identically named dataset items in different folders.
+
+    If duplicates are found, display a warning message but do not block the download.
+
+    Parameters
+    ----------
+    download_functions : List[Callable[[], None]]
```
dorfmanrobert, line 728: could maybe add description or remove as type is in the fn def

darwin/dataset/download_manager.py
```diff
@@ -708,3 +712,39 @@ def _remove_empty_directories(images_path: Path) -> bool:
...
+    """
+    image_paths_to_download = [
+        download_function.args[2] for download_function in download_functions
+    ]
+
+    path_counts = Counter(image_paths_to_download)
```
dorfmanrobert, line 734: why is instantiating this Counter needed?

JBWilkie: It creates `path_counts`, which is a count of the occurrences of each unique path. We use it below to detect duplicate paths (paths with a count > 1)

darwin/dataset/download_manager.py
JBWilkie: Added description

darwin/options.py
JBWilkie: Apologies, I should have added a comment here before. This isn't clean but it's necessary to preserve the optional nature of the `--folders` flag. Currently, `--folders` behaves like this:
- If not passed, the param is `False` (pull flat structure)
- If passed, the param is `True` (pull folder structure)
- It does not accept any passed value

To preserve as much existing API behaviour as possible, the goal is this PR is the following:
- If not passed, the param is `True` (pull folder structure)
- If passed, the param is `True` (pull folder structure)
- If passed with a value, unless that value is `True`, set the param to `False` (Essentially still allowing CLI-based pulls to be pulled as flat structures)

JBWilkie: Re-writing to employ the use of a `--no-folders` arg

JBWilkie:
> nice, looks good! just checking, should there now be a ticket to eventually deprecate `--folders` since it doesn't actually do anything?

Yep, created DAR-3189 for it

PR #895, opened 2024-08-01
[DAR-2981][External] Catch & handle blocking / non-blocking errors when importing annotations to slots

JBWilkie:
# Problem
When importing annotations to multi-slotted or multi-channeled items, there are a few ways users can go wrong that we want to start handling:

**Multi-slotted items:**
All annotations that are not linked to a slot are imported to the default slot, which is the 1st slot listed under `slots` in the [List Items](https://docs.v7labs.com/reference/items-list) API response. There is no feedback that this is happening

**Multi-channeled items:**
All annotations that are not linked to a slot are also imported to the default slot, which is the same as above. However, multi-channeled items require that all of their annotations exist in the base slot, which is not necessarily the same as the default slot. The base slot is the first slot listed under `layout.slots_grid` in the [List Items](https://docs.v7labs.com/reference/items-list) API response. If annotations are uploaded to a non-base slot, they will not be visible in the workview

# Solution
To assist users when they run into these situations, we introduce:
- 1: A non-blocking warning for annotations imported to multi-slotted items that don't specify the slot. The warning explains that the annotation will be uploaded to the default slot
- 2: A non-blocking warning for annotations imported to multi-channeled items that don't specify the slot. We import it to the base slot & provide feedback on this in the warning, as well as why this is necessary
- 3: A blocking error for annotations imported to multi-channeled items that specify the non-base slot. We do not import these items, and explain in the warning that these annotations need to be pointed to the base slot and re-imported

Warnings and errors accumulate during the import process, and are displayed afterwards. All files except those that generate at least one error are imported. A sample warning looks like this:
```
WARNING: Although they imported successfully, the following 2 file(s) had the following non-blocking warnings:
- File: /Users/john/Desktop/import/cars-backwards-slots.json, warnings:
  - Annotation imported to multi-slotted item /dir1/cars-backwards-slots not assigned slot. Uploading to the default slot: 1
  - Annotation imported to multi-slotted item /dir1/cars-backwards-slots not assigned slot. Uploading to the default slot: 1
- File: /Users/john/Desktop/import/cars-channels_2.json, warnings:
  - Annotation imported to multi-channeled item /dir2/cars-channels not assigned a slot. Uploading to the base slot: 3
  - Annotation imported to multi-channeled item /dir2/cars-channels not assigned a slot. Uploading to the base slot: 3
WARNING: 1 file(s) have the following blocking issues and could not be imported. Please resolve these issues and re-import them.
- File: /Users/john/Desktop/import/cars-channels.json, errors:
  - Annotation linked to slot 4 of the multi-channeled item /dir2/cars-channels. Annotations uploaded to multi-channeled items have to be uploaded to the base slot, 
which for this item is 3.
  - Annotation linked to slot 1 of the multi-channeled item /dir2/cars-channels. Annotations uploaded to multi-channeled items have to be uploaded to the base slot, 
which for this item is 3.
```
Finally, if users attempt to import annotations to multi-slotted or multi-channeled items with a format other than Darwin JSON 2.0, we raise a `TypeError`. This is because only Darwin JSON 2.0 supports multi-slotted & multi-channeled items. For example:
```
raise TypeError(
TypeError: You are attempting to import annotations to multi-slotted or multi-channeled items using an annotation format (coco) that doesn't support them. To import annotations to multi-slotted or multi-channeled items, please use the Darwin JSON 2.0 format: https://docs.v7labs.com/reference/darwin-json
```

# Changelog
Introduced non-blocking warnings & blocking errors when importing annotations to multi-slotted and multi-channeled items. If errors or warnings are encountered, we provide feedback on how to resolve the errors


darwin/importer/importer.py
```diff
@@ -824,6 +848,11 @@ def import_annotations(  # noqa: C901
         else:
             local_files.append(parsed_file)
 
+    annotation_format = _get_annotation_format(importer)
+    _check_annotation_format_slot_compatibility(
+        local_files, remote_files, annotation_format
+    )
+
```
JBWilkie, line 855: It's best to check early on if the user is trying to import non-Darwin JSON 2.0 annotations to multi-slotted/multi-channeled items

darwin/importer/importer.py
```diff
@@ -1012,30 +1051,68 @@ def process_local_file(local_path):
                 for file in tqdm(
                     files_to_track, desc="Importing annotations from local file"
                 ):
-                    import_annotation(file)
+                    remote_file = remote_files[file.full_path]
+                    file_slot_warnings, file_slot_errors = (
+                        _validate_slot_alignment_and_import(
+                            file, remote_file, annotation_format
```
JBWilkie, line 1057: A note here that I initially wanted to perform all slot <-> annotation validation **before** getting to this late stage of the import. However, this isn't possible because:
- Early on in the import, we have access to the `local_files` and `remote_files` variables that allow us to check for warnings and errors
- However, if we find any annotations imported to multi-channeled items that we need to adjust so that they are pointed at the base slot, we run into a problem. If at that stage, we make changes to the `local_files` variable (which represents the annotation data to be imported), the changes aren't persisted through to the stage where we perform the imports. This is because the flow of `import_annotations()` is such that we read in local files twice, [presumably to prevent memory issues(](https://github.com/v7labs/darwin-py/blob/v0.8.10/darwin/importer/importer.py#L356)?)

You can see my initial attempt in [this commit](https://github.com/v7labs/darwin-py/pull/895/commits/30bcbe9c0a88692cb6e3f22d320824741d4fdd99). I've done some digging, and I'm unable to determine why we need to read the annotations in twice. Would it be better to go back to the original approach in the above commit, and simply pass the annotations from the 1st read all the way through?

PR #896, opened 2024-08-07
[DAR-2981][External] Catch & handle blocking errors / non-blocking warnings when importing annotations to slots

JBWilkie:
# Problem
When importing annotations to multi-slotted or multi-channeled items, there are a few ways users can go wrong that we want to start handling:

**Multi-slotted items:**
All annotations that are not linked to a slot are imported to the default slot, which is the 1st slot listed under `slots` in the [List Items](https://docs.v7labs.com/reference/items-list) API response. There is no feedback that this is happening

**Multi-channeled items:**
All annotations that are not linked to a slot are also imported to the default slot, which is the same as above. However, multi-channeled items require that all of their annotations exist in the base slot, which is not necessarily the same as the default slot. The base slot is the first slot listed under `layout.slots_grid` in the [List Items](https://docs.v7labs.com/reference/items-list) API response. If annotations are uploaded to a non-base slot, they will not be visible in the workview

# Solution
To assist users when they run into these situations, we introduce:
- 1: A non-blocking warning for annotations imported to multi-slotted items that don't specify the slot. The warning explains that the annotation will be uploaded to the default slot
- 2: A non-blocking warning for annotations imported to multi-channeled items that don't specify the slot. We import it to the base slot & provide feedback on this in the warning, as well as why this is necessary
- 3: A blocking error for annotations imported to multi-channeled items that specify the non-base slot. We do not import these annotations from affected annotation files, and explain in the warning that these annotations need to be pointed to the base slot and re-imported

Warnings and errors accumulate during the import process, and are displayed afterwards. All files except those that generate at least one error are imported. A sample warning looks like this:
```
WARNING: 2 file(s) have the following non-blocking warnings. Imports of these files will continue:
- File: /Users/john/Desktop/import/cars-channels_2.json, warnings:
  - Annotation imported to multi-channeled item not assigned a slot. Uploading to the base slot: 3
  - Annotation imported to multi-channeled item not assigned a slot. Uploading to the base slot: 3
- File: /Users/john/Desktop/import/cars-backwards-slots.json, warnings:
  - Annotation imported to multi-slotted item not assigned slot. Uploading to the default slot: 1
  - Annotation imported to multi-slotted item not assigned slot. Uploading to the default slot: 1
WARNING: 1 file(s) have the following blocking issues and will not be imported. Please resolve these issues and re-import them.
- File: /Users/john/Desktop/import/cars-channels.json, issues:
  - Annotation is linked to slot 4 of the multi-channeled item /dir2/cars-channels. Annotations uploaded to multi-channeled items have to be uploaded to the base slot, which for this item is 3.
  - Annotation is linked to slot 1 of the multi-channeled item /dir2/cars-channels. Annotations uploaded to multi-channeled items have to be uploaded to the base slot, which for this item is 3.
```
Finally, if users attempt to import annotations to multi-slotted or multi-channeled items with a format other than Darwin JSON 2.0, we raise a `TypeError`. This is because only Darwin JSON 2.0 supports multi-slotted & multi-channeled items. For example:
```
raise TypeError(
TypeError: You are attempting to import annotations to multi-slotted or multi-channeled items using an annotation format (coco) that doesn't support them. To import annotations to multi-slotted or multi-channeled items, please use the Darwin JSON 2.0 format: https://docs.v7labs.com/reference/darwin-json
```

# Changelog
Introduced non-blocking warnings & blocking errors when importing annotations to multi-slotted and multi-channeled items. If errors or warnings are encountered, we provide feedback on how to resolve the errors


darwin/importer/importer.py
```diff
@@ -1031,11 +1062,8 @@ def process_local_file(local_path):
                 except Exception as exc:
                     console.print(f"Generated an exception: {exc}", style="error")
     else:
-        for local_path in tqdm(
-            {local_file.path for local_file in local_files},
-            desc="Processing local annotation files",
-        ):
-            process_local_file(local_path)
+        for local_file in tqdm(local_files, desc="Processing local annotation files"):
+            process_local_file(local_file)
```
JBWilkie, line 1066: This is a change I'd like someone to review. Currently, the importer reads local annotation files in twice:
- [Once early on](https://github.com/v7labs/darwin-py/blob/f974484ca2d96e4aa2bbb2d3042476d4e59b3085/darwin/importer/importer.py#L791) in the import flow to perform validation of the contents of the files
- [A second time](https://github.com/v7labs/darwin-py/blob/f974484ca2d96e4aa2bbb2d3042476d4e59b3085/darwin/importer/importer.py#L953) just before import to send the annotation data

I can't figure out why we would do this, even though [it was explicitly commented here](https://github.com/v7labs/darwin-py/blob/v0.8.10/darwin/importer/importer.py#L356) previously. I don't believe it would help with memory, since we hold all the annotations in memory after the 1st read anyway.

This change means we read in the annotations only once at the beginning, then pass them all the way through to the stage where we're sending import requests. This is useful because the slot <-> annotation validation this PR introduces needs to make changes to the slots of some annotations. If we do this early on and then read the files in again later, the changes we made don't persist. With this change, they do persist

darwin/importer/importer.py
```diff
@@ -1472,3 +1500,168 @@ def _overwrite_warning(
...
+    slot_warnings : Dict[str, List[str]]
+        A dictionary of non-blocking warnings for each file
+    """
+
+    slot_errors, slot_warnings = {}, {}
+    for local_file in local_files:
+        remote_file = remote_files[local_file.full_path]
+        local_file_path = str(local_file.path)
+        if len(remote_file["slot_names"]) == 1:
+            continue  # Skip single-slotted items
```
JBWilkie, line 1568: While working on this, I noticed another case we could check against. When uploading annotations to any item (even single-slotted ones), it's possible to upload them to a slot that doesn't exist. In this case they simply don't appear in the workview, but these "black holed" annotations are in exports of the item. Should we block this?

If so, I can change this section and add that validation here

darwin/importer/importer.py
```diff
@@ -184,33 +184,57 @@ def _build_attribute_lookup(dataset: "RemoteDataset") -> Dict[str, Unknown]:
...
 
-def _get_slot_name(remote_file: DatasetItem) -> str:
-    slot = next(iter(remote_file.slots), {"slot_name": "0"})
-    if slot:
-        return slot["slot_name"]
-    else:
-        return "0"
+def _get_slot_names(remote_file: DatasetItem) -> List[str]:
+    """
+    Returns the a list of slot names for a dataset item:
```
JBWilkie, line 215:
```suggestion
    Returns alist of slot names for a dataset item:
```

shernshiou (approved):
Discussed with @JBWilkie . 👍🏼 for the docs.

PR #898, opened 2024-08-14
[DAR-3424][Externa] Added `Dockerfile`

JBWilkie:
# Problem
There have historically been many requests to use darwin-py as cloud infrastructure. These have been tricky to support due to build dependencies in darwin-py

# Solution
This PR allows darwin-py to be deployed easily on a container, removing the difficulties of build dependencies

# Changelog
Added `Dockerfile` for darwin-py

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,55 @@
+# Use an official Python runtime as a parent image
+FROM python:3.10-slim
+
+# Environment variables
+ARG YOUR_ENV
+
+ENV YOUR_ENV=${YOUR_ENV:-development} \
+    PYTHONFAULTHANDLER=1 \
+    PYTHONUNBUFFERED=1 \
+    PYTHONHASHSEED=random \
+    PIP_NO_CACHE_DIR=off \
+    PIP_DISABLE_PIP_VERSION_CHECK=on \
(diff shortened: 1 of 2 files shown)


shernshiou:
Overall the Dockerfile looks good, however I am not sure about `YOUR_ENV`. If I understand correctly, the purpose is to select whether it is development or production. Since the purpose of this Dockerfile is solely for development purpose, we can assume that the environment is development and remove `YOUR_ENV`. 

Otherwise, we can find other alternative name to `YOUR_ENV` to correctly describe it's purpose.

JBWilkie:
> Overall the Dockerfile looks good, however I am not sure about `YOUR_ENV`. If I understand correctly, the purpose is to select whether it is development or production. Since the purpose of this Dockerfile is solely for development purpose, we can assume that the environment is development and remove `YOUR_ENV`.
> 
> Otherwise, we can find other alternative name to `YOUR_ENV` to correctly describe it's purpose.

The intention is actually to allow people to deploy darwin-py easily in a production environment. We have fairly regular requests from clients who want to import annotations or data as part of an automated pipeline they have running in cloud infra. This can be tricky with darwin-py due to it's somewhat complex dependencies, so the idea is that a Docker image solves this

shernshiou (approved):

PR #903, opened 2024-08-20
[DAR-3431][External] Adding passing E2E tests as a requirement to release

JBWilkie:
# Problem
We've recently had a couple annotation import bugs that would have been caught by E2E tests. However:
- The Slack bot we have configured to notify us of failing nightly E2Es is broken
- Passing E2Es are not a requirement to release new darwin-py versions

# Solution
This PR:
- 1: Fixes the Slack bot that notifies us when nightly E2E runs fail
- 2: Introduces an E2E test run that needs to pass during the release process for it to succeed

# Changelog
Introduced a requirement for E2E tests to pass before new darwin-py versions can be released

### DIFF
--- a/.github/workflows/JOB_e2e.yml
+++ b/.github/workflows/JOB_e2e.yml
@@ -7,6 +7,7 @@ on:
   pull_request:
     paths:
       - "e2e_tests/**"
+  workflow_call:
 jobs:
   e2e:
     name: End to End Testing
--- a/.github/workflows/JOB_e2e.yml
+++ b/.github/workflows/JOB_e2e.yml
@@ -42,14 +42,10 @@ jobs:
           E2E_ENVIRONMENT: ${{ secrets.E2E_ENVIRONMENT }}
           E2E_TEAM: ${{ secrets.E2E_TEAM }}
 
-      # Add this step to force a failure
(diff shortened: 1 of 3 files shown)


.github/workflows/JOB_e2e.yml
```diff
@@ -7,6 +7,7 @@ on:
   pull_request:
     paths:
       - "e2e_tests/**"
+  workflow_call:
```
JBWilkie, line 10: Need this trigger to allow it to be called during the release event

darwin/importer/importer.py
```diff
@@ -957,7 +957,7 @@ def import_annotations(  # noqa: C901
 
     def import_annotation(parsed_file):
         image_id = remote_files[parsed_file.full_path]["item_id"]
-        default_slot_name = remote_files[parsed_file.full_path]["slot_names"]
+        default_slot_name = remote_files[parsed_file.full_path]["slot_names"][0]
```
JBWilkie, line 960: Bonus bug fix when importing annotation files that don't specify a slot name. This is one of the bugs we would have caught with E2E failures

dorfmanrobert (approved):

PR #914, opened 2024-08-21
Granted write access to EVENT_tag so it can create releases

JBWilkie:
# Problem
`EVENT_tag.yml` needs `write` permission to create releases, but it only has `read` permissions

# Solution
Grant it `write` permissions instead


saurbhc (approved):
LGTM

PR #915, opened 2024-08-21
[DAR-3041][External] Change the behaviour of the `force_slots` argument of `pull()`

JBWilkie:
# Problem
Currently when pulling a dataset release with darwin-py, if:
- 1: Any item is a multi-slotted item, or:
- 2: Any item has a slot with multiple source files

Then every item is named and structured into directories in the multi-slotted way. That is the paths include the slot name: `{prefix}/{item_name}/{slot_name}/{file_name}`

This is because if any item meets either condition 1 or 2 above, we set `force_slots=True` for all items

If a user passes `force_slots=False` to `pull()` (which is the default behaviour), we should pass `force_slots=False` for all single-slotted, single-source-file items, but as `True` for all other items, since in that case it only makes sense for the local paths of multi-slotted / multi-source-file items to include the slot names

# Solution
Instead of passing the same value of `force_slots` for all items, determine & record the value of `force_slots` for every item to be downloaded, then propagate these values into the download functions

# Changelog
When downloading dataset releases containing a mixture of multi-slotted & single-slotted items, only represent the slots of the multi-slotted items locally


darwin/dataset/download_manager.py
```diff
@@ -64,8 +64,9 @@ def download_all_images_from_annotations(
         Recreate folders
     video_frames : bool, default: False
         Pulls video frames images instead of video files
-    force_slots: bool
+    force_slots: bool, default: False
         Pulls all slots of items into deeper file structure ({prefix}/{item_name}/{slot_name}/{file_name})
+        If False, all multi-slotted items and items with slots containing multiple source files will be downloaded as the deeper file structure
```
dorfmanrobert, line 69: Shouldn't this say: "If True, <rest>"

dorfmanrobert (approved):
nice lgtm

PR #925, opened 2024-09-12
[DAR-3771][External] Import of annotation-level properties

JBWilkie:
# Problem
darwin-py is currently incapable of handling:
- 1: A per-property `granularity` field in the `.v7/metadata.json` properties manifest file
- 2: `frame_index=null` values for annotation-level property values

Both concepts must be supported by darwin-py to allow the same import functionality as section-level properties (which are currently supported properties)

# Solution
- Extend the `Property` class with `granularity`
- Introduce the `PropertyGranularity` enum to represent the 3 property granularities

Adjusted the properties import functions to ensure :
1. Annotation-level property values can be imported, but that `granularity` is optional so that section-level property values can still be imported
2. Annotation-level property values present in the properties manifest but that are missing in the team can be created in the team
3. Annotation-level properties present in the properties manifest but that are missing in the team can be created in the team

We also add unit tests for `_import_annotations`, as none previously existed. These test that for both section & annotation-level properties:
- Existing property values can be imported as expected
- New property values can be created from the manifest file
- New properties themselves can be created from a manifest file, and section-level properties can still be created without specifying a `granularity`

# Changelog
Added support for annotation-level properties

### DIFF
--- a/darwin/future/tests/core/fixtures.py
+++ b/darwin/future/tests/core/fixtures.py
@@ -8,7 +8,11 @@ import pytest
 from darwin.future.core.client import ClientCore, DarwinConfig
 from darwin.future.data_objects.dataset import DatasetCore
 from darwin.future.data_objects.item import ItemCore, ItemLayout, ItemSlot
-from darwin.future.data_objects.properties import FullProperty, PropertyValue
+from darwin.future.data_objects.properties import (
+    FullProperty,
+    PropertyValue,
+    PropertyGranularity,
+)
 from darwin.future.data_objects.team import TeamCore, TeamMemberCore
 from darwin.future.data_objects.team_member_role import TeamMemberRole
 from darwin.future.data_objects.workflow import WorkflowCore
@@ -38,6 +42,7 @@ def base_property_object(base_property_value: PropertyValue) -> FullProperty:
         annotation_class_id=0,
         property_values=[base_property_value],
         options=[base_property_value],
+        granularity=PropertyGranularity("section"),
     )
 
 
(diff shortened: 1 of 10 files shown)


AndriiKlymchuk (approved):

darwin/future/tests/core/fixtures.py
```diff
@@ -38,6 +42,7 @@ def base_property_object(base_property_value: PropertyValue) -> FullProperty:
         annotation_class_id=0,
         property_values=[base_property_value],
         options=[base_property_value],
+        granularity=PropertyGranularity("section"),
```
umbertoDifa, line 45: This is questionable: currently the BE has a default 'section' if no granularity is provided. Here we're building a fixture that disregards the BE default. It's not wrong, but I think it'd be interesting to also have a test that doesn't specify the granularity at all when creating a property

JBWilkie: The unit tests I just added test creating section-level properties from a manifest file where no `granularity` is specified

darwin/torch/transforms.py
```diff
@@ -368,8 +368,11 @@ def _pre_process(self, image: np.ndarray, annotation: dict) -> dict:
         if (
             masks is not None and masks.numel() > 0
         ):  # using numel() to check if tensor is non-empty
-            print("WE GOT MASKS")
-            albumentation_dict["masks"] = masks.numpy()
+            if isinstance(masks, torch.Tensor):
+                masks = masks.numpy()
+            if masks.ndim == 3:  # Ensure masks is a list of numpy arrays
+                masks = [masks[i] for i in range(masks.shape[0])]
+            albumentation_dict["masks"] = masks
```
JBWilkie, line 375: Change that is unrelated to the ticket, but necessary to resolve [these failing tests](https://github.com/v7labs/darwin-py/actions/runs/10852759702/job/30119531670):

PR #928, opened 2024-09-20
[DAR-4036][External] E2E tests for `push` and some coverage for `import`

JBWilkie:
# Problem
darwin-py's E2E tests are sparse

# Solution
Add a `e2e_tests/cli/test_push.py` file with the following tests:
- `test_push_mixed_filetypes` - Test pushing a directory of files containing various filetypes and verify they finish processing
- `test_push_nested_directory_of_images` - Test pushing a nested directory structure of some images with the `preserve_folders` flag. Verify they finish processing and they end up in the correct remote paths
- `test_push_videos_with_non_native_fps` - Test that if FPS is set, that the value is respected in the resulting video items

These tests will wait a maximum of 10 minutes for all items to finish processing. If this timeout is exceeded, the test will fail

Add a `e2e_tests/cli_test_import.py` file with the following tests:
- `test_import_basic_annotations_to_images` - Test importing a set of basic image annotations (no sub-types or properties) to a set of pre-registered files in a dataset
- `test_import_annotations_with_subtypes_to_images` - Test importing a set of image annotations including subtypes & properties to a set of pre-registered files in a dataset
- `test_annotation_classes_are_created_on_import` - Test that importing non-existent annotation classes creates those classes in the target Darwin team
- `test_annotation_classes_are_created_with_properties_on_import` - Test that importing non-existent annotation classes with properties creates those classes and properties in the target Darwin team
- `test_appending_annotations` - Test that appending annotations to an item with already existing annotations does not overwrite the original annotations
- `test_overwriting_annotations` - Test that the `--overwrite` flag allows bypassing of the overwrite warning when importing to items with already existing annotations
- `test_annotation_overwrite_warning` - Test that importing annotations to an item with already existing annotations throws a warning if not using the `--append` or `--overwrite` flags

# Changelog
Expanded darwin-py E2E


e2e_tests/cli/test_import.py
```diff
@@ -0,0 +1,333 @@
...
+                expected_annotation_type = (
+                    expected_annotation.annotation_class.annotation_type
+                )
+                actual_annotation = next(
+                    (
+                        annotation
+                        for annotation in actual_annotations
+                        if annotation.data == expected_annotation_data
+                        and annotation.annotation_class.annotation_type
+                        == expected_annotation_type
```
JBWilkie, line 75: This is necessary because the `data` field of both the `tag` and `mask` types is just `{}`, so we need to check that `annotation_type` matches too

e2e_tests/helpers.py
```diff
@@ -48,6 +72,9 @@ def run_cli_command(
     if yes:
         command = f"yes Y | {command}"
 
+    # Prefix the command with 'poetry run' to ensure it runs in the Poetry shell
+    command = f"poetry run {command}"
```
JBWilkie, line 76: This makes running E2Es locally less error prone by forcing them to run in the poetry shell that's guaranteed to point to the correct `darwin-py` installation

e2e_tests/cli/test_push.py
```diff
@@ -0,0 +1,112 @@
+from pathlib import Path
```
umbertoDifa, line 1: beautiful tests

e2e_tests/cli/test_convert.py
```diff
@@ -0,0 +1,33 @@
+# from pathlib import Path
```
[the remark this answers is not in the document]
umbertoDifa: I see there are  TODO's also in other files, let's make sure to clean them up before merge

umbertoDifa (commented):
Really excited for these e2e!

e2e_tests/helpers.py
saurbhc: btw, python also has [sys.executable](https://docs.python.org/3/library/sys.html#sys.executable) 😄

e2e_tests/cli/test_import.py
```diff
@@ -0,0 +1,333 @@
...
+    assert_cli(result, 0)
+    # 2nd import to append more annotations
+    result = run_cli_command(
+        f"darwin dataset import {local_dataset.name} darwin {import_dir} --append"
+    )
+    assert_cli(result, 0)
+    with tempfile.TemporaryDirectory() as tmp_dir_str:
+        tmp_dir = Path(tmp_dir_str)
+        export_and_download_annotations(tmp_dir, local_dataset, config_values)
+        validate_downloaded_annotations(tmp_dir, import_dir, appending=True)
```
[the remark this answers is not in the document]
JBWilkie: This makes sense - What I can do is create a new `import_dir` specifically for this test containing 2 files (half & half), then I can remove the `appending` concept from `validate_download_annotations)`

e2e_tests/helpers.py
JBWilkie: I didn't know this, it's much nicer. thank you!

e2e_tests/cli/test_convert.py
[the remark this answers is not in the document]
JBWilkie: Removed all the TODOs. There are for future E2E PRs so don't have to be present here

e2e_tests/cli/test_import.py
```diff
@@ -0,0 +1,333 @@
...
+        expected_annotation_files = {
+            file.name: str(file)
+            for file in import_dir.iterdir()
+            if file.is_file() and not file.name.startswith(".")
+        }
+        actual_annotation_files = {
+            file.name: str(file)
+            for file in annotations_dir.iterdir()
+            if file.is_file() and not file.name.startswith(".")
+        }
```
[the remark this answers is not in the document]
JBWilkie: This is all sensible feedback, no reason not to improve the naming convention for others who will work on these tests in the future

e2e_tests/cli/test_import.py
```diff
@@ -0,0 +1,292 @@
...
+    """
+    local_dataset.register_read_only_items(config_values)
+    expected_annotations_dir = (
+        Path(__file__).parents[1]
+        / "data"
+        / "import"
+        / "image_annotations_split_in_two_files"
+    )
+    result = run_cli_command(
+        f"darwin dataset import {local_dataset.name} darwin {expected_annotations_dir}"
```
umbertoDifa, line 225: shouldn't this include --append?

e2e_tests/cli/test_import.py
```diff
@@ -0,0 +1,292 @@
...
+    )
+    assert_cli(result, 0)
+    assert_cli(result, 0)
```
umbertoDifa, line 228: double assert?

umbertoDifa (approved):

PR #929, opened 2024-09-23
[DAR-3919][External] Warn when importing annotations with multiple instance IDs

JBWilkie:
# Problem
With the release of static instance IDs, if uploading an annotation with multiple instance IDs - Only the value of the instance ID on the first keyframe is used for import

This may confuse users who expect static instance IDs to operate like dynamic instance IDs, which is the current behaviour

# Solution
If the static instance IDs feature flag is present & enabled, OR if the flag doesn't exist, we perform the following check: For each frame of every video annotation check sub-annotations. If there is >1 unique instance ID value in any annotation, record this

After checking all annotations, print a warning for each affected file as follows:
```
The following files have at least one annotation with multiple instance IDs. Your team has static instance IDs enabled, so only the first instance ID of each annotation will be imported:
- File: {local_annotation_file_path_1} has {n_1} annotation(s) with multiple instance IDs
- File: {local_annotation_file_path_2} has {n_2} annotation(s) with multiple instance IDs
```

We check if the flag is present so that the code will still work when the flag is deprecated in the future. From this point onward, we will always perform the check

# Changelog
Added a warning when importing annotations with multiple instance ID values where only 1 value will be imported due to static instance IDs

### DIFF
--- a/darwin/importer/importer.py
+++ b/darwin/importer/importer.py
@@ -865,6 +865,22 @@ def import_annotations(  # noqa: C901
         slot_errors, slot_warnings, annotation_format, console
     )
 
+    if annotation_format == "darwin":
+        dataset.client.load_feature_flags()
+        
+        # Check if the flag exists. WHen the flag is deprecated in the future we will always perform this check
+        static_instance_id_feature_flag_exists = any(
+            feature.name == "STATIC_INSTANCE_ID"
+            for feature in dataset.client.features.get(dataset.team, [])
+        )
+        check_for_multi_instance_id_annotations = (
+            static_instance_id_feature_flag_exists
+            and dataset.client.feature_enabled("STATIC_INSTANCE_ID")
+        ) or not static_instance_id_feature_flag_exists
+
+        if check_for_multi_instance_id_annotations:
+            warn_for_annotations_with_multiple_instance_ids(local_files, console)
+
     console.print(
         f"{len(local_files) + len(local_files_missing_remotely)} annotation file(s) found.",
         style="info",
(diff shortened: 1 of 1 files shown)


darwin/importer/importer.py
```diff
@@ -1680,3 +1696,53 @@ def _display_slot_warnings_and_errors(
...
+        console.print(
+            "The following files have at least one annotation with multiple instance IDs. Your team has static instance IDs enabled, so only the first instance ID of each annotation will be imported:",
+            style="warning",
+        )
+        for file in files_with_multi_instance_id_annotations:
+            console.print(
+                f"- File: {file} has {files_with_multi_instance_id_annotations[file]} annotation(s) with multiple instance IDs"
+            )
+        if not secure_continue_request():
+            raise SystemExit("Execution stopped by user.")
```
JBWilkie, line 1748: I'm not sure if we want to throw a warning that requires input to continue here. If we do want to, maybe we should provide a way of bypassing the warning? Open to feedback

darwin/importer/importer.py
```diff
@@ -1680,3 +1696,53 @@ def _display_slot_warnings_and_errors(
...
+
+            if len(unique_instance_ids) > 1:
+                if file.path not in files_with_multi_instance_id_annotations:
+                    files_with_multi_instance_id_annotations[file.path] = 1
+                else:
+                    files_with_multi_instance_id_annotations[file.path] += 1
+
+    if files_with_multi_instance_id_annotations:
+        console.print(
+            "The following files have at least one annotation with multiple instance IDs. Your team has static instance IDs enabled, so only the first instance ID of each annotation will be imported:",
```
JBWilkie, line 1740: There's probably better wording for this warning

darwin/importer/importer.py
```diff
@@ -865,6 +865,22 @@ def import_annotations(  # noqa: C901
         slot_errors, slot_warnings, annotation_format, console
     )
 
+    if annotation_format == "darwin":
+        dataset.client.load_feature_flags()
+
+        # Check if the flag exists. When the flag is deprecated in the future we will always perform this check
+        static_instance_id_feature_flag_exists = any(
+            feature.name == "STATIC_INSTANCE_ID"
+            for feature in dataset.client.features.get(dataset.team, [])
```
dorfmanrobert, line 874: why do we need this `static_instance_id_feature_flag_exists`? it exists, so seems this is unneeded

JBWilkie: We only want to run the check that this PR introduces if:
- 1: The `STATIC_INSTANCE_ID` feature flag exists **and,** it's value for the relevant workspace is `True`, or:
- 2: The flag doesn't exist. At this point, the feature will be globally enabled so we will always want to run the check

If we run the check and either of these things isn't true, imported instance IDs can have multiple values throughout an annotation, so the warnings would be false

dorfmanrobert: oooh makes sense thanks!

dorfmanrobert (approved):

PR #930, opened 2024-09-24
[DAR-3920][External] Bug fix for importing property values to identically named properties during property creation

JBWilkie:
# Problem
This PR address an issue that occurs when creating two identically named properties on separate classes **and** then importing values of those properties in the same import action. In this scenario, we get this type of error:
```
Errors importing 000000580635.jpg
        {'errors': {'annotations': [{}, {}, {}, {}, {'annotation_properties': ["index '0', property 'bc5b7760-7786-4a0c-af89-8fdace15a0bf': property is not linked to the annotation's class"]}, {'annotation_properties': ["index '0', 
property 'bc5b7760-7786-4a0c-af89-8fdace15a0bf': property is not linked to the annotation's class"]}]}}
```
However: The import does successfully create the properties & their values. Additionally, a 2nd import will import the annotations without error

The problem is that the `annotation_property_map` returned from `_import_properties` is constructed incorrectly in this scenario. Specifically, we only used to check that the name of a property matched the name of the property to be imported. If importing annotations containing multiple identically named properties, this logic allows for mis-matching

# Solution
In addition to checking that the property names match, check that the annotation classes match too

# Changelog
Fixed an edge case when importing property values during property creation

### DIFF
--- a/darwin/importer/importer.py
+++ b/darwin/importer/importer.py
@@ -671,12 +671,23 @@ def _import_properties(
         if not annotation_id_map.get(annotation_id):
             continue
         annotation = annotation_id_map[annotation_id]
-
+        annotation_class = annotation.annotation_class
+        annotation_class_name = annotation_class.name
+        annotation_type = (
+            annotation_class.annotation_internal_type
+            or annotation_class.annotation_type
+        )
+        annotation_class_id = annotation_class_ids_map[
+            (annotation_class_name, annotation_type)
+        ]
         for a_prop in annotation.properties or []:
             frame_index = str(a_prop.frame_index)
 
             for prop in created_properties + updated_properties:
-                if prop.name == a_prop.name:
+                if (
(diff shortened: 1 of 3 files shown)


dorfmanrobert (approved):
Nice find! Non-blocking, but seems an alternative approach could be to construct `created_properties` and `updated_properties` as maps with keys as `annotation_class_id` rather than lists of all created/updated properties, so that only need to loop over the relevant properties for each ann when updating the `annotation_property_map`

JBWilkie:
> Nice find! Non-blocking, but seems an alternative approach could be to construct `created_properties` and `updated_properties` as maps with keys as `annotation_class_id` rather than lists of all created/updated properties, so that only need to loop over the relevant properties for each ann when updating the `annotation_property_map`

I think this logic will change in the item-level property import that I'm about to continue working on. Given this, I'm going to merge this but have included this as a point in the relevant ticket

PR #931, opened 2024-09-26
[DAR-4041][External] Add multi-file E2E tests

JBWilkie:
# Problem
No E2E tests for:
- Pushing multi-file items
- Importing annotations to multi-file items

# Solution
Add the following E2E tests:
`push`:
- `test_push_multi_slotted_item`: Upload a multi-slotted item via the CLI
- `test_push_multi_channel_item`: Upload a multi-channel item via the CLI
- `test_push_dicom_series`: Upload a multi-file DICOM series via the CLI

`import`:
- `test_import_annotations_to_multi_slotted_item_without_slots_defined`: Upload annotations to a multi-slotted item without aligning each annotation to a slot. All annotations should end up in the item's first slot
- `test_import_annotations_to_multi_slotted_item_with_slots_defined`: Upload annotations to a multi-slotted item where each annotation is aligned with a particular slot. Each annotation should end up in the correct slot
- `test_import_annotations_to_multi_channel_item_without_slots_defined`: Upload annotations to a multi-channel item without aligning each annotation to a slot. All annotations should end up the base slot
- `test_import_annotations_to_multi_channel_item_with_slots_defined`: Upload annotations to a multi-channel item where each annotation is aligned with the base slot. All annotations should end up in the base slot
- `test_import_annotations_to_multi_channel_item_non_base_slot`: Upload annotations to a multi-channel item where each annotation is aligned with a non-base slot. The importer should throw an error

# Changelog
Added E2E test coverage for uploading & importing annotations to multi-file items


e2e_tests/cli/test_import.py
```diff
@@ -78,15 +80,55 @@ def assert_same_annotation_properties(
...
+    Returns the base slot for the nth item in a specific `E2EDataset`. The base slot is
+    always the first listed slot
+    """
+    items = list_items(
+        config_values.api_key,
+        dataset_id,
+        config_values.team_slug,
+        config_values.server,
+    )
+    return items[item_idx]["slots"][0]["slot_name"]
```
umbertoDifa, line 96: This might not be an issue now, but what if the BE decided to return the list of items in a different order? Wouldn't it be safer to find an item by item_id or name?

e2e_tests/cli/test_import.py
```diff
@@ -78,15 +80,55 @@ def assert_same_annotation_properties(
             assert expected_property in actual_properties  # type : ignore
 
 
+def get_base_slot_of_item(
```
umbertoDifa, line 83: `get_base_slot_name_of_item` it returns the name not the slot right?

e2e_tests/cli/test_import.py
```diff
@@ -78,15 +80,55 @@ def assert_same_annotation_properties(
...
+    - `single_slotted`: Perform no checks
+    - `multi_slotted`: Ensures `actual_annotation.slot_names` is equivalent to
+      `expected_annotation.slot_names`
+    - `multi_channel`: Ensures the `actual_annotation` is tied to the base slot
+    """
+    if item_type == "multi_slotted":
+        if expected_annotation.slot_names:
+            assert expected_annotation.slot_names == actual_annotation.slot_names
+        else:
+            assert actual_annotation.slot_names == [base_slot]
```
umbertoDifa, line 117: Trying to think when we have this case: if it's a multislot and I don't specify the slot_name of an annotation then by default it gets uploaded to the base slot?

e2e_tests/cli/test_import.py
```diff
@@ -121,8 +163,11 @@ def compare_annotations_export(
             actual_annotation = find_matching_actual_annotation(
                 expected_annotation, actual_annotations
             )
-            assert_same_annotation_data(actual_annotation, expected_annotation)
+            assert_same_annotation_data(expected_annotation, actual_annotation)
             assert_same_annotation_properties(expected_annotation, actual_annotation)
+            assert_annotation_slot_alignment(
```
umbertoDifa, line 168: nit:I'd probably expect something named `assert_same_annotation_slot_name`?

e2e_tests/cli/test_import.py
```diff
@@ -311,3 +374,148 @@ def test_annotation_overwrite_warning(
...
+    assert_cli(result, 0)
+    base_slot = get_base_slot_of_item(config_values, local_dataset.id, item_idx=0)
+    with tempfile.TemporaryDirectory() as tmp_dir_str:
+        actual_annotations_dir = Path(tmp_dir_str)
+        export_and_download_annotations(
+            actual_annotations_dir, local_dataset, config_values
+        )
+        compare_annotations_export(
+            actual_annotations_dir, expected_annotations_dir, item_type, base_slot
+        )
```
umbertoDifa, line 496: it looks like these methods are identical and only the `expected_annotations_dir` changes. I wonder if it can be useful to simplify. The rule of thumb is: if you repeat it 3 times than you can DRY the code. Up to you

e2e_tests/cli/test_import.py
```diff
@@ -311,3 +374,148 @@ def test_annotation_overwrite_warning(
...
+
+
+def test_import_annotations_to_multi_channel_item_non_base_slot(
+    local_dataset: E2EDataset, config_values: ConfigValues
+) -> None:
+    """
+    Upload annotations to a multi-channel item where each annotation is aligned with a
```
umbertoDifa, line 503: What happens if I have at least one annotation that is aligned to a base_slot? Would that be the only one imported?

umbertoDifa (commented):
These tests are 🔥 , I just left minor comments

e2e_tests/cli/test_import.py
JBWilkie: Exactly

e2e_tests/cli/test_import.py
JBWilkie: No - If there is any annotation that is not aligned with the base slot, the importer will throw an error

e2e_tests/cli/test_import.py
```diff
@@ -78,15 +80,55 @@ def assert_same_annotation_properties(
             assert expected_property in actual_properties  # type : ignore
 
 
+def get_base_slot_of_item(
+    config_values: ConfigValues, dataset_id: int, item_idx: int
+) -> str:
+    """
+    Returns the base slot for the nth item in a specific `E2EDataset`. The base slot is
+    always the first listed slot
```
AndriiKlymchuk, line 88: When we talk about base slot for layers, we should retrieve it from the layout, not actual item slots. Unless I'm missing something.

e2e_tests/cli/test_import.py
AndriiKlymchuk: DRYing the code would make it way faster to read and understand

e2e_tests/cli/test_push.py
```diff
@@ -110,3 +110,170 @@ def test_push_videos_with_non_native_fps(
...
+                f"darwin dataset push {local_dataset.name} {tmp_dir}/flat_directory_of_2_dicom_files --item-merge-mode {merge_mode}"
+            )
+            assert_cli(result, 0)
+            wait_until_items_processed(config_values, local_dataset.id)
+            items = list_items(
+                config_values.api_key,
+                local_dataset.id,
+                config_values.team_slug,
+                config_values.server,
+            )
```
AndriiKlymchuk, line 272: The logic from within lines 258-272 repeats in every test. Maybe you can de-duplicate it. At least it is easier to od within lines 261-272

AndriiKlymchuk (approved):
Nothing critical. Just consider to address existing comments before merge.

e2e_tests/cli/test_push.py
AndriiKlymchuk, line 272: The logic from within lines 258-272 repeats in every test. Maybe you can de-duplicate it. At least it is easier to od within lines 261-272

AndriiKlymchuk, line 272: The logic here repeats in every test. Maybe you can de-duplicate it. At least it is easier to dot it from line 261

JBWilkie:
> These tests are 🔥 , I just left minor comments

Useful points, thank you! All addressed

JBWilkie:
> Nothing critical. Just consider to address existing comments before merge.

All addressed, thank you!

PR #936, opened 2024-10-09
[DAR-2860][External] Import of item-level properties

JBWilkie:
# Problem
darwin-py cannot currently import item-level properties

# Solution
- Extend the import code to handle item-level properties
- Add 2 end-to-end tests to cover import of item-level properties

# Changelog
Support for importing item-level properties


darwin/backend_v2.py
```diff
@@ -293,3 +293,18 @@ def _get_remote_annotations(
             The team slug.
         """
         return self._client._get(f"v2/teams/{team_slug}/items/{item_id}/annotations")
+
+    def _get_properties_state_for_item(
+        self, item_id: str, team_slug
```
dorfmanrobert, line 298: missing type on team_slug

darwin/datatypes.py
```diff
@@ -547,6 +550,9 @@ class AnnotationFile:
     #: List of ``VideoAnnotation``\s or ``Annotation``\s.
     annotations: Sequence[Union[Annotation, VideoAnnotation]]
 
+    # Item-level properties
+    item_properties: Optional[List[Dict[str, str]]] = None
```
dorfmanrobert, line 554: any reason we're using List, Dict rather than list, dict like for property_values?

darwin/future/data_objects/properties.py
```diff
@@ -81,31 +81,39 @@ class FullProperty(DefaultDarwin):
...
     options: Optional[List[PropertyValue]] = None
     granularity: PropertyGranularity = PropertyGranularity("section")
 
     def to_create_endpoint(
         self,
     ) -> dict:
-        if self.annotation_class_id is None:
+        if (
+            self.annotation_class_id is None
+            and self.granularity != PropertyGranularity.item
```
dorfmanrobert, line 94: double checking this now so could maybe have

```
    if self.granularity != PropertyGranularity.item:
        if self.annotation_class_id is None:
            raise ValueError("annotation_class_id must be set")
        include_fields["annotation_class_id"] = True
```
after `include_fields = ` so is clearer

darwin/importer/importer.py
```diff
@@ -278,18 +284,42 @@ def _resolve_annotation_classes(
     return local_classes_not_in_dataset, local_classes_not_in_team
 
 
-def _get_team_properties_annotation_lookup(client, team_slug):
+def _get_team_properties_annotation_lookup(
+    client: "Client", team_slug: str
+) -> Tuple[Dict[Tuple[str, Optional[int]], FullProperty], Dict[str, FullProperty]]:
```
dorfmanrobert, line 289: Some of these types are looking a bit complicated, could be nice when doing the refactor to define some more types

darwin/importer/importer.py
```diff
@@ -278,18 +284,42 @@ def _resolve_annotation_classes(
...
+    """Returns two lookup dictionaries for team properties:
+     - team_properties_annotation_lookup: (property-name, annotation_class_id): FullProperty object
+     - team_item_properties_lookup: property-name: FullProperty object
```
dorfmanrobert, line 292: nit: would be nice to have consistency in naming, makes it easier to understand: `team_item_properties_lookup` and  `team_annotation_properties_lookup`

darwin/importer/importer.py
```diff
@@ -318,12 +352,104 @@ def _update_payload_with_properties(
...
 
+def _update_payload_with_item_level_properties(
+    item_property_values, client, dataset, import_annotators, import_reviewers
+) -> List:
+    """
+    Adds item-level properties to the annotation import payload if any are present
+
+    Args:
+        item_property_values
+        cliennt
```
dorfmanrobert, line 363: client

darwin/importer/importer.py
```diff
@@ -318,12 +352,104 @@ def _update_payload_with_properties(
             annotation["annotation_properties"] = dict(_map)
 
 
+def _update_payload_with_item_level_properties(
+    item_property_values, client, dataset, import_annotators, import_reviewers
```
dorfmanrobert, line 356: types missing

darwin/importer/importer.py
dorfmanrobert, line 1: There seems to be unneeded duplicate calls in _import_annotations:
- We call _import_properties, which calls _get_team_properties_annotation_lookup
- We then call _update_payload_with_item_level_properties, which also calls _get_team_properties_annotation_lookup again

darwin/importer/importer.py
```diff
@@ -318,12 +352,104 @@ def _update_payload_with_properties(
             annotation["annotation_properties"] = dict(_map)
 
 
+def _update_payload_with_item_level_properties(
```
dorfmanrobert, line 355: Lets rename this and update the description unless we are updating the payload like we do in `_update_payload_with_properties`, since its not updating the payload

darwin/importer/importer.py
```diff
@@ -318,12 +352,104 @@ def _update_payload_with_properties(
...
+    serialized_item_level_properties: List[Any] = []
+    actors: List[dt.DictFreeForm] = []
+    # Get team properties
+    _, team_item_properties_lookup = _get_team_properties_annotation_lookup(
+        client, dataset.team
+    )
+    for item_property_value in item_property_values:
+        item_property = team_item_properties_lookup[item_property_value["name"]]
+        item_property_id = item_property.id
+        item_property_value_id = next(
```
dorfmanrobert, line 378: not sure how many we expect, but if is many might be worth making a lookup for this

darwin/importer/importer.py
```diff
@@ -581,37 +716,106 @@ def _import_properties(
...
+                console.print(
+                    f"- Updating property {full_property.name} ({full_property.type})",
+                )
             prop = client.update_property(
                 team_slug=full_property.slug, params=full_property
             )
             updated_properties.append(prop)
 
     # get latest team properties
     team_properties_annotation_lookup = _get_team_properties_annotation_lookup(
```
dorfmanrobert, line 817: doesn't this return two lookups now?

dorfmanrobert, line 817: this returns the item level prop lookup as well, so this check
```
        if (prop_name, annotation_class_id) not in team_properties_annotation_lookup:
            continue
```
will always be entered it seems

darwin/importer/importer.py
dorfmanrobert, line 1: As general comment, _import_properties really needs to be broken into pieces, but I assume that's what the future refactor is about

dorfmanrobert (changes requested):
A handful of small things, there is one that regards the actual logic though: https://github.com/v7labs/darwin-py/pull/936#discussion_r1795114064

darwin/datatypes.py
JBWilkie: No - This makes more sense

darwin/future/data_objects/properties.py
JBWilkie: Nice suggestion, this is cleaner

darwin/importer/importer.py
JBWilkie: Totally agree, will take place as part of DAR-4295

darwin/importer/importer.py
JBWilkie: Unfortunately I think `team_properties_annotation_lookup` has to be named this way. If it was called `team_annotation_properties_lookup`, this would imply it only contains data for annotation-level properties. However, it contains data for annotation-level and section-level properties. The naming convention for variables will be addressed in the DAR-4295 refactor

darwin/importer/importer.py
JBWilkie: Correct, DAR-4295

darwin/importer/importer.py
JBWilkie: Good catch, working on a fix now

darwin/importer/importer.py
dorfmanrobert: will be handled in refactor

dorfmanrobert (approved):

PR #940, opened 2024-10-10
[DAR-4299][External] Allow import of NifTI annotations to dataset items with >1 slot

JBWilkie:
# Problem
In darwin-py 1.0.4 as part of [DAR-2981](https://linear.app/v7labs/issue/DAR-2981/darwin-py-consider-channel-annotation-upload-behaviour), we prevented non-Darwin JSON 2.0 annotations from being imported to dataset items with >1 slot. This makes sense, except in the case of NifTI annotations, where we do want to allow users to import annotations to multi-slotted items

Therefore this ticket adds `nifti` to the list of supported multi-slotted formats. No other format should be supported at the time of this ticket
# Solution
This ticket adds `nifti` to the list of annotation formats that allow import to multi-slotted dataset items

# Changelog
Allow import of `nifti` annotations to multi-slotted dataset items


dorfmanrobert (approved):

darwin/importer/importer.py
```diff
@@ -2118,9 +2118,10 @@ def _display_slot_warnings_and_errors(
         If there are any warnings generated and the annotation format is not Darwin JSON 2.0
```
dorfmanrobert, line 2118: update this as well probably

PR #942, opened 2024-10-10
[DAR-4064][External] Annotation import E2E test for warning about importing multi-value instance IDs

JBWilkie:
# Problem
When using `darwin dataset import`, if a user pushes an annotation with >1 value AND the `STATIC_INSTANCE_ID` FF is on, we throw a warning that only the first value is imported. We have no E2E test for this currently

# Solution
Add an E2E test for the above

# Changelog
Add test coverage for warning about importing multiple instance ID values


JBWilkie:
Closing since on reflection I don't think this should be an E2E test, it's covered by unit tests

PR #948, opened 2024-10-24
[DAR-4489][External] Import item-level properties without annotations

JBWilkie:
# Problem
When importing annotations to items in Darwin JSON 2.0, item-level property values are ignored if there are no annotations. This happens because before item-level properties, all properties were tied to annotations, so it made sense to skip import in these cases

However, item-level properties are independent of annotations and so should be able to be imported independently

# Solution
Do not skip annotation files for import if they contain item-level properties but no annotations

# Changelog
Allow import of item-level properties without annotations

### DIFF
--- a/darwin/importer/importer.py
+++ b/darwin/importer/importer.py
@@ -1419,7 +1419,8 @@ def import_annotations(  # noqa: C901
         files_to_not_track = [
             file_to_track
             for file_to_track in parsed_files
-            if not file_to_track.annotations and (not delete_for_empty)
+            if not (file_to_track.annotations or file_to_track.item_properties)
+            and (not delete_for_empty)
         ]
 
         for file in files_to_not_track:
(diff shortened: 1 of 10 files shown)


darwin/importer/importer.py
```diff
@@ -1419,7 +1419,8 @@ def process_local_file(local_file):
         files_to_not_track = [
             file_to_track
             for file_to_track in parsed_files
-            if not file_to_track.annotations and (not delete_for_empty)
+            if not (file_to_track.annotations or file_to_track.item_properties)
+            and (not delete_for_empty)
```
dorfmanrobert, line 1423: This is equivalent to:
`not file_to_track.annotations and not file_to_track.item_properties`

Maybe I'm misunderstanding, but shouldn't we want
`not file_to_track.annotations or not file_to_track.item_properties`

dorfmanrobert: hmmm think i was misunderstanding and this is correct that we do want to add to `files_not_track` only if `not file_to_track.annotations and not file_to_track.item_properties`

dorfmanrobert (approved):

PR #953, opened 2024-11-03
[DAR-4140][External] Fixed NifTI import orientation

JBWilkie:
# Problem
Currently, the `NifTI` importer attempts to apply the `LPI` orientation to annotations to be imported, since all uploaded NifTI / DICOM files are transformed to `LPI`. However, the used function (`apply_orientation()`) flips all 3 axes. Why? The LPI orientation can be represented by [-1, -1, -1], which we feed into the function as the target orientation. However, that function just flips an axis if the corresponding value in the above 3-element array is -1, and doesn't flip if not. This means that for any input orientation **not** in `RAS`, at least one axis will be flipped incorrectly

# Solution
Make sure that for non-legacy imports, we only flip the annotation axes that are not already aligned with the `LPI` orientation

# Changelog
Resolved orientation issue when importing annotations in the `NifTI` format

### DIFF
--- a/tests/darwin/importer/formats/import_nifti_test.py
+++ b/tests/darwin/importer/formats/import_nifti_test.py
@@ -9,6 +9,7 @@ from zipfile import ZipFile
 import numpy as np
 import pytest
 from scipy import ndimage
+import nibabel as nib
 
 from darwin.datatypes import (
     Annotation,
@@ -17,7 +18,7 @@ from darwin.datatypes import (
     SubAnnotation,
     VideoAnnotation,
 )
-from darwin.importer.formats.nifti import get_new_axial_size, parse_path
+from darwin.importer.formats.nifti import get_new_axial_size, parse_path, process_nifti
 from tests.fixtures import *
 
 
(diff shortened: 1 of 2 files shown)


tests/darwin/importer/formats/import_nifti_test.py
```diff
@@ -238,6 +239,29 @@ def test_get_new_axial_size_with_isotropic():
     assert new_size == (20, 10)
 
 
+def test_process_nifti_orinetation(team_slug_darwin_json_v2):
```
dorfmanrobert, line 242:
```suggestion
def test_process_nifti_orientation(team_slug_darwin_json_v2):
```

tests/darwin/importer/formats/import_nifti_test.py
```diff
@@ -238,6 +239,29 @@ def test_get_new_axial_size_with_isotropic():
...
+    with tempfile.TemporaryDirectory() as tmpdir:
```
dorfmanrobert, line 243: may be worth having the file be non RAS or adding another test with non RAS

tests/darwin/importer/formats/import_nifti_test.py
```diff
@@ -238,6 +239,29 @@ def test_get_new_axial_size_with_isotropic():
...
+                Path(tmpdir)
+                / team_slug_darwin_json_v2
+                / "nifti"
+                / "releases"
+                / "latest"
+                / "annotations"
+                / "vol0_brain.nii.gz"
+            )
+            lpi_ornt = [[0.0, -1.0], [1.0, -1.0], [2.0, -1.0]]
+            ras_file = nib.load(filepath)
```
dorfmanrobert, line 256: may be worth having the file be non RAS or adding another test with non RAS

dorfmanrobert (approved):

PR #967, opened 2024-11-19
[DAR-4815][External] Full-cycle E2E tests for videos, multi-slotted items, and multi-channel items

JBWilkie:
# Problem
Currently our only full-cycle E2E test is for images, we should expand this coverage

# Solution
Add full-cycle E2E tests for videos, multi-slotted items, and multi-channel items

# Changelog
Improved E2E test coverage for more data types


darwin/dataset/download_manager.py
```diff
@@ -672,7 +672,7 @@ def _get_planned_image_paths(
         for slot in annotation.slots:
             slot_name = Path(slot.name)
             for source_file in slot.source_files:
-                file_name = source_file.file_name
+                file_name = source_file["file_name"]
```
JBWilkie, line 675: This change is related to DAR-4788 where it was discovered that multi-slotted items cannot be pulled due to this. Without the fix here, these tests would not pass

DAR-4788 remains open as we need to investigate the root cause of that - If it was a darwin-py issue or a backend regression

PR #968, opened 2024-11-19
[DAR-4932][Internal] correct pixdim application in mask to polygon

dorfmanrobert:
# Problem
When pixdims[0] ≠ pixdims[1], 3D mask annotations are incorrectly scaled when imported as polygons from a NifTI file.

# Solution
The key thing to note is in numpy 
- rows are indexed top to bottom
- columns are indexed left to right
But in visualisation space, the numpy columns represent the x axis and the numpy rows represent the y axis. Since pixdims[i] represents axis i in visualisation space, we need to switch the application of the pixdims.

# Changelog
(Will appear in release docs)

### DIFF
--- a/darwin/importer/formats/nifti.py
+++ b/darwin/importer/formats/nifti.py
@@ -376,7 +376,7 @@ def mask_to_polygon(
             else:
                 return {"x": y, "y": x}
         else:
-            return {"x": y * pixdims[1], "y": x * pixdims[0]}
+            return {"x": y * pixdims[0], "y": x * pixdims[1]}
 
     _labels, external_paths, _internal_paths = find_contours(mask)
     if len(external_paths) > 1:


tests/darwin/data/nifti/no-legacy/BRAINIX_NIFTI_ROI.nii.json
dorfmanrobert, line 1: The pixdims are `(0.7986109, 0.798611, 6.0000024)`, so this test covers the code changes. However, due to this bug, the expected outputs needed to be recreated.

darwin/importer/formats/nifti.py
```diff
@@ -376,7 +376,7 @@ def adjust_for_pixdims(x, y, pixdims):
             else:
                 return {"x": y, "y": x}
         else:
-            return {"x": y * pixdims[1], "y": x * pixdims[0]}
+            return {"x": y * pixdims[0], "y": x * pixdims[1]}
```
dorfmanrobert, line 379: This is the real change in logic.

darwin/importer/formats/nifti.py
```diff
@@ -529,9 +529,14 @@ def process_nifti(
...
-    data_array = nib.orientations.apply_orientation(img.get_fdata(), ornt)
-    pixdims = img.header.get_zooms()
+    orig_ax_codes = nib.orientations.aff2axcodes(img.affine)
+    orig_ornt = nib.orientations.axcodes2ornt(orig_ax_codes)
+    transform = nib.orientations.ornt_transform(orig_ornt, ornt)
+    reoriented_img = img.as_reoriented(transform)
+
+    data_array = reoriented_img.get_fdata()
+    pixdims = reoriented_img.header.get_zooms()
+
```
dorfmanrobert, line 539: These changes to `process_nifti` are unrelated and make the function more explicitly able to handle arbitrarily oriented outputs, even though this is not needed for importing to LPI. 

Before, the pixdims were taken from volume reoriented to RAS, meaning the output orientation needed to be in RAS order for the pixdims to be ordered correctly. Now the output's axes can be any order.

JBWilkie (approved):

PR #972, opened 2024-11-21
[DAR-4788][External] Allow download of frames extracted from videos & multi-slotted items

JBWilkie:
# Problem
- 1: `darwin-py` will download 2 files for every frame that was extracted from a video upon upload: The frame itself, and the video file. This results in massively duplicated data, much slower downloads (due to re-downloading videos), and an incorrect local folder structure
- 2: `darwin-py` will crash when trying to pull a release where any file is a multi-slotted item or a DICOM series

# Solution
Problem 1 occurs because on upload, each extracted frame is associated with two source files:
- 1: The frame itself
- 2: The video from which the frame was extracted

When darwin-py encounters a slot associated with >1 source file, it tries to download every source file. This PR makes the following change: There are only 2 supported types of data where >1 source file can be associated with a slot, so we check:
- If the slot is a DICOM series, or:
- If the slot is an extracted frame, in which case it will have exactly 2 source files: 1 image and 1 video

If we detect an extracted frame, we select only the image for download. If we find that the slot is neither supported type of data, we raise an error

Problem 2 occurs because we are trying to access each source file with `source_file.file_name`. `source_file` is a dictionary, so instead we need to do `source_file["file_name"]`

# Changelog
Resolved issues with:
- Pulling frames extracted from videos
- Pulling multi-slotted items & DICOM series


darwin/dataset/download_manager.py
```diff
@@ -670,9 +674,44 @@ def _get_planned_image_paths(
             return [images_path / filename]
     else:
         for slot in annotation.slots:
+            if len(slot.source_files) > 1:
+                # Check that the item is either a DICOM series or a frame extracted from a video
+                is_dicom_series = all(
+                    source_file["file_name"].lower().endswith(".dcm")  # type: ignore
```
dorfmanrobert, line 680: is there a reason we can't or shouldn't check this with the type?

darwin/dataset/download_manager.py
```diff
@@ -670,9 +674,44 @@ def _get_planned_image_paths(
...
+                )
+                is_extracted_frame = (
+                    len(slot.source_files) == 2
+                    and any(
+                        source_file["file_name"].lower().endswith(ext)  # type: ignore
+                        for ext in SUPPORTED_VIDEO_EXTENSIONS
+                        for source_file in slot.source_files
+                    )
+                    and any(
+                        source_file["file_name"].lower().endswith(ext)  # type: ignore
```
dorfmanrobert, line 691: could loop over SUPPORTED_IMAGE_EXTENSIONS and SUPPORTED_VIDEO_EXTENSIONS in the same any

dorfmanrobert (approved):

darwin/dataset/download_manager.py
JBWilkie: Spoke in-person about this - Because we store `NifTI` files as type `dicom`

PR #973, opened 2024-11-22
[DAR-4940][External] - Remove the need to pass the `legacy` flag when importing and converting `NifTI` annotations

JBWilkie:
# Problem
Currently, medical files in the platform may be divided into two groups:
- 1: Files that were isotropically scaled upon upload - These are all files uploaded with the `MED_2D_VEWER` FF disabled
- 2: Files that were not isotropically scaled upon upload - These are all files uploaded with the `MED_2D_VIEWER` FF enabled

Now that `MED_2D_VIEWER` is globally enabled, all further uploaded files will not be isotropically scaled

When importing `NifTI` annotations to these items, or when converting their annotations to the `NifTI` format, we need to apply exactly the same scaling as was applied to the item itself upon upload. Currently, we control this with the `legacy` flag, but it's very manual and can be tricky for users to manage

# Solution
This PR removes the need to pass the `legacy` flag in both the import and convert flows by:
- Importing annotations: Using the API to lookup the state of each file targeted by the import. If it was scaled, we apply scaling to the imported annotations, and vice versa
- Converting annotations: We use the presence of a newly added Darwin JSON medical metadata field `"handler": "MONAI"` to determine if a file was scaled or not, and apply scaling during the conversion if required

After this is merged, we'll release a new darwin-py version and update backend with it. This will enable backend to automatically scale exported `NifTI` annotations per-file, and then we'll fully remove the `legacy` flag and all related code some weeks in the future

# Changelog
Removed the need to pass a flag when importing or converting `NifTI` annotations. Instead, we handle it automatically, significantly reducing confusion and the likelihood of errors


darwin/exporter/formats/nifti.py
```diff
@@ -70,14 +83,20 @@ def export(
...
-            style="bold blue",
+            "The `legacy` flag is now non-functional and will be deprecated soon. The annotation import process now automatically detects if legacy annotation scaling is required.",
+            style="warning",
         )
 
     video_annotations = list(annotation_files)
     for video_annotation in video_annotations:
+        try:
+            medical_metadata = video_annotation.slots[0].metadata
+            legacy = not medical_metadata.get("handler") == "MONAI"  # type: ignore
```
dorfmanrobert, line 96: I think we also need to check if `is_volume` is true

darwin/importer/importer.py
```diff
@@ -1206,10 +1225,23 @@ def import_annotations(  # noqa: C901
     console.print("Retrieving local annotations ...", style="info")
     local_files = []
     local_files_missing_remotely = []
-
-    maybe_parsed_files: Optional[Iterable[dt.AnnotationFile]] = _find_and_parse(
-        importer, file_paths, console, use_multi_cpu, cpu_limit
-    )
+    print(importer)
```
dorfmanrobert, line 1228: extra print

darwin/options.py
```diff
@@ -416,9 +420,11 @@ def cpu_default_types(input: Any) -> Optional[int]:  # type: ignore
...
         parser_convert.add_argument(
             "--legacy",
-            action="store_true",
-            default=False,
-            help="Import annotation files using legacy process (isotropic transformation).",
+            type=lambda x: (str(x).lower() == "true") if x is not None else None,
+            nargs="?",
+            const=True,
+            default=None,
+            help="Import annotation files using legacy process (isotropic transformation). Pass 'true' to enable, or omit for None.",
```
dorfmanrobert, line 427: Think this should be aligned w the console logs that say

`"The `legacy` flag is now non-functional and will be deprecated soon. The annotation import process now automatically detects if legacy annotation scaling is required.",`

darwin/exporter/formats/nifti.py
JBWilkie: I actually don't think so. I tested, and files handles as slices return `"handler": null`

JBWilkie: Notes from our in-person conversation - These are the scenarios:

Before `MED_2D_VIEWER`:
- Not `MONAI`: Files are scaled --> It's possible to upload unscaled files, but these are files uploaded as slices (not volumes), which NifTI annotations cannot be exported from

After `MED_2D_VIEWER`:
- `MONAI`: Files are guaranteed to be unscaled
- Not `MONAI`: Files are scaled --> It's possible to upload unscaled files, but these are files uploaded as slices (not volumes), which NifTI annotations cannot be exported from

dorfmanrobert (approved):

PR #979, opened 2024-12-11
[DAR-4924][External] Resolving issues with import & export of `NifTI` annotations

JBWilkie:
# Problem
This PR tackles 3 issues related to importing & exporting annotations in the `NifTI` format. Issue 1 is major, issues 2 & 3 are minor
- 1: `NifTI` annotations are being incorrectly oriented upon import and export in some situations
- 2: It is not possible to import `NifTI` annotations to items in dataset folders
- 3: When importing `NifTI` annotations as polygons, the last frame is always omitted

# Solution
To tackle these problems:
- 1: We were re-orienting `NifTI` annotations incorrectly, and in places we didn't need to be. This PR changes the logic to be the following:

**When importing `NifTI` annotations:** If importing to a post `MED_2D_VIEWER` file, re-orient to `LPI` according to `nibabel`. This is the exact same logic we apply to medical file processing in the backend. Otherwise, if importing to a pre `MED_2D_VIEWER` file:
- If it's a `NifTI` dataset item: Re-orient to `LPI`, since all pre `MED_2D_VIEWER` `NifTI` files were re-oriented this way
- If it's a DICOM dataset item, look up current affine of the target dataset item, and re-orient the annotations based on that

**When exporting `NifTI` annotations:** Always re-orient the exported `NifTI` file based on the original affine of the dataset item it came from

- 2: Allow import of `NifTI` annotations to items in dataset folders 
- 3 Resolves the issue with removing the last frame of `NifTI` polygon annotations

**IMPORTANT NOTE:** Due to what I believe were some issues in the pre `MED_2D_VIEWER` backend processing pipeline, we may have incorrectly stored affines and original affines for some pre `MED_2D_VIEWER` files. Since this is all we have to go off when re-orienting `NifTI` annotations, there may be issues in these cases, and we should keep an eye out for them

# Changelog
Resolved issues with automatic scaling & orientation of `NifTI` annotations


darwin/exporter/formats/nifti.py
```diff
@@ -489,18 +495,10 @@ def unnest_dict_to_list(d: Dict) -> List:
     volumes = unnest_dict_to_list(output_volumes)
     for volume in volumes:
         img = nib.Nifti1Image(
-            dataobj=np.flip(volume.pixel_array, (0, 1, 2)).astype(np.int16),
+            dataobj=volume.pixel_array.astype(np.int16),
             affine=volume.affine,
         )
-        if legacy and volume.original_affine is not None:
```
JBWilkie, line 495: Until this PR, this line was flipping the orientation of every axis in every `NifTI` export

darwin/importer/formats/nifti.py
```diff
@@ -78,16 +76,20 @@ def parse_path(
         return None
     annotation_files = []
     for nifti_annotation in nifti_annotations:
-        legacy = nifti_annotation["image"] in remote_files_that_require_legacy_scaling
+        remote_file_path = Path(nifti_annotation["image"])
+        if not str(remote_file_path).startswith("/"):
+            remote_file_path = Path("/" + str(remote_file_path))
```
JBWilkie, line 81: Allow import of `NifTI` annotation to files in dataset folders

darwin/importer/formats/nifti.py
```diff
@@ -353,7 +362,7 @@ def nifti_to_video_polygon_annotation(
     if len(all_frame_ids) == 1:
         segments = [[all_frame_ids[0], all_frame_ids[0] + 1]]
     elif len(all_frame_ids) > 1:
-        segments = [[min(all_frame_ids), max(all_frame_ids)]]
+        segments = [[min(all_frame_ids), max(all_frame_ids) + 1]]
     video_annotation = dt.make_video_annotation(
```
JBWilkie, line 366: 1 line bugfix to resolve issue where imported `NifTI` polygon annotations longer than 1 frame would always omit the last frame

darwin/importer/formats/nifti.py
```diff
@@ -513,16 +522,22 @@ def correct_nifti_header_if_necessary(img_nii):
...
+    remote_files_that_require_legacy_scaling: Dict[Path, Dict[str, Any]] = {},
 ) -> Tuple[np.ndarray, Tuple[float]]:
     """
     Converts a nifti object of any orientation to the passed ornt orientation.
     The default ornt is LPI.
 
     Args:
         input_data: nibabel nifti object.
-        ornt: (n,2) orientation array. It defines a transformation from RAS.
+        ornt: (n,2) orientation array. It defines a transformation to LPI
```
JBWilkie, line 534: `nibabel` and `ITK` disagree on what orientation is represented by `LPI`, `RAS`, etc. We process `NifTI` annotations exclusively with `nibabel`, which represents `LPI` as:

`[[0.0, -1.0], [1.0, -1.0], [2.0, -1.0]]`

darwin/importer/formats/nifti.py
```diff
@@ -531,9 +546,13 @@ def process_nifti(
     img = correct_nifti_header_if_necessary(input_data)
     orig_ax_codes = nib.orientations.aff2axcodes(img.affine)
     orig_ornt = nib.orientations.axcodes2ornt(orig_ax_codes)
+    if remote_file_path in remote_files_that_require_legacy_scaling:
+        slot_affine_map = remote_files_that_require_legacy_scaling[remote_file_path]
+        affine = slot_affine_map[next(iter(slot_affine_map))]  # Take the 1st slot
```
JBWilkie, line 551: There was work done in `MED-13` to allow export of `NifTI` annotations from multi-slotted items, but it has never been possible to import `NifTI` annotations to multi-slotted items. We therefore take the 1st slot in this case

darwin/exporter/formats/nifti.py
```diff
@@ -489,18 +495,10 @@ def unnest_dict_to_list(d: Dict) -> List:
     volumes = unnest_dict_to_list(output_volumes)
     for volume in volumes:
         img = nib.Nifti1Image(
-            dataobj=np.flip(volume.pixel_array, (0, 1, 2)).astype(np.int16),
+            dataobj=volume.pixel_array.astype(np.int16),
             affine=volume.affine,
         )
```
JBWilkie, line 500: Until this PR, this line was flipping the orientation of every axis in every NifTI export

darwin/exporter/formats/nifti.py
```diff
@@ -510,6 +508,27 @@ def unnest_dict_to_list(d: Dict) -> List:
...
+
+    Parameters
+    ----------
+    img: nib.Nifti1Image
+        The NIfTI image to be reoriented
+    volume: Dict
+        The volume containing the affine and original affine
+    """
+    if volume.original_affine is not None:
+        img_ax_codes = nib.orientations.aff2axcodes(volume.affine)
```
dorfmanrobert, line 523: I don't think this is correct for pre MED_2D -- the volume.affine is (expected to be at least) RAS

So this will try to go from RAS -> original orientation.

But really its meant to go from LPI -> original orientation. I think this is why `dataobj=np.flip(volume.pixel_array, (0, 1, 2)).astype(np.int16)` used to be invoked: to go from LPI to RAS so that this logic would make sense for pre MED_2D files

darwin/importer/formats/nifti.py
```diff
@@ -531,9 +549,16 @@ def process_nifti(
...
     orig_ax_codes = nib.orientations.aff2axcodes(img.affine)
     orig_ornt = nib.orientations.axcodes2ornt(orig_ax_codes)
+    if remote_file_path in remote_files_that_require_legacy_scaling:
+        is_dicom = remote_file_path.suffix.lower() == ".dcm"
+        slot_affine_map = remote_files_that_require_legacy_scaling[remote_file_path]
+        affine = slot_affine_map[next(iter(slot_affine_map))]  # Take the 1st slot
+        ax_codes = nib.orientations.aff2axcodes(affine)
+        ornt = nib.orientations.axcodes2ornt(ax_codes)
+        if not is_dicom:
+            img = nib.Nifti1Image(np.flip(img.get_fdata(), (0, 1, 2)), affine)
```
dorfmanrobert, line 559: I think it might be more robust to do `as_closest_canonical` on `img` to get it to RAS, since this will only work if the Nifti annotation file is in LPI right

dorfmanrobert:
Having some comments on why we need these transforms for pre MED_2D would be helpful for future ref I think

darwin/importer/formats/nifti.py
JBWilkie: Updated

darwin/exporter/formats/nifti.py
JBWilkie: Updated

dorfmanrobert (approved):
nice work!

PR #980, opened 2024-12-12
[DAR-5390][External] Add 'extract video-artifacts' command

AndriiKlymchuk:
# Problem
For read only registration of videos in SECTIONED_V2 format clients need to prepare all processing artifacts them-self.
It is very complicated as requires to perform a lot of steps. 

# Solution
Develop `extract video-artifacts` vommand that creates all artifacts and generates metadata file containing payload for registration.


AndriiKlymchuk:
> Looks good, and works well when I test it. To get into the repo I think we should consider:
> 
> * Allowing the script to process multiple videos from a single command, perhaps it could work on a directory of files
> * Adding an explicit warning / documentation stating that this may become out of sync with our backend video processing pipeline
> * Placing it in a new `darwin/scripts` directory or similar, as no scripts of this type exist in darwin-py currently

Agree. Also we need to cover it by tests.
I also sync maybe it shouldn't be a separate script, but rather part of darwin py cli?

JBWilkie (approved):

JBWilkie:
@AndriiKlymchuk Only issue is that `ffmpeg@5` will be missing from the CI environment. Perhaps we can add a command [here](https://github.com/v7labs/darwin-py/blob/90c3fc48739083b9352017e37204ff6ffd786f7e/.github/workflows/JOB_tests.yml#L45) to install it?

PR #989, opened 2025-01-02
[DAR-5342][External] Allow import of complex polygons in `COCO`

JBWilkie:
# Problem
COCO segmentations can have >1 polygon (an array of arrays containing `x, y`). Currently we only select the first element

# Solution
Allow import of multiple segmentation array elements as a single polygon. This effectively adds support for importing complex polygons

# Changelog
Allowed import of multi-array COCO segmentations

### DIFF
--- a/darwin/importer/formats/coco.py
+++ b/darwin/importer/formats/coco.py
@@ -88,7 +88,7 @@ def parse_json(
         annotation["segmentation"]
         if image_id not in image_annotations:
             image_annotations[image_id] = []
-        image_annotations[image_id].append(
+        image_annotations[image_id].extend(
             parse_annotation(annotation, category_lookup_table)
         )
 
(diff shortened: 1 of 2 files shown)


dorfmanrobert (commented):
I would think the opposite assumption would make most sense here: that users expect an instance mask to be treated as a single complex polygon rather than separate polygon annotations. Since we only hit this with `is_crowd=0`, my understanding is that this would mean the masks are components of the same instance

Is there a reason we are assuming users want multiple polygon annotations for each segmentation that has disconnected components?

JBWilkie:
> I would think the opposite assumption would make most sense here: that users expect an instance mask to be treated as a single complex polygon rather than separate polygon annotations
> 
> Is there a reason we are assuming users want multiple polygon annotations for each segmentation that has disconnected components?

Yes - [In the COCO format](https://cocodataset.org/#format-data), each polygon represented by coordinates in `segmentation` is treated as an individual polygon

Complex polygons are represented by RLE, instead of series of `(x, y)` coordinates

dorfmanrobert (approved):

PR #990, opened 2025-01-08
[DAR-5371] Removed NifTI export `pixdim` scaling

JBWilkie:
# Problem
When exporting annotations from medical files, we need to perform a transformation from millimetre space to pixel space by dividing annotation coordinates by the appropriate `pixdim` values based on the primary plane. The problem is that currently, we only apply this transformation for `NifTI` exports, and we want to apply it to all annotation formats. Additionally, the existing transformation does not take the primary plane into account. We assume that the file was axially-acquired since we use `pixdim[0]` and `pixdim[1]`

# Solution
This PR removes the `pixdim` scaling from the `NifTI` exporter. It is being re-applied to all appropriate medical files in backend as part of DAR-5339, where we will introduce handling for non-axially-acquired images as well

There is also an unrelated change here where we remove the `RemoteDataset.reset` method since it has been deprecated since the V1 --> V2 Darwin migration

# Changelog
Removed `pixdim` scaling from the `NifTI` exporter as this will be placed into backend directly


darwin/cli_functions.py
```diff
@@ -1075,8 +1075,6 @@ def set_file_status(dataset_slug: str, status: str, files: List[str]) -> None:
         )
         if status == "archived":
             dataset.archive(items)
-        elif status == "clear":
-            dataset.reset(items)
```
JBWilkie, line 1079: This is the unrelated change to remove a deprecated method

darwin/dataset/remote_dataset.py
```diff
@@ -467,68 +467,57 @@ def fetch_remote_files(
...
-
-        Parameters
-        ----------
-        items : Iterator[DatasetItem]
-            The ``DatasetItem``\\s to be reset.
-        """
-
-    @abstractmethod
-    def complete(self, items: Iterator[DatasetItem]) -> None:
+    def complete(self, items: Iterable[DatasetItem]) -> None:
```
JBWilkie, line 503: This is the unrelated change to remove a deprecated method

tests/darwin/cli_functions_test.py
```diff
@@ -289,25 +289,6 @@ def test_calls_dataset_archive(
...
-                with patch.object(RemoteDatasetV2, "reset") as mock:
-                    set_file_status(dataset_identifier, "clear", ["one.jpg", "two.jpg"])
-                    get_remote_dataset_mock.assert_called_once_with(
-                        dataset_identifier=dataset_identifier
-                    )
-                    fetch_remote_files_mock.assert_called_once_with(
-                        {"item_names": "one.jpg,two.jpg"}
-                    )
-                    mock.assert_called_once_with(fetch_remote_files_mock.return_value)
-
```
JBWilkie, line 310: This is the unrelated change to remove a deprecated method

darwin/dataset/remote_dataset_v2.py
```diff
@@ -449,13 +438,13 @@ def complete(self, items: Iterator[DatasetItem]) -> None:
             team_slug=self.team,
         )
 
-    def delete_items(self, items: Iterator[DatasetItem]) -> None:
+    def delete_items(self, items: Iterable[DatasetItem]) -> None:
```
JBWilkie, line 441: Changing the type of `items` from `Iterator` to `Iterable` makes these methods more flexible

darwin/exporter/formats/nifti.py
```diff
@@ -549,7 +548,7 @@ def shift_polygon_coords(
         else:
             return [{"x": p["y"], "y": p["x"]} for p in polygon]
     else:
-        return [{"x": p["y"] // pixdim[1], "y": p["x"] // pixdim[0]} for p in polygon]
+        return [{"x": p["y"], "y": p["x"]} for p in polygon]
```
JBWilkie, line 551: This is the consequential change in this PR

PR #992, opened 2025-01-15
[DAR-5421] Support for non-axial NifTI annotation imports and exports

JBWilkie:
# Problem
Currently when importing or exporting annotations in the `NifTI` format, it's assumed the primary plane of the target slot axial. This is not always the case. Assuming an `AXIAL` primary plane is problematic because incorrect `pixdim` values are selected for scaling and slicing

# Solution
This PR makes changes to take the primary plane of the target slot into account when scaling and slicing `NifTI` annotations on import and export

# Changelog
Added support for importing and exporting non-axial `NifTI` volumes as annotations


darwin/importer/formats/nifti.py
```diff
@@ -192,30 +198,53 @@ def get_polygon_video_annotations(
...
+            if slot_names:
+                slot_name = slot_names[0]
+            else:
+                slot_name = list(item_pixdims_and_primary_planes.keys())[0]
+            slot_pixims_and_primary_plane = item_pixdims_and_primary_planes[slot_name]
+            primary_plane = slot_pixims_and_primary_plane[1]
+            pixdims = slot_pixims_and_primary_plane[0]
+        else:
+            pixdims = [1.0, 1.0, 1.0]
+            primary_plane = "AXIAL"
```
JBWilkie, line 219: This allows `NifTI` annotations to be imported to non-medical files. In these cases, we won't do any scaling

When importing to medical files, we import to the specified slot. If the slot isn't specified, we import to the first medical slot in the dataset item

darwin/importer/formats/nifti.py
```diff
@@ -227,19 +256,31 @@ def get_mask_video_annotations(
...
+        if slot_names:
+            slot_name = slot_names[0]
+        else:
+            slot_name = list(item_pixdims_and_primary_planes.keys())[0]
+        slot_pixdims_and_primary_plane = item_pixdims_and_primary_planes[slot_name]
+        primary_plane = slot_pixdims_and_primary_plane[1]
+        pixdims = slot_pixdims_and_primary_plane[0]
+    else:
+        pixdims = [1.0, 1.0, 1.0]
+        primary_plane = "AXIAL"
```
JBWilkie, line 279: This allows `NifTI` annotations to be imported to non-medical files. In these cases, we won't do any scaling

When importing to medical files, we import to the specified slot. If the slot isn't specified, we import to the first medical slot in the dataset item

dorfmanrobert (approved):

PR #995, opened 2025-01-30
Adding codespaces config

umbertoDifa:


.devcontainer/setup.sh
```diff
@@ -0,0 +1,60 @@
...
+source /workspaces/darwin-py/.venv/bin/activate #poetry shell
+poetry install
+
+# Install additional Python dependencies
+pip install darwin-py[test]
+pip install darwin-py[ml]
+pip install darwin-py[ocv]
+
+pip install connected-components-3d
+pip install nibabel
```
JBWilkie, line 43: I'd replace this with:

```suggestion
# Install all dependencies
poetry install --all-extras
```

Less lines, and installs dependencies from all dependency groups

JBWilkie (approved):
Tested and works well

1 non-blocking suggestion

PR #996, opened 2025-01-30
[DAR-5614][External] Make the `NifTI` importer re-orient files to orientation of target file instead of `LPI`

JBWilkie:
# Problem
Since we released the `DISABLE_REORIENTATION_ON_UPLOAD` feature flag, not every post-`MED_2D_VIEWER` file will be in the `LPI` orientation, but the current `NifTI` importer assumes they are

This ticket is to introduce logic into the `NifTI` importer similar to the export logic so that we re-orient inbound annotations to the target file instead of statically to `LPI`

# Solution
Re-orient inbound `NifTI` annotations to the orientation of the remote file it's targeting

# Changelog
Support for importing `NifTI` annotations to non-`LPI` oriented dataset files

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@
 
 Darwin-py can both be used from the [command line](#usage-as-a-command-line-interface-cli) and as a [python library](#usage-as-a-python-library).
 
+📚 Check out the complete [SDK Reference Documentation](https://darwin-py-sdk.v7labs.com/) for detailed API information.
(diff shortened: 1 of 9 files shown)


JBWilkie:
Plans have changed - We are going to remove the `DISABLE_REORIENTATION_ON_UPLOAD` feature flag, so this change is not necessary

PR #1008, opened 2025-02-20
[DAR-5588][External] Apply automated triaxial re-orientation of non-NifTI annotations for import

JBWilkie:
# Problem
When importing non-NifTI annotations, we do not consider if the target dataset items were medical items with flipped X, Y, or Z axes. Non-NifTI annotations generated for items that were re-oriented (All non `LPI` files) out of the platform will appear flipped in at least one axis when imported

# Solution
This PR:
- 1: Determines the set of axes that were flipped on upload for items targeted by every annotation import. We do this by comparing `affine` with `original_affine` in the [same way](https://github.com/v7labs/darwin-py/blob/90c3fc48739083b9352017e37204ff6ffd786f7e/darwin/exporter/formats/nifti.py#L500) that we already compute transforms for NifTI exports
- 2: Applies the necessary X, Y, and Z axial flips to inbound annotations:
    - X & Y are flipped by subtracting their current coordinates from slot width & height
    - Z is flipped (only for sequential items) by reversing slot frame indices

Raster annotations require alternative handling to flip in X & Y. This requires decoding the `dense_rle`, flipping the decoded data, then re-encoding it into another `dense_rle`

We only apply these flips to post-`MED_2D_VIEWER` files. This PR also takes the opportunity to refactor `_get_remote_medical_file_transform_requirements` into the simplier `_get_remote_file_medical_metadata` which returns a single more understandable dictionary of data

This cannot be released until DAR-5589 is ready to merge on the backend

# Changelog
Applied re-orientation of non-NifTI annotations upon import to medical dataset slots


darwin/importer/importer.py
```diff
@@ -2488,3 +2521,108 @@ def slot_is_medical(slot: Dict[str, Any]) -> bool:
 
 def slot_is_handled_by_monai(slot: Dict[str, Any]) -> bool:
     return slot.get("metadata", {}).get("medical", {}).get("handler") == "MONAI"
+
+
+def _apply_axial_flips_to_annotations(
```
dorfmanrobert, line 2526: I have concerns abt the correctness of this logic 

We set `axial_flips = nib.orientations.ornt_transform(original_ornt, ornt)[:, 1]`. This means if we have a file that is uploaded as LIP and annotated w polygons, won't we be applying a flip that should be applied to a different axis?

In the case of Nifti imports this seems to be handled by the view_idx that is based on what the primary_plane is, but here it seems we are ignoring if the axes were reordered

darwin/importer/importer.py
```diff
@@ -2488,3 +2519,167 @@ def slot_is_medical(slot: Dict[str, Any]) -> bool:
...
+) -> List[int]:
+    """
+    Returns a list of 3 values that are either 1 or -1.
+    -1 indicates that the corresponding slot axis was mirrored on upload.
+
+    Takes the primary plane of the slot into account, since re-orienting medical files to
+    `LPI` will result in axial re-ordering for coronally and sagitally oriented files:
+    - Axially acquired files axial codes are ordered: L/R, P/A, I/S
+    - Coronally acquired files axial codes are ordered: L/R, I/S, P/A
+    - Sagittally acquired files axial codes are ordered: I/S, L/R, P/A
```
dorfmanrobert, line 2535: This is true for DICOM, but for Nifti if we are using Nibabel to decide the primary plane, I'm not sure that this is enforced to be true (this was the diff in that other PR on orig implementation vs. nibabel on deciding what hte primary plane was for Niftis)

Rather than
```
    if primary_plane == "CORONAL":
        axial_flips[0], axial_flips[1], axial_flips[2] = (
            axial_flips[0],
            axial_flips[2],
            axial_flips[1],
        )
    elif primary_plane == "SAGITTAL":
        axial_flips[0], axial_flips[1], axial_flips[2] = (
            axial_flips[2],
            axial_flips[0],
            axial_flips[1],
        )
```
        
It might be safer to use the first column values of the nib.orientations.ornt_transform(original_ornt, ornt)

darwin/datatypes.py
```diff
@@ -244,6 +244,116 @@ def get_sub(self, annotation_type: str) -> Optional[SubAnnotation]:
                 return sub
         return None
 
+    def _flip_annotation_in_x_or_y(self, axis_limit: int, axis: CartesianAxis):
```
dorfmanrobert, line 247: Would deduplicate this and remove all the if else statements. One way to do it is to do two fns since we already decide on what it is based on an if/else when it is used in:
`_flip_annotation_in_x`
`_flip_annotation_in_y`
and those both call a shared helper _flip_annotation_coordinates that just changes to either use key "x" or "y" based on which of `_flip_annotation_in_x`/`_flip_annotation_in_y` is calling it

darwin/datatypes.py
```diff
@@ -244,6 +244,116 @@ def get_sub(self, annotation_type: str) -> Optional[SubAnnotation]:
...
+
+        elif annotation_type == "skeleton":
+            if axis == CartesianAxis.X:
+                for node in self.data["nodes"]:
+                    node["x"] = axis_limit - node["x"]
+            elif axis == CartesianAxis.Y:
+                for node in self.data["nodes"]:
+                    node["y"] = axis_limit - node["y"]
+
+    def _flip_raster_layer_in_x_or_y(
```
dorfmanrobert, line 306: Maybe you can put the raster logic  as another case you are matching against in `flip_annotation_coords/flip_annotation_coords` helper rather than having a separate method? and in that case you call the logic from another helper to declutter as well.

darwin/datatypes.py
```diff
@@ -1567,3 +1677,9 @@ class StorageKeyDictModel(BaseModel):
 
 class StorageKeyListModel(BaseModel):
     storage_keys: List[str]
+
+
+class CartesianAxis(Enum):
```
dorfmanrobert, line 1682: If take approach suggested above, can remove

darwin/importer/importer.py
```diff
@@ -2488,3 +2519,167 @@ def slot_is_medical(slot: Dict[str, Any]) -> bool:
...
+                    if axial_flips[1] == -1:
+                        annotation._flip_annotation_in_x_or_y(
+                            axis_limit=slot_metadata["height"],
+                            axis=dt.CartesianAxis.Y,
+                        )
+
+    return parsed_files
+
+
+def _flip_annotation_in_z(
```
dorfmanrobert, line 2666: This should maybe be a method of Annotation as well?

darwin/importer/importer.py
```diff
@@ -2488,3 +2519,167 @@ def slot_is_medical(slot: Dict[str, Any]) -> bool:
...
+            axial_flips[2],
+            axial_flips[1],
+        )
+    elif primary_plane == "SAGITTAL":
+        axial_flips[0], axial_flips[1], axial_flips[2] = (
+            axial_flips[2],
+            axial_flips[0],
+            axial_flips[1],
+        )
+    return axial_flips
```
dorfmanrobert, line 2580: Should we call these axial_flips still? It might also make it clearer if it was some kind of mapping

{"Coronal": -1,
"Sagittal": 1,
"Axial": -2}

so that its use in _apply_axial_flips_to_annotations is clearer than axial_flips[index], which is hard to read

darwin/importer/importer.py
JBWilkie: Resolved

darwin/importer/importer.py
```diff
@@ -2488,3 +2519,167 @@ def slot_is_medical(slot: Dict[str, Any]) -> bool:
...
+            for annotation in parsed_file.annotations:
+                annotation_slot_name = annotation.slot_names[0]
+                if annotation_slot_name not in medical_metadata:
+                    continue
+
+                slot_metadata = medical_metadata[annotation_slot_name]
+                if slot_metadata["legacy"]:
+                    continue
+
+                axial_flips = slot_metadata["axial_flips"]
```
dorfmanrobert, line 2602: The logic below 
1) Should probably be deduplicated out into helper functions

2) If the changes suggested for `_flip_annotation_in_x_or_y` and `_flip_raster_layer_in_x_or_y` are taken, then can remove a lot of the if elses here

3) Can we also loop over frame_idx, frame_annotation once? aren't the flips order independent?

darwin/importer/importer.py
JBWilkie: I'll make it a method of `VideoAnnotation`

PR #1009, opened 2025-02-25
[DAR-5929] Allow annotation and text properties import

aleksandar-ivic:
# Problem
Currently importing annotation or section text properties fails.

# Solution
Add support for annotation and section text properties. Expand e2e tests to cover this.

# Changelog
Allow annotation and text properties import


darwin/importer/importer.py
```diff
@@ -1962,6 +1986,8 @@ def _import_annotations(
     payload["overwrite"] = _get_overwrite_value(append)
 
     try:
+        console = Console(theme=_console_theme())
+        console.print(f"### Payload {payload}", style="info")
```
JBWilkie, line 1990: Let's drop this debugging print

JBWilkie (commented):
I'd add text properties to 1 additional E2E test: `test_annotation_classes_are_created_with_properties_on_import`

This creates a fresh ontology of classes and properties from local annotation files and an accompanying `metadata.json` file, then runs the standard import test. This will ensure we can create new text properties from manifest files. To do this, we'll need to:
- 1: Add text property values to some JSON files in the `e2e_tests/data/import/image_new_annotations_with_properties` directory
- 2: Add a text property to the `.v7/metadata.json` file in the same directory, such as:

```
      "properties": [
        {
          "name": "ms1",
          "type": "multi_select",
          "description": "",
          "required": true,
          "property_values": [
            {
              "value": "1",
              "color": "rgba(173,255,0,1.0)"
            },
            {
              "value": "2",
              "color": "rgba(143,255,0,1.0)"
            }
          ]
        },
        {
          "name": "my_text_prop",
          "type": "text",
          "description": "Write some text",
          "required": false,
          "property_values": []
        },
```

JBWilkie (approved):

darwin/importer/importer.py
```diff
@@ -623,6 +623,11 @@ def _import_properties(
                     t_prop: FullProperty = team_properties_annotation_lookup[
                         (a_prop.name, annotation_class_id)
                     ]
+                    if t_prop.type == "text":
+                        annotation_property_map[annotation_id][str(a_prop.frame_index)][
+                            t_prop.id
+                        ] = [{"text": a_prop.value}]
+                        continue
```
umbertoDifa, line 630: Basically this simulates the fact that we 'found' a text property with that specific arbitrary value in the .metadata so that it does not complain that it's missing?

darwin/importer/importer.py
```diff
@@ -759,14 +764,18 @@ def _import_properties(
...
             # check if property value is different in m_prop (.v7/metadata.json) options
-            for m_prop_option in m_prop_options:
-                if m_prop_option.get("value") == a_prop.value:
-                    break
-            else:
-                if a_prop.value:
-                    raise ValueError(
-                        f"Annotation: '{annotation_name}' -> Property '{a_prop.value}' not found in .v7/metadata.json, found: {m_prop.property_values}"
-                    )
+            if m_prop.type != "text" and m_prop.granularity in [
```
umbertoDifa, line 767: This skips the check in case it's a text property. Why do we also need to check the granularity if we werent' before?

umbertoDifa (commented):
LGTM a part from the ugly size of what this import has become which isn't the purpose of this PR anyways.

A few tests we should have:
- Have we tested the scenario when we set a text section-level property to empty?
- Have we tested the scenario when we set a text annotation-level property to empty?
- What happens if we have text section level property with the same name of text annotation level property?
- What happens if the have a text annotation level property and an item level property with the same name?

darwin/importer/importer.py
aleksandar-ivic: no need for granularity, I will remove it.

darwin/importer/importer.py
aleksandar-ivic: This tells us that we found actual property on annotation and we add it to the list that will be imported later on.

darwin/importer/importer.py
```diff
@@ -782,43 +791,53 @@ def _import_properties(
...
             assert t_prop.id is not None
             assert t_prop_val.id is not None
-            annotation_property_map[annotation_id][str(a_prop.frame_index)][
-                t_prop.id
-            ].add(t_prop_val.id)
+
+            if t_prop.type == "text":
+                annotation_property_map[annotation_id][str(a_prop.frame_index)][
+                    t_prop.id
+                ] = [{"text": a_prop.value}]
```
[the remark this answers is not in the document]
aleksandar-ivic: I will move it to the function.

aleksandar-ivic:
> * What happens if we have text section level property with the same name of text annotation level property?



> LGTM a part from the ugly size of what this import has become which isn't the purpose of this PR anyways.
> 
> A few tests we should have:
> 
> * Have we tested the scenario when we set a text section-level property to empty?
> * Have we tested the scenario when we set a text annotation-level property to empty?
> * What happens if we have text section level property with the same name of text annotation level property?
> * What happens if the have a text annotation level property and an item level property with the same name?

- Empty values fail, I will check that now. Thanks for pointing it out
- Section and annotation level properties can't have the same name, this is forbidden on API level.
- Same name for `item` and `section`/`annotation` level props is allowed, I've added one test case to cover this.

e2e_tests/cli/test_import.py
```diff
@@ -1,5 +1,7 @@
 from pathlib import Path
 
+from cv2 import exp
```
aleksandar-ivic: This is accidental, I've removed in later commit

darwin/importer/importer.py
```diff
@@ -782,43 +791,50 @@ def _import_properties(
...
-            annotation_property_map[annotation_id][str(a_prop.frame_index)][
-                t_prop.id
-            ].add(t_prop_val.id)
+
+            if t_prop.type == "text":
+                set_text_property_value(
+                    annotation_property_map, annotation_id, a_prop, t_prop
+                )
+            else:
+                assert t_prop_val.id is not None
```
aleksandar-ivic, line 834: This was moved here as it only applies to `single-select` and `multi-select`

darwin/importer/importer.py
```diff
@@ -473,24 +474,32 @@ def _serialize_item_level_properties(
                 None,
             )
             value = {"id": item_property_value_id}
```
AndriiKlymchuk, line 476: Should we also skip appending if value is None here?

AndriiKlymchuk: If so, we can just check if `value` variable is not `None` instead of introducing  `append_property` variable

aleksandar-ivic: I haven't found a case where that value is null

AndriiKlymchuk:
Can we add some tests to ensure fixes work as expected? Or the are already there and just data were missing?

aleksandar-ivic:
> Can we add some tests to ensure fixes work as expected? Or the are already there and just data were missing?

There are tests present, I've just expanded them with new data

AndriiKlymchuk (approved):

PR #1039, opened 2025-03-25
[DAR-6238][External] Allow to import annotations without pixdims, affine of plane_map

umbertoDifa:
# Problem
Some DICOM files might not have a defined pixdim value. That is the case for ultrasound images.

# Solution
Gracefully handle those scenarios by printing a warning and skipping pixdim calculations.

# Changelog
Fix bug blocking import for DICOMs without pixdims

### DIFF
--- a/darwin/importer/importer.py
+++ b/darwin/importer/importer.py
@@ -1,7 +1,7 @@
 import concurrent.futures
-import uuid
-import json
 import copy
+import json
+import uuid
 from collections import defaultdict
 from logging import getLogger
 from multiprocessing import cpu_count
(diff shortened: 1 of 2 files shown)


DawidMyslak (commented):
I'm reviewing the backend PR for the very first time (don't trust me), I see we're trying to build protection against missing pixdims or the plane map.

My only worry is this:
https://github.com/v7labs/darwin-frontend/blob/master/src/modules/Editor/MedicalMetadata.ts#L67

So if the `metadata.medical` exists on the slot, both `pixdims` and `plane_map` must be present with expected typing. We're creating a risk of errors like "Cannot read properties of undefined".

I think it would be good to default `pixdims` to be `[1, 1, 1]` and plane map to:
```
       plane_map = {
          [slot_name]: metadata.primary_plane || 'AXIAL',
        }
```

darwin/importer/importer.py
dorfmanrobert, line 1: Something like

```
        primary_plane = _parse_primary_plane(medical_metadata, slot_name, remote_file, console)
        pixdims = _parse_pixdims(medical_metadata, slot_name, remote_file, console)

        if primary_plane is None or pixdims is None:
            continue

        pixdim_plane_mappings[slot_name] = (pixdims, primary_plane)

        if not slot_is_handled_by_monai(slot):
            affine = _parse_affine(medical_metadata, slot_name, remote_file, console)
            if affine is not None:
                .....
```

would probably be more readable

dorfmanrobert (approved):

DawidMyslak (approved):

PR #1041, opened 2025-03-26
[Internal] Isolated E2E tests

umbertoDifa:
# Problem
E2E take 45min to run, which is painfully long to wait for.
E2E cannot be parallelized cause they all run in the same team thus rely on the state left by other tests.

# Solution
[Option 1] [John implemented test prefixes](https://github.com/v7labs/darwin-py/pull/1017/files) to ensure each test would rely on independent data (dataset, classes, etc)
[Option 2] I've implemented isolation by ensuring each e2e runs in a separate temporary team, following the pattern of our darwin-frontend.

**Option 1**
- Does not use fixture endpoints
- Less code changes
- E2E do not pass (I'm not sure why)

**Option 2**
- Uses fixture endpoints to create temporary teams, users, etc
- Each test (present and future) is guaranteed to be isolated
- Many code changes (yet only on tests not on the darwinpy code)

# Next steps
This PR by itself is semi-useless, as it doesn't change the problem and tests still run in 45min, yet it puts the base of parallelization by isolating each test to run in a team.
Given the setup in `Option 2` there are two possible ways to implement parallelization:
1. Add a `--team` flag to darwinpy CLI to execute commands within a specific team
2. Prefix each command with `darwin team <team_name> && <actual_command_to_run>` to ensure the command runs with the selected team

To be honest, none of these two options are easy and I'm not sure they'll work. The main issue is that darwinpy CLI executes commands based on the `default` selected team. Thus switching between one team and the other requires continuously switching the default team saved in `~/.darwin/config.yml` which ends up in race conditions.

If we approach the isolation with `Option 1` IMO it's much less resilient (we might have weird data conflicts) but the parallelization is easier cause all tests execute in the same team and we don't need to switch it.

# Changelog
Isolate E2E tests


e2e_tests/conftest.py
```diff
@@ -144,3 +93,60 @@ def local_dataset(
...
+
+        # Register the cleanup function to run even if test fails
+        request.addfinalizer(cleanup)
+
+        yield team_config
+
+    except Exception as e:
+        # If team creation fails, ensure we try to clean up if team was partially created
+        if team_data:
+            try:
```
umbertoDifa, line 148: This is the core fixture that creates an isolated team and deletes it after the test completes or fail

e2e_tests/fixtures/class_management.py
```diff
@@ -0,0 +1,321 @@
+from typing import Dict, List
```
umbertoDifa, line 1: All the files withing the `fixtures` folder are pure API helpers used to setup the temporary team, user, classes, etc

e2e_tests/fixtures/class_management.py
```diff
@@ -0,0 +1,321 @@
...
+from e2e_tests.exceptions import DataAlreadyExists, E2EException
+from e2e_tests.helpers import api_call, generate_random_string
+from e2e_tests.logger_config import logger
+from e2e_tests.objects import (
+    ConfigValues,
+    E2EAnnotationClass,
+    E2EItemLevelProperty,
+    TeamConfigValues,
+)
+
```
umbertoDifa, line 15: All the functions in this file were moved from `setup_tests.py` the file that used to setup the main team before. Now instead these functions run for each team creation (if needed to setup the data).

e2e_tests/helpers.py
```diff
@@ -281,3 +302,90 @@ def delete_annotation_uuids(
                 del annotation.frames[frame].id
                 if annotation.annotation_class.annotation_type == "raster_layer":
                     del annotation.frames[frame].data["mask_annotation_ids_mapping"]
+
+
+def api_call(
```
umbertoDifa, line 307: moved from `setup_tests.py`

e2e_tests/helpers.py
```diff
@@ -281,3 +302,90 @@ def delete_annotation_uuids(
...
+    headers = {"Authorization": f"ApiKey {api_key}"}
+    action = getattr(requests, verb)
+    if payload:
+        response = action(url, headers=headers, json=payload)
+    else:
+        response = action(url, headers=headers)
+    return response
+
+
+def generate_random_string(
```
umbertoDifa, line 342: moved from `setup_tests.py`

e2e_tests/helpers.py
```diff
@@ -281,3 +302,90 @@ def delete_annotation_uuids(
...
+
+    Returns
+    -------
+    str
+        The generated prefix, of length (length).  Matches [a-z0-9]
+    """
+    return "".join(random.choice(alphabet) for i in range(length))
+
+
+def create_random_image(
```
umbertoDifa, line 361: moved from `setup_tests.py`

e2e_tests/logger_config.py
```diff
@@ -0,0 +1,44 @@
+import logging
+from os import environ
+
+from dotenv import load_dotenv
+
+# Load environment variables from .env file
+load_dotenv()
+
+
+def setup_logger():
```
umbertoDifa, line 10: Added a nice logger to see exactly what happens in each test, for the sake of mental health

dorfmanrobert:
There seems to be mixed usage of 
- logging
- logger
- print

Should we choose one and make this consistent (at least in the new files)?

PR #1043, opened 2025-04-15
[DAR-6195][external] Support Reports v3 in Darwin Py

vvihorev:
# Problem
It is impossible to query V3 annotator reports using the SDK. Reports V1 can be queried, but are not created since 11/05/2024.

# Solution
Add a new top-level command and an action for V3 annotator reports: `darwin report annotators`.

Instead of requesting an annotator report synchronously, start a report job, and poll for results with exponential backoff. For now, log the id of the created job and raise an error if the happy path fails.

# Changelog
Added top-level `report` command, added `report annotators` command to query V3 annotator reports

### DIFF
--- a/darwin/cli.py
+++ b/darwin/cli.py
@@ -6,6 +6,7 @@ from argparse import ArgumentParser, Namespace
 from pathlib import Path
 
 import requests.exceptions
+from darwin.datatypes import AnnotatorReportGrouping
 from rich.console import Console
 
 import darwin.cli_functions as f
@@ -219,6 +220,19 @@ def _run(args: Namespace, parser: ArgumentParser) -> None:
             silent=args.silent,
             output=args.output,
         )
+
+    # Report generation commands
+    elif args.command == "report":
+        if args.action == "annotators":
+            f.report_annotators(
+                args.datasets,
+                args.start,
+                args.stop,
+                [AnnotatorReportGrouping(value) for value in args.group_by],
+                args.pretty,
+            )
+        elif args.action == "help" or args.action is None:
+            f.help(parser, "report")
 
 
 if __name__ == "__main__":
--- a/darwin/cli_functions.py
+++ b/darwin/cli_functions.py
@@ -1391,18 +1391,16 @@ def report_annotators(
         group_by,
     ).text
 
+    # the API does not return CSV headers if the report is empty
+    if not report:
+        report = "timestamp,dataset_id,dataset_name,dataset_slug,workflow_id,workflow_name,current_stage_id,current_stage_name,actor_id,actor_type,actor_email,actor_full_name,active_time,total_annotations,review_pass_rate,total_items_annotated,time_per_annotation,time_per_item\n"
+
     if not pretty:
-        # if no one worked in the report, we print nothing
         print(report)
         return
 
     lines: List[str] = report.split("\n")
     lines.pop(0)  # remove csv headers
-
-    if not lines:
-        console.print("No one has worked on this dataset yet!\n", style="success")
-        return
-
     lines.pop()  # remove last line, which is empty
 
     table: Table = Table(show_header=True, header_style="bold cyan")
--- a/darwin/cli_functions.py
+++ b/darwin/cli_functions.py
@@ -1372,7 +1372,7 @@ def report_annotators(
     pretty : bool
         If ``True``, it will print the output in a Rich formatted table.
     """
-    client: Client = _load_client(offline=True)
+    client: Client = _load_client()
     console = Console(theme=_console_theme())
 
     dataset_ids = []
@@ -1406,24 +1406,28 @@ def report_annotators(
     lines.pop()  # remove last line, which is empty
 
     table: Table = Table(show_header=True, header_style="bold cyan")
+
     table.add_column("Date")
-    table.add_column("Dataset Id", justify="right")
-    table.add_column("Dataset Name", justify="right")
-    table.add_column("Dataset Slug", justify="right")
-    table.add_column("Workflow Id", justify="right")
-    table.add_column("Workflow Name", justify="right")
-    table.add_column("Current Stage Id", justify="right")
-    table.add_column("Current Stage Name", justify="right")
-    table.add_column("User Id", justify="right")
-    table.add_column("User Type", justify="right")
-    table.add_column("Email", justify="right")
-    table.add_column("Full Name", justify="right")
-    table.add_column("Active Time", justify="right")
-    table.add_column("Total Annotations", justify="right")
-    table.add_column("Review Pass Rate", justify="right")
-    table.add_column("Total Items Annotated", justify="right")
-    table.add_column("Time Per Annotation", justify="right")
-    table.add_column("Time Per Item", justify="right")
+    for header in [
+        "Dataset Id",
+        "Dataset Name",
+        "Dataset Slug",
+        "Workflow Id",
+        "Workflow Name",
+        "Current Stage Id",
+        "Current Stage Name",
+        "User Id",
+        "User Type",
+        "Email",
+        "Full Name",
+        "Active Time",
+        "Total Annotations",
+        "Review Pass Rate",
+        "Total Items Annotated",
+        "Time Per Annotation",
+        "Time Per Item",
+    ]:
+        table.add_column(header, justify="right")
 
     for row in lines:
         table.add_row(*row.split(","))
--- a/darwin/cli_functions.py
+++ b/darwin/cli_functions.py
@@ -35,6 +35,7 @@ from darwin.dataset.split_manager import split_dataset
 from darwin.dataset.upload_manager import LocalFile
 from darwin.dataset.utils import get_release_path
 from darwin.datatypes import (
+    AnnotatorReportGrouping,
     ExportParser,
     ImportParser,
     NumberLike,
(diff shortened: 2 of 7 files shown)


darwin/cli.py
```diff
@@ -145,7 +146,13 @@ def _run(args: Namespace, parser: ArgumentParser) -> None:
         elif args.action == "remove":
             f.remove_remote_dataset(args.dataset)
         elif args.action == "report":
-            f.dataset_report(args.dataset, args.granularity or "day", args.pretty)
+            f.dataset_report(
+                args.dataset,
+                args.start or datetime(2024, 5, 11, tzinfo=timezone.utc),
```
umbertoDifa, line 151: I think this default deserves a comment for the future

darwin/client.py
```diff
@@ -54,6 +62,11 @@
 INITIAL_WAIT = int(os.getenv("DARWIN_RETRY_INITIAL_WAIT", "60"))
 MAX_WAIT = int(os.getenv("DARWIN_RETRY_MAX_WAIT", "300"))
 MAX_RETRIES = int(os.getenv("DARWIN_RETRY_MAX_ATTEMPTS", "10"))
+REPORT_JOB_MAX_RETRIES = int(os.getenv("DARWIN_RETRY_REPORT_JOB_MAX_ATTEMPTS", "10"))
```
umbertoDifa, line 65: I'm not sure this should be an additional env var, it's quite granular.
I'd say we should have a single env `MAX RETRIES` and that should apply for all the flows involving retries.
I doubt we're going to want to set different retries attempts per action

darwin/cli.py
AndriiKlymchuk, line 151: lets make start and stop required

darwin/client.py
```diff
@@ -628,11 +650,64 @@ def get_report(
 
         the_team_slug: str = the_team.slug
 
-        return self._get_raw(
-            f"/reports/{the_team_slug}/annotation?group_by=dataset,user&dataset_ids={dataset_id}&granularity={granularity}&format=csv&include=dataset.name,user.first_name,user.last_name,user.email",
-            the_team_slug,
+        report_job = cast(
+            Dict[str, UnknownType],
```
umbertoDifa, line 654: What do you think about typing the response structure? I'm not sure due to the fact we don't reuse much this `report_job`.
```
class ReportJob(TypedDict):
    job_id: str
    status: str
    created_at: str
    ...
    ```

File: darwin/cli_functions.py [hunk 4]
```diff
@@ -339,16 +349,21 @@ def dataset_report(dataset_slug: str, granularity: str, pretty: bool) -> None:
...
-        table.add_column("Images Annotated", justify="right")
-        table.add_column("Images Approved", justify="right")
-        table.add_column("Images Rejected", justify="right")
+        table.add_column("Full Name", justify="right")
+        table.add_column("Active Time", justify="right")
+        table.add_column("Total Annotations", justify="right")
+        table.add_column("Review Pass Rate", justify="right")
+        table.add_column("Total Items Annotated", justify="right")
+        table.add_column("Time Per Annotation", justify="right")
+        table.add_column("Time Per Item", justify="right")
```
AndriiKlymchuk, line 366: Lets make sure we have a doc describing the report

darwin/dataset/remote_dataset.py
```diff
@@ -719,14 +719,23 @@ def export(
         """
 
     @abstractmethod
-    def get_report(self, granularity: str = "day") -> str:
+    def get_report(
```
umbertoDifa, line 722: Seeing that we have a `remote_dataset` and `remote_dataset_v2` I'm tempted to say this version is no longer in use and we created the split when we converted datasets from v1 to v2.
This isn't the place to remove such abstraction, https://www.notion.so/v7labs/darwin-py-onboarding-19d15c85f35480feb2e1c46d8c307dca?pvs=4#19d15c85f354801a8229e566c0ecf2c6 I'll schedule a ticket and we'll priortize later

darwin/client.py
AndriiKlymchuk, line 65: Lets just keep it as a constant

darwin/dataset/remote_dataset_v2.py
```diff
@@ -504,22 +505,34 @@ def export(
             team_slug=self.team,
         )
 
-    def get_report(self, granularity: str = "day") -> str:
+    def get_report(
+        self,
+        start: datetime = datetime(2024, 5, 11, tzinfo=timezone.utc),
```
umbertoDifa, line 510: We have a default in the CLI command and we've a default here as well (but only for `start`)?
I guess this raises the question about where it is best to put a default considering dpy is used both as CLI and as SDK

darwin/client.py
```diff
@@ -628,11 +650,64 @@ def get_report(
...
+            self._post(
+                f"/v3/teams/{the_team_slug}/reports/annotator/jobs",
+                {
+                    "start": start.isoformat(timespec="seconds"),
+                    "stop": stop.isoformat(timespec="seconds"),
+                    "dataset_ids": [dataset_id],
+                    "group_by": (
+                        ["annotators", "stages"] if group_by_stage else ["annotators"]
```
AndriiKlymchuk, line 662: Lets clarify it it make sense to not group by annotators

darwin/options.py
```diff
@@ -205,10 +206,20 @@ def __init__(self) -> None:
...
+            type=lambda dt: datetime.strptime(dt, "%Y-%m-%dT%H:%M:%S%z"),
+            help="Report start DateTime in RFC3339 format (e.g. 2020-01-20T14:00:00Z)",
+        )
+        parser_report.add_argument(
+            "--stop",
+            type=lambda dt: datetime.strptime(dt, "%Y-%m-%dT%H:%M:%S%z"),
+            help="Report end DateTime in RFC3339 format (e.g. 2020-01-20T15:00:00Z)",
+        )
+        parser_report.add_argument(
+            "--group-by-stage",
```
umbertoDifa, line 219: Currently, we can group by:
- datasets
- annotators
- annotators and stages
I'm not sure a `--group-by-stage` is flexible enough, especially if we iterate on this in the future and we add more groupings (are we going to add a `--group-by-X`?).

Another option would be `--group-by [datasets, annotators, stages]`, here the challenge is that I'm not sure you can use any combination of those? eg. can we group-by `datasets stages`? or only `datasets` alone?
@balysv WDYT?

tests/darwin/client_test.py
```diff
@@ -680,3 +680,192 @@ def test_retry_respects_rate_limit_headers(self, mock_sleep, client):
...
+        with pytest.raises(ValueError, match="Building an annotator report failed"):
+            darwin_client.get_report(
+                dataset_id, start_time, end_time, False, team_slug_darwin_json_v2
+            )
+
+        assert start_job_endpoint_mock.call_count == 1
+        assert get_job_status_endpoint_mock.call_count == 1
+
+    @responses.activate(registry=OrderedRegistry)
+    def test_raises_if_started_job_cannot_be_quieried(
```
umbertoDifa, line 737: typo: queried

umbertoDifa (commented):
Looks good! I've added a few discussion point we should discuss about

darwin/client.py
```diff
@@ -628,11 +650,64 @@ def get_report(
...
+            the_team_slug, report_job["id"]
         )
 
+        return self._get_raw_from_full_url(finished_report_job["url"], the_team_slug)
+
+    @retry(
+        wait=wait_exponential(),
+        stop=stop_after_attempt(REPORT_JOB_MAX_RETRIES),
+        retry=retry_if_exception_type(JobPendingException),
+        before_sleep=log_retry_error,
```
AndriiKlymchuk, line 691: Lets remove it

darwin/client.py
```diff
@@ -628,11 +650,64 @@ def get_report(
...
+        )
+
+        finished_report_job = self.poll_pending_report_job(
+            the_team_slug, report_job["id"]
         )
 
+        return self._get_raw_from_full_url(finished_report_job["url"], the_team_slug)
+
+    @retry(
+        wait=wait_exponential(),
```
AndriiKlymchuk, line 688: lest cap at at 30 seconds

darwin/client.py
```diff
@@ -628,11 +650,64 @@ def get_report(
...
+        stop=stop_after_attempt(REPORT_JOB_MAX_RETRIES),
```
AndriiKlymchuk, line 689: Lets set max 2 hours

darwin/dataset/remote_dataset.py
```diff
@@ -719,14 +719,23 @@ def export(
...
+        self,
+        start: datetime = datetime(2024, 5, 11, tzinfo=timezone.utc),
```
AndriiKlymchuk, line 724: Lets make it required

tests/darwin/client_test.py
AndriiKlymchuk, line 1: Lets also add a cli test

darwin/dataset/remote_dataset_v2.py
vvihorev: Here `stop` also gets a default value (lines 531-532), I did not put it as a default argument value because `datetime.now()` would be taken at parse time instead of at call time if it was a default in kwargs.

We can remove defaults altogether, and make `start` required like in the API.

darwin/client.py
umbertoDifa: I think a large report might take more than 30 seconds, or if there's a queue of job reports

vvihorev: Here we meant, capping delay between retries to 30 seconds, and stopping retries completely after 2 hours

tests/darwin/cli_functions_test.py
```diff
@@ -448,3 +451,35 @@ def test_extract_video(self, tmp_path):
                 repair=False,
                 storage_key_prefix="test/prefix",
             )
+
+
+class TestDatasetReport:
+    def test_dataset_report(self, remote_dataset: RemoteDataset):
```
vvihorev, line 457: This test slightly breaks the test border in the module. It starts from CLI options, instead of passing args to a cli_function directly.

I didn't want to add an e2e test for reports, but also wanted to ensure datetimes and list of group by values were parsed correctly.

Validation of params happens `Client.get_report` to validate for CLI and library users in one place. Validation is covered by unittests for the client.

AndriiKlymchuk (approved):

darwin/cli_functions.py
```diff
@@ -1348,6 +1349,88 @@ def post_comment(
...
+
+    report: str = client.get_annotators_report(
+        dataset_ids,
+        start,
+        stop,
+        group_by,
+    ).text
+
+    if not pretty:
+        # if no one worked in the report, we print nothing
```
AndriiKlymchuk, line 1395: What this comment is about?

AndriiKlymchuk, line 1395: Will we print only csv headers if there is no data?

vvihorev: Not even headers, just empty output, as returned by the backend. This comment and logic is taken from V1 reports

darwin/cli_functions.py
```diff
@@ -1348,6 +1349,88 @@ def post_comment(
...
+    table.add_column("User Id", justify="right")
+    table.add_column("User Type", justify="right")
+    table.add_column("Email", justify="right")
+    table.add_column("Full Name", justify="right")
+    table.add_column("Active Time", justify="right")
+    table.add_column("Total Annotations", justify="right")
+    table.add_column("Review Pass Rate", justify="right")
+    table.add_column("Total Items Annotated", justify="right")
+    table.add_column("Time Per Annotation", justify="right")
+    table.add_column("Time Per Item", justify="right")
```
AndriiKlymchuk, line 1426: To make it more DRY, there can be just a list of column names and the for loop to add columns. Up to you.

vvihorev (commented):
started a review by accident

umbertoDifa (approved):

PR #1061, opened 2025-04-23
[DAR-6479][External] Get primary_plane from medical metadata of the slot during nifti export

vvihorev:
# Problem
Non-axial DICOM files with extracted views and at least one mask annotation break during export in nifti format. We need to know which is the primary_plane for non-axial files in Darwin Py to export them properly. The primary_plane (not axial) determined by the volumetric extractor is not stored in annotation files.

DICOM files uploaded without exporting views have one slot. If the file is non-axial this slot would have the primary plane in the plane map in metadata e.g. `plane_map: {"0": "CORONAL"}`. 

However, files uploaded with views exported have three slots, and always have `plane_map: {"0.1": "AXIAL", "0.2": "CORONAL", "0.3": "SAGITTAL"}`. Taking the plane name from the first slot is incorrect, and we need to know which plane is the primary one explicitly.

[PR to support `primary_plane` storage in annotations on the backend](https://github.com/v7labs/darwin-backend/pull/5687)

# Solution
We get `primary_plane` from medical metadata of DICOM annotation files. We use this `primary_plane` set by the volumetric extractor instead of the plane of the first slot to populate volumes from raster layer during nifti export.

# Changelog
Use `primary_plane` from medical metadata of DICOM annotation files to populate volumes from raster layer during nifti export


darwin/exporter/formats/nifti.py
```diff
@@ -83,7 +83,9 @@ def export(
             medical_metadata = video_annotation.slots[0].metadata
             legacy = not medical_metadata.get("handler") == "MONAI"  # type: ignore
             plane_map = medical_metadata.get("plane_map", {slot_name: "AXIAL"})
-            primary_plane = plane_map.get(slot_name, "AXIAL")
+            primary_plane = medical_metadata.get(
+                "primary_plane", plane_map.get(slot_name, "AXIAL")
```
AndriiKlymchuk, line 87: Plane map show which plane represents specific slot, right?
Is it correct at all to use it as fallback for primary plane?
Maybe just default to the most commonly used here?

AndriiKlymchuk: Oh, the idea is that annotations can be created only in the primary plane ...
When is it not the case?
Nevertheless, let's leave it as is.

AndriiKlymchuk (approved):

PR #1063, opened 2025-04-29
[DAR-6239][External] Fix nifti export for multi slotted items

vvihorev:
# Problem
When a multi-slot item containing multiple DICOM slots is exported as a nifti volume, only the first slot is exported.

# Solution
Iterate over all slots when exporting a file. If there are more than one slot, create a separate `AnnotationFile` for each slot, thus generating a separate output volume for each mask in each slot.

# Changelog
Fix nifti export broken for multi-slotted items


e2e_tests/data/import/multi_slotted_annotations_with_dicom_slots/multi_slotted_dicom_item.json
```diff
@@ -0,0 +1,196 @@
...
+          ],
+          "pixdim": "(1.4583334, 1.4583334, 1.1999999)",
+          "plane_map": {
+            "0": "AXIAL"
+          }
+        }
+      }
+    ]
+  },
+  "annotations": [
```
vvihorev, line 137: A bit confused about importing annotation for multiple slots:
- in e2e tests there is a notion of `base_slot`, which is the one containing all annotations, other slots are not annotated
- during import, we take the first slot as the default slot, and it looks like we associate imported annotations to it

darwin/utils/utils.py
```diff
@@ -597,11 +601,29 @@ def _parse_darwin_v2(path: Path, data: Dict[str, Any]) -> dt.AnnotationFile:
...
     else:
-        slot = slots[0]
+        filename = item["name"]
+
+        if slot_index is None or len(slots) == 1:
+            slot = slots[0]
+        else:
+            slot = slots[slot_index]
+            slots = [slot]
+            filename = item["name"] + "-" + slot.name
```
umbertoDifa, line 611: Is this naming convention aligned with the `push` and `pull` command?
When we `pull` a release that contains multislot item each image/video is pulled. IIRC we create a folder named with the `item["name"]` and we create files using the `slot.name`.

When we `push` -item-merge-mode=slots https://docs.v7labs.com/docs/push 
https://github.com/v7labs/darwin-py/blob/e053c427f874ab0b7ab9e3f3f26def140c90545b/darwin/dataset/upload_manager.py#L277
(I might be wrong, just a hint to double-check the consistency of our naming)

darwin/utils/utils.py
```diff
@@ -597,11 +601,29 @@ def _parse_darwin_v2(path: Path, data: Dict[str, Any]) -> dt.AnnotationFile:
...
+            annotations = [
+                annotation
+                for annotation in annotations
+                if hasattr(annotation, "slot_names")
+                and annotation.slot_names
+                and annotation.slot_names[0] == slot.name
+            ]
```
umbertoDifa, line 618: I wonder if we have an utility "get_annotations_in_slot" or something similar?

e2e_tests/data/import/multi_slotted_annotations_with_dicom_slots/multi_slotted_dicom_item.json
umbertoDifa: `base_slot` was a concept introduced with `channels` https://docs.v7labs.com/docs/labelling-with-channels?utm_campaign=%5BDarwin%5D%20Product%20Newsletter&utm_source=hs_email&utm_medium=email&_hsenc=p2ANqtz--Y2gN_R_TlpqtwZK3My2u5k-KrQhhje-htdoxByW6xK2Fkjggl2-YM-pXua-aODGJyXopP#importing-channel-files the reason is: when we upload files as channels they visually 'overlap'. So annotations are always associated to the first slot for simplicity.

When the modality is slots instead, each slot has its own annotations

e2e_tests/helpers.py
```diff
@@ -229,6 +229,7 @@ def export_release(
         "Authorization": f"ApiKey {api_key}",
     }
     response = requests.post(create_export_url, json=payload, headers=headers)
+    response.raise_for_status()
```
vvihorev, line 232: I added a status check because 422 return code was not caught while testing. Turns out there is a failing e2e import test.

darwin/utils/utils.py
vvihorev: I aligned the naming convention by storing exported nifti volumes in a `{item_name}/{slot_name}/{filename}` directory structure

However we can only push only one directory at a time with `--item-merge-mode=slots`, and the directory is expected to be not nested: `{item_name}/{filename}`, slot names are set to indices of files in the item directory

darwin/utils/utils.py
vvihorev: I don't see one yet, looks like we don't often work with annotations made to different slots

darwin/cli_functions.py
```diff
@@ -1243,7 +1243,9 @@ def dataset_convert(
             output_dir = Path(output_dir)
         output_dir.mkdir(parents=True, exist_ok=True)
 
-        export_annotations(parser, [annotations_path], output_dir)
+        export_annotations(
+            parser, [annotations_path], output_dir, split_sequences=(format != "nifti")
+        )
```
vvihorev, line 1248: While testing I noticed that only `darwin convert` handled nifti export properly, `darwin dataset convert` logic was not updated (nifti files were exported as separate slices)

e2e_tests/helpers.py
vvihorev: created a ticket [DAR-6571: Failing e2e import test](https://linear.app/v7labs/issue/DAR-6571/failing-e2e-import-test)

darwin/cli_functions.py
vvihorev: Sure, ticketed: [DAR-6570: `dataset convert` and `convert` logic are out of sync for nifti exports](https://linear.app/v7labs/issue/DAR-6570/dataset-convert-and-convert-logic-are-out-of-sync-for-nifti-exports)

umbertoDifa (approved):
LGTM!

darwin/exporter/formats/nifti.py
```diff
@@ -78,95 +80,107 @@ def export(
     """
     video_annotations = list(annotation_files)
     for video_annotation in video_annotations:
```
AndriiKlymchuk, line 82: This names are confusing. It should rather be
```
for annotation_file in annotation_files:
````

File: darwin/exporter/formats/nifti.py [hunk 7]
```diff
@@ -544,10 +553,23 @@ def unnest_dict_to_list(d: Dict) -> List:
...
+                / f"{filename_stem}_{volume.class_name}_m.nii.gz"
+            )
         else:
-            output_path = Path(output_dir) / f"{image_id}_{volume.class_name}.nii.gz"
+            output_path = (
+                Path(output_dir)
+                / item_name
+                / slot_name
+                / f"{filename_stem}_{volume.class_name}.nii.gz"
+            )
```
AndriiKlymchuk, line 572: Does this change affect the structure of an export? If so, it must be backward compatible.

vvihorev: It does affect the export structure for all nifti exports, however the new structure is now the same as how source files of multislot items are laid out by `dataset pull`
We decided to make it backwards incompatible and bump to v3.0.0

AndriiKlymchuk (approved):

PR #1080, opened 2025-06-20
Import coco annotation tags

BenediktHeinK:
Implemented functionality to import tags assigned to annotations when using darwin.importer with coco.json files.

Requires the annotation tags to reside in annotation["extras"]["attributes"] (Same format as when exporting darwin annotations with tags to coco format)

# Problem
darwin importer ignored annotation attributes (or tags) when importing based on a coco.json file

# Solution
When reading an annotation, check for availability of field annotation["extras"]["attributes"], add them like so subs.append(dt.make_attributes(annotation["extra"]["attributes"]))

# Changelog
Support annotation attributes import for coco format

### DIFF
--- a/darwin/importer/formats/coco.py
+++ b/darwin/importer/formats/coco.py
@@ -132,6 +132,7 @@ def parse_annotation(
         )
         return []
 
+
     if len(segmentation) == 0 and len(annotation["bbox"]) == 4:
         x, y, w, h = map(int, annotation["bbox"])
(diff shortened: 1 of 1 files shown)


vvihorev (changes requested):
Hello, thank you for the PR!

Could you please cover this change in the [coco importer tests](https://github.com/v7labs/darwin-py/blob/master/tests/darwin/importer/formats/import_coco_test.py)?

PR #1082, opened 2025-07-03
[External] Add multi-segment nifti export support

Jo-Byr:
# Problem
The current nifti exporter creates one nifti file per slot and label, it would be useful to generate one multi-label nifti per slot.

# Solution
This solution merges all generated nifti file into one, while maintaining the previous ones, with the suffix "Segments".

Modified nifti exporter to do the above.
Added unit test comparing produced multi-label nifti's array to a manually generated nifti

# Changelog
Add multi-segment nifti export support


vvihorev (approved):

vvihorev:
Closing and moving to a copy of this PR opened from a branch in origin: https://github.com/v7labs/darwin-py/pull/1086

PR #1151, opened 2026-04-20
[DAR-7810] feat: SDK support for nested properties (read + create + import)

umbertoDifa:
Closes [DAR-7810](https://linear.app/v7labs/issue/DAR-7810/add-importexport-functionality).

## Context

The nested-properties feature has now landed on both backend and frontend (Linear project: [Nested Properties](https://linear.app/v7labs/project/nested-properties-5609880da576/issues)):

**Backend (merged on master):**
- [DAR-7801](https://linear.app/v7labs/issue/DAR-7801) — create/read API + ``dataset_ids`` inheritance ([#6115](https://github.com/v7labs/darwin-backend/pull/6115))
- [DAR-7811](https://linear.app/v7labs/issue/DAR-7811) — ``NESTED_PROPERTIES`` feature flag ([#6119](https://github.com/v7labs/darwin-backend/pull/6119))
- [DAR-7802](https://linear.app/v7labs/issue/DAR-7802) — cascade delete on nested property/value ([#6116](https://github.com/v7labs/darwin-backend/pull/6116))
- [DAR-7803](https://linear.app/v7labs/issue/DAR-7803) — scoped sibling reorder ([#6117](https://github.com/v7labs/darwin-backend/pull/6117))
- [DAR-7804](https://linear.app/v7labs/issue/DAR-7804) — conditional required validation ([#6118](https://github.com/v7labs/darwin-backend/pull/6118))
- [DAR-7825](https://linear.app/v7labs/issue/DAR-7825) — SSP write guard inherits allow-list from root ([#6134](https://github.com/v7labs/darwin-backend/pull/6134))

**Frontend (merged on master):**
- [DAR-7805](https://linear.app/v7labs/issue/DAR-7805) — author nested in Class Editor ([#9951](https://github.com/v7labs/darwin-frontend/pull/9951))
- [DAR-7806](https://linear.app/v7labs/issue/DAR-7806) — scope-aware drag reorder ([#9952](https://github.com/v7labs/darwin-frontend/pull/9952))
- [DAR-7807](https://linear.app/v7labs/issue/DAR-7807) — dynamic visibility in Workview
- [DAR-7808](https://linear.app/v7labs/issue/DAR-7808) — item-level nested end-to-end ([#9964](https://github.com/v7labs/darwin-frontend/pull/9964))
- [DAR-7823](https://linear.app/v7labs/issue/DAR-7823) / [DAR-7824](https://linear.app/v7labs/issue/DAR-7824) — SSP write-guard fix + shared ancestry helper

## What this PR adds (darwin-py)

Closes the loop: ``darwin dataset import`` and the SDK programmatic API now understand nested properties.

### User-facing capability

```bash
# .v7/metadata.json declares nesting by NAME (consistent with the rest of darwin-py):
#   { "parent_name": "Defect Type", "trigger_condition": { "type": "value_match", "values": ["Contamination"] } }
$ darwin dataset import my-dataset darwin path/to/annotations
# Importer creates parents first (topological sort), then resolves each child's
# parent_name + trigger value names against the freshly-created parent on the
# destination team to produce the API-shape parent_property_id + property_value_ids.
```

```python
# Programmatic creation works against the merged BE
client.create_property(FullProperty(
    name="Bone", type="single_select", required=False,
    annotation_class_id=42,
    parent_property_id=parent.id,
    trigger_condition=TriggerCondition(
        type="value_match",
        property_value_ids=[v.id for v in parent.property_values if v.value == "Fracture"],
    ),
    property_values=[PropertyValue(value="Femur")],
    granularity=PropertyGranularity.annotation,
))
```

### Scope

- ``TriggerCondition`` model (name-based ``values`` for SDK + UUID-based ``property_value_ids`` for the API).
- ``FullProperty`` gains ``parent_name`` (SDK-native) and ``parent_property_id`` (API).
- ``PropertyKey = Tuple[str, Optional[int]]`` matching ``TeamPropertyLookups`` keying — so ``lookups.annotation_properties[property_key(prop)]`` is drop-in.
- Importer:
  - ``_topologically_sort_properties_to_create`` — parents created before children.
  - ``_enrich_properties_with_metadata_values`` — every metadata value sent on initial create so trigger references resolve.
  - ``_resolve_parent_for_create`` — names → UUIDs just-in-time via ``team_property_lookups.refresh()``.
  - ``_assign_item_properties_to_dataset`` skips nested item children — they inherit ``dataset_ids`` from the root ancestor (per [DAR-7801](https://linear.app/v7labs/issue/DAR-7801)).
- E2E fixture covering all 3 granularities (annotation + section + item) with children listed before parents.
- E2E test: ``test_full_cycle_nested_properties``.
- ~50 unit tests covering: trigger validators, topological sort, parent resolution, metadata enrichment, ``property_key``, contract conformance.

### Alignment audit

Diffed darwin-py against current ``v7labs/darwin-backend@master`` schema — no wire contract drift. The TDD ``parent_property_id`` + ``trigger_condition.{type, property_value_ids}`` shape on ``CreateRequest`` is identical between this PR and the merged BE.

## Test plan

- [x] ``poetry run python -m pytest tests/ darwin/future/tests/`` — 1171 passing, 0 regressions
- [x] Black 26.x + ruff clean on touched files
- [ ] ``test_full_cycle_nested_properties`` (live BE) — passes once the staging team has ``NESTED_PROPERTIES`` enabled and the merged BE is deployed.

---

> [!NOTE]
> **Medium Risk**
> Changes property create/import behavior by adding parent/trigger resolution and ordering logic; mistakes could create incorrect property hierarchies or fail imports, though guarded by extensive unit tests and validation.
> 
> **Overview**
> Adds SDK support for **nested properties** by introducing `TriggerCondition` plus `parent_name`/`parent_property_id` fields, validating nested-metadata consistency, and ensuring API payloads only include the UUID-based trigger shape.
> 
> Updates `darwin dataset import` to create nested properties reliably: metadata values are fully included on create, properties are topologically sorted so parents are created before children, and child `parent_name` + trigger value names are resolved to server UUIDs before calling `create_property`. Lookup handling is refactored via `TeamPropertyLookups.register`, and item-level dataset assignment now skips nested children (which inherit `dataset_ids`).
> 
> Adds comprehensive coverage: a new nested-properties E2E fixture and round-trip assertion (marked xfail pending BE export support) plus a large unit test suite covering validation, sorting, resolution, payload shaping, and lookup keying (`property_key`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e9ec78f44b7d9d01dd35d8fb6e9b3de983fb9d0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


umbertoDifa:
Thanks Bugbot — both issues addressed in 9fdc41ba:

- **Medium (value-ID remap with unused parent values)**: Added `_enrich_properties_with_metadata_values`, which runs before the topological-sort + create loop and supplements each to-create property with every value declared in `.v7/metadata.json`. This guarantees `old_to_new_value_id` has an entry for every value any child trigger could reference, regardless of whether the annotations happen to use it.
- **Low (cycle guard in `is_visible`)**: Added an in-flight visitor set; a cyclic `parent_property_id` graph now resolves to "hidden" instead of recursing until `RecursionError`. The backend still rejects cycles at create time, but the public `get_visible_properties` utility is now safe on arbitrary input.

Three new unit tests cover both fixes.

umbertoDifa:
Addressed both review notes in 0a4aeb26:

**1. `model_validator(mode="after")` was inconsistent with the codebase.** The established pattern across `item.py`, `workflow.py`, `advanced_filters.py`, `sorting.py`, and `client.py` is `model_validator(mode="before")` for cross-field validation. `TriggerCondition` now uses that pattern.

**2. metadata.json now uses names for nesting, consistent with the rest of darwin-py.** Properties and values are identified by name everywhere else in the SDK (`_import_properties` resolves annotation property names against `TeamPropertyLookups`, `_update_payload_with_properties` keys on name, etc.) — so nesting follows suit:

- `parent_name: Optional[str]` on each property in `.v7/metadata.json`
- `trigger_condition.values: List[str]` (parent value **names**, not UUIDs)

`FullProperty.parent_property_id` and `TriggerCondition.property_value_ids` remain because the REST API contract uses UUIDs, but the importer now resolves names → UUIDs just-in-time after each parent is created on the destination team. The source→destination UUID remap helpers (`_remap_property_for_create`, `_record_created_property_id_mapping`) are gone; `_resolve_parent_for_create` replaces them with a name-based lookup against `TeamPropertyLookups`.

46 nested-property unit/integration tests passing, 1163 repo-wide.

umbertoDifa:
Bugbot comments from `ba36ba7f` and `9fdc41ba` are now stale — they flagged the source→destination UUID remap helpers (`_record_created_property_id_mapping`, `_remap_property_for_create`, `old_to_new_property_id`, `old_to_new_value_id`) which were **entirely deleted** in `0a4aeb26`.

The name-based refactor replaced the remap approach with `_resolve_parent_for_create`, which resolves the parent by name against a freshly-refreshed `TeamPropertyLookups` right before each child is sent to the API. This sidesteps both flagged issues by construction:

- "Trigger value IDs not mapped when absent from annotations" → no longer applies; trigger values are looked up by **name** against the just-created parent's server-side `property_values`, not through a precomputed ID map
- "Unmapped parent ID when parent exists on destination" → no longer applies; parent resolution goes through `team_property_lookups.annotation_properties[(parent_name, class_id)]` / `item_properties[parent_name]`, so parents that already exist on the team are found exactly like parents just created in the batch

The third comment (cycle guard in `is_visible`) is still relevant and the guard remains in place after the refactor.

darwin/future/data_objects/properties.py
```diff
@@ -172,3 +280,191 @@ class SelectedProperty(DefaultDarwin):
...
+        visibility_cache[key] = result
+        return result
+
+    visible: List[SelectedProperty] = []
+    for selected in annotation_properties:
+        definition = definitions_by_name.get(selected.name)
+        # No definition available -> be conservative and keep the value.
+        if definition is None or is_visible(definition):
+            visible.append(selected)
+    return visible
```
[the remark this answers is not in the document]
umbertoDifa: Disagree — this is intentional public SDK surface, not dead code.

Per the backend TDD section 12 (`Export Visibility Filter`):

> It is also exposed as a public utility for darwin-py users who fetch raw annotations via the API (which returns all stored values including invisible ones) and want to filter client-side.

The darwin-py exporter itself intentionally does not include property values in its output (see the TDD: `The existing export pipeline currently does not include properties in rebuilt JSON — it only serializes geometry`), so there is no internal call site to wire up right now. The function exists for two audiences:

1. External consumers who call `GET /items/:id/annotations` directly and need to mirror the frontend `VisibilityEngine` / backend `GetRequiredPropertiesPresence` filtering.
2. Future darwin-py code when/if the exporter starts serializing properties.

Resolving the thread.

darwin/future/data_objects/properties.py
```diff
@@ -172,3 +280,16 @@ class SelectedProperty(DefaultDarwin):
...
+# ``(annotation_class_id, name)`` — item-level properties use ``None`` for the
+# class id. This matches ``TeamPropertyLookups.annotation_properties`` (keyed
+# by ``(name, class_id)``) and ``item_properties`` (keyed by ``name``). Code
+# that needs to locate or deduplicate property definitions should use this
+# helper rather than inventing its own keying.
+PropertyKey = Tuple[Optional[int], str]
+
+
+def property_key(prop: "FullProperty") -> PropertyKey:
+    return (prop.annotation_class_id, prop.name)
```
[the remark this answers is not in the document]
umbertoDifa: Good catch — fixed in 37d1c165. The tuple order was reversed from ``TeamPropertyLookups.annotation_properties`` which made the helper pay-for-play rather than drop-in. Flipped ``property_key`` to ``(name, annotation_class_id)``, updated the single caller + tests, and added ``test_matches_team_property_lookups_key_order`` so future reorders fail loudly.

darwin/future/data_objects/properties.py
```diff
@@ -19,6 +19,80 @@
...
+    values: Optional[List[str]] = None
+
+    @model_validator(mode="before")
+    @classmethod
+    def _validate_shape(cls, data: Any) -> Any:
+        # ``mode="before"`` runs for every construction path — ``model_validate``
+        # with a raw dict, direct ``TriggerCondition(type=..., ...)`` keyword
+        # construction, and nested validation where Pydantic may hand us an
+        # already-built instance. Only the dict path needs cross-field checks;
+        # the others have already been validated once.
```
umbertoDifa: remove this

darwin/future/data_objects/properties.py
```diff
@@ -19,6 +19,80 @@
...
+                )
+            if property_value_ids:
+                raise ValueError(
+                    "trigger_condition.property_value_ids must be empty/None for 'any_value'"
+                )
+        elif trigger_type == "value_match":
+            if not values and not property_value_ids:
+                raise ValueError(
+                    "trigger_condition must set either 'values' (names) or "
+                    "'property_value_ids' (UUIDs) for 'value_match'"
```
umbertoDifa, line 49: I'm not sure why either should be set. During parsing of the annotation file only the names can be set, later on also property_values_ids (before the API calls are made).

darwin/future/data_objects/properties.py
```diff
@@ -85,11 +159,48 @@ class FullProperty(DefaultDarwin):
     dataset_ids: Optional[List[int]] = None
     options: Optional[List[PropertyValue]] = None
     granularity: PropertyGranularity = PropertyGranularity("section")
+    # Nesting metadata. ``parent_name`` is the darwin-py / ``.v7/metadata.json``
+    # representation (identify the parent by its property name, consistent
+    # with how the SDK identifies everything else). ``parent_property_id`` is
+    # populated from REST API responses and sent back on create — it is
+    # resolved from ``parent_name`` by the importer after the parent has been
+    # created on the server.
```
umbertoDifa: remove this comment

darwin/future/data_objects/properties.py
```diff
@@ -85,11 +159,48 @@ class FullProperty(DefaultDarwin):
...
     ) -> dict:
-        include_fields = {
+        if (
+            self.granularity != PropertyGranularity.item
+            and self.annotation_class_id is None
+        ):
+            raise ValueError("annotation_class_id must be set")
+        # A nested property must carry both an API-ready parent UUID and a
+        # trigger condition. ``parent_name`` is the SDK-local representation
+        # and must be resolved to ``parent_property_id`` before calling this.
```
umbertoDifa: remove comment

darwin/future/data_objects/properties.py
```diff
@@ -85,11 +159,48 @@ class FullProperty(DefaultDarwin):
...
+            raise ValueError(
+                "parent_property_id and trigger_condition must both be set or "
+                "both be None. If you have a parent_name, resolve it to a "
+                "parent_property_id first."
+            )
+        # ``parent_name`` carries no information for the BE — it's the
+        # SDK-local mirror of ``parent_property_id``. Letting it be set on
+        # its own would silently send a flat-property payload while the
+        # caller thinks they're creating a nested child. Catch that here
+        # rather than producing a degraded property server-side.
```
umbertoDifa: remove comment

darwin/future/data_objects/properties.py
```diff
@@ -99,12 +210,19 @@ def to_create_endpoint(
...
+        if self.parent_property_id is not None:
+            include_fields["parent_property_id"] = True
+
+        payload = self.model_dump(mode="json", include=include_fields)
+        if self.trigger_condition is not None:
+            # ``TriggerCondition.to_api_payload`` owns the wire shape so
+            # ``to_create_endpoint`` doesn't have to reach into a nested
+            # model's representation.
+            payload["trigger_condition"] = self.trigger_condition.to_api_payload()
+        return payload
```
umbertoDifa: the order of this ops is weird. Before we've already checked that parent_property_id and trigger_condition must be both present or both absent. But here we check each one independently and the trigger_condition even after the payload creation. I don't think this # ``TriggerCondition.to_api_payload`` owns the wire shape so
            # ``to_create_endpoint`` doesn't have to reach into a nested
            # model's representation. is adding good abstraction or readability.

darwin/future/data_objects/properties.py
```diff
@@ -113,6 +231,9 @@ def to_update_endpoint(self) -> Tuple[str, dict]:
         updated_base = self.to_create_endpoint()
         updated_base.pop("annotation_class_id", None)  # Can't update this field
         del updated_base["granularity"]  # Can't update this field
+        # parent_property_id and trigger_condition are immutable after creation
+        updated_base.pop("parent_property_id", None)
+        updated_base.pop("trigger_condition", None)
```
umbertoDifa: use del

darwin/future/data_objects/properties.py
```diff
@@ -172,3 +293,17 @@ class SelectedProperty(DefaultDarwin):
...
     value: Optional[str] = None
+
+
+# Properties are identified within a team by the tuple
+# ``(name, annotation_class_id)`` — item-level properties use ``None`` for the
+# class id. The order mirrors ``TeamPropertyLookups.annotation_properties``
+# (keyed ``(name, annotation_class_id)``) so a caller can do
+# ``lookups.annotation_properties[property_key(prop)]`` directly. Code that
+# needs to locate or deduplicate property definitions should use this helper
+# rather than inventing its own keying.
```
umbertoDifa: remove this comment

darwin/future/data_objects/properties.py
umbertoDifa: it looks like we're mixing parse time schema with API schema

umbertoDifa:
@cursoragent / Bugbot — addressed the line 93 finding in 76ecf494.

Going further than the suggested ``is not None``: the silent-drop path is now an explicit ``ValueError``. For ``type == "value_match"`` we always require a non-empty ``property_value_ids`` (with the error pointing the caller at ``_resolve_parent_for_create``), and only emit the ``property_value_ids`` key in that branch. ``any_value`` continues to omit the field entirely so the BE's non-nullable schema doesn't 422 on ``null``.

New unit test ``test_to_api_payload_value_match_raises_without_property_value_ids`` covers the failure mode.

Same commit also picks up self-review TODOs on this file (drop the ``TriggerConditionType`` alias, group ``parent_property_id`` + ``trigger_condition`` payload assembly behind the existing both-or-neither invariant, and remove a few comment blocks that just restated nearby code/error messages).

darwin/importer/importer.py
```diff
@@ -855,6 +1073,22 @@ def _import_properties(
...
+        # names a parent value no annotation happens to use can still be
+        # resolved to a UUID after the parent is created.
+        _enrich_properties_with_metadata_values(
+            properties_to_create,
+            metadata_cls_prop_lookup,
+            metadata_item_prop_lookup,
+            annotation_class_ids_map,
+        )
+        # Create parents before children so that a child's ``parent_name``
+        # is resolvable against a freshly-created parent on the server.
```
umbertoDifa: remove comment

darwin/importer/importer.py
```diff
@@ -889,6 +1132,11 @@ def _import_properties(
             )
             updated_properties.append(prop)
 
+    # Re-fetch the team's properties so downstream lookups reflect the
+    # latest server state — including any mutations from a concurrent
+    # client editing the same team. The per-create ``register()`` covers
+    # in-batch nested resolution; this refresh is the freshness guard
+    # before the metadata-extras loop reads the lookups again.
```
umbertoDifa: remove this comment

darwin/importer/importer.py
```diff
@@ -918,6 +1166,7 @@ def _import_properties(
             )
             updated_properties.append(prop)
 
+        # Same freshness contract as the post-create-batch refresh above.
```
umbertoDifa: remove this comment

darwin/future/data_objects/properties.py
umbertoDifa: Both fields exist as optional because `TriggerCondition` has three legal construction states from three different sources:

| Source | `values` | `property_value_ids` |
|---|---|---|
| `metadata.json` parse | ✓ | — |
| API response (`client.list_properties`, etc.) | — | ✓ |
| Post-resolution (in `_resolve_parent_for_create`) | ✓ | ✓ |

The validator only catches the genuinely-bad fourth state (neither set), which happens only on hand-constructed objects.

Happy to tighten the error message to mention both paths — something like `"trigger_condition.value_match must carry 'values' (parse path) or 'property_value_ids' (API path)"` — let me know if you want that.

darwin/future/data_objects/properties.py
umbertoDifa: `parent_property_id` and `trigger_condition` are conditionally added in `to_create_endpoint` (only when both are non-`None`); flat properties never have these keys in the payload. `del` would raise `KeyError` for any flat-property update.

This file already follows the convention: `del` for fields always present (`granularity`, line 220) and `pop(..., None)` for conditional ones (`annotation_class_id`, line 219).

If you specifically want `del` semantics, the alternative is `if "parent_property_id" in updated_base: del updated_base["parent_property_id"]` — extra line of code for the same effect. Let me know which you'd prefer.

darwin/future/data_objects/properties.py
umbertoDifa: You're right that the current shape mixes the two schemas. The trade-off I made:

**One type with optional fields (current)**
- Pros: one import, one mental model, easy to round-trip metadata→API→metadata in the importer.
- Cons: invariants are enforced by a validator instead of by the type system; readers have to know which field is set when.

**Two types — `MetadataTriggerCondition` + `ApiTriggerCondition`**
- Pros: each type has only the fields legal in its source schema; the resolution step becomes an explicit conversion function.
- Cons: callers (importer, factories, tests) have to pick the right type; `_resolve_parent_for_create` becomes a typed adapter; more boilerplate.

I went with one type because the importer is the only consumer that ever holds both at once, and only briefly (between resolution and the `client.create_property` call). For everything else it's effectively single-source.

If you'd rather see the typed split, happy to do it as a follow-up — it's a contained refactor. Want me to file a ticket / open a draft PR after this lands?

darwin/future/data_objects/properties.py
umbertoDifa: Removed in 87ff08b0.

darwin/future/data_objects/properties.py
umbertoDifa: Removed in 87ff08b0.

darwin/importer/importer.py
umbertoDifa: Removed in 87ff08b0.

darwin/importer/importer.py
umbertoDifa: Removed in 87ff08b0.

darwin/importer/importer.py
umbertoDifa: Removed in 87ff08b0.

darwin/future/data_objects/properties.py
umbertoDifa: Comment block already removed in an earlier refactor; this thread is stale.

darwin/future/data_objects/properties.py
umbertoDifa: Comment block already removed in an earlier refactor; this thread is stale.

darwin/future/data_objects/properties.py
umbertoDifa: Comment block already removed in an earlier refactor; this thread is stale.

darwin/future/data_objects/properties.py
umbertoDifa: You're right that the original ordering was awkward. Current shape (after the refactor):

1. Both invariants are checked **upfront** — `(parent_property_id is None) != (trigger_condition is None)` and `parent_name without parent_property_id`. If either is violated, raise before building anything.
2. Build the base payload via `model_dump(include=...)`.
3. **Conditionally** extend the payload with `parent_property_id` and `trigger_condition` (only when both are non-`None`, which the upfront check has already established as the only valid nesting state).

So the bottom check `if self.parent_property_id is not None and self.trigger_condition is not None` is technically redundant given the upfront validator, but it serves as a static guard for type-checkers (the optional types still allow `None`). I can drop it and rely solely on the upfront check if you prefer — either way works.

umbertoDifa:
Superseded by stack #1159 → #1160 → #1161:

- #1159 — (1/3) Nest properties against existing team parent (base: `master`)
- #1160 — (2/3) Nest parent + child created in the same batch (base: #1159)
- #1161 — (3/3) E2E round-trip across all three nesting granularities (base: #1160; xfail on DAR-7832)

Same code as `2e9ec78f`, just split into three reviewable chunks. Leaving this PR in draft for archaeology — please review the stack instead.

PR #1159, opened 2026-05-06
[DAR-7810] (1/3) Nest properties against existing team parent

umbertoDifa:
## Context

Stack 1/3 slicing #1151 (rollup) into reviewable chunks. Each PR in the
stack is independently green; the rollup itself was previously reviewed
end-to-end. The goal of the slice is to make the same code reviewable in
smaller pieces — no behavioural drift from the rollup.

- PR 1 (this one): nest a child whose parent already exists on the team.
- PR 2: nest parent + child created in the same batch (topo sort + name
  resolution against in-flight lookups, item-level granularity).
- PR 3: e2e round-trip across all three granularities (xfail, blocked on
  DAR-7832 BE export).

## Capability

The importer can create a nested child property when its parent already
exists on the team. Class-level only (annotation + section). No
batch-creation of parent + child.

## Included in this PR

- ``TriggerCondition`` model with name-based ``values`` and UUID-based
  ``property_value_ids``, plus ``to_api_payload`` that strips SDK-local
  fields and refuses to serialise unresolved triggers.
- ``parent_name`` / ``parent_property_id`` / ``trigger_condition`` on
  ``FullProperty``, with create/update endpoint guards.
- ``parent_name`` / ``trigger_condition`` on the ``Property`` dataclass,
  parsed from ``.v7/metadata.json`` via ``_property_from_metadata_dict``
  + ``_validate_metadata_nesting``.
- ``_property_value_from_metadata`` helper (replaces five call sites).
- ``_resolve_parent_for_create`` — class-level branch only. Resolves
  parent names and ``trigger_condition.values`` to UUIDs against
  ``team_property_lookups.annotation_properties`` before
  ``client.create_property``.
- ``PropertyKey`` type alias and ``property_key()`` helper.
- Class-level path in ``_import_properties`` propagates ``parent_name``
  and ``trigger_condition`` from ``Property`` to ``FullProperty``.

## Explicitly NOT in this PR (lands in PR 2 / 3)

- ``_topologically_sort_properties_to_create``
- ``_enrich_properties_with_metadata_values``
- ``TeamPropertyLookups.register()`` (class stays as-is for now)
- Item-level branch of ``_resolve_parent_for_create``
- ``_create_update_item_properties`` ``parent_name`` / ``trigger_condition``
  propagation
- ``_assign_item_properties_to_dataset`` skip for nested item-level
- E2E test + fixture (PR 3, xfail on DAR-7832)

## Tests

40 new unit tests in
[...]

PR #1160, opened 2026-05-06
[DAR-7810] (2/3) Nest parent + child created in the same batch

umbertoDifa:
## Context

Stack 2/3 slicing #1151 (rollup). Stacked on #1159.

- PR 1 (#1159): nest a child whose parent already exists on the team.
- PR 2 (this one): nest parent + child created in the same batch
  (topo sort, name resolution against in-flight lookups, item-level).
- PR 3: e2e round-trip across all three granularities (xfail, blocked
  on DAR-7832 BE export).

## Capability

The importer can create a parent and its nested children in the same
batch — even when ``metadata.json`` lists children before parents — at
both class-level (annotation + section) and item-level granularities.

## Included in this PR

- ``_topologically_sort_properties_to_create`` (with cycle detection).
- ``_enrich_properties_with_metadata_values`` — appends every
  metadata-declared value to each property about to be created so the
  child can resolve ``trigger_condition.values`` against a parent
  that's also being created in the same batch.
- ``TeamPropertyLookups.register`` — in-place mutator used after every
  ``client.create_property`` to feed the newly-created property back
  into ``team_property_lookups`` for sibling resolution. No more
  per-child network refresh on the hot path. ``refresh`` is now
  expressed in terms of ``register``.
- Item-level branch in ``_resolve_parent_for_create``.
- Item-level path in ``_create_update_item_properties`` propagates
  ``parent_name`` / ``trigger_condition`` (with
  ``_validate_metadata_nesting`` enforcement).
- ``_assign_item_properties_to_dataset`` skips nested item-level
  properties (BE inheritance contract: nested children's
  ``dataset_ids`` is always ``[]``).
- Annotation-import update path now uses ``register(prop)``.

## Tests

+18 cases on top of PR 1's 40 (total 58 in
``tests/darwin/future/data_objects/test_nested_properties.py``):

- ``TestTopologicalSort`` (7 cases)
- ``TestEnrichWithMetadataValues`` (2 cases)
- ``TestTeamPropertyLookupsRegister`` (3 cases)
- ``TestCreateUpdateItemPropertiesNestingValidation`` (2 cases)
- ``TestAssignItemPropertiesToDataset`` (2 cases)
- New item-level case under ``TestResolveParentForCreate``
- Second integration test
  ``TestImportPropertiesNestedInOneBatch.test_creates_parent_before_child_and_resolves_names_to_uuids``
  exercising the topo sort + register flow with a child listed before
  its parent in metadata.

## Review

Recommended: review #1159 first. This PR is purely additive on top of
that foundation.

Made with [Cursor](https://cursor.com)

---

> [!NOTE]
> **Medium Risk**
> Touches property import/create/update flows and dataset assignment logic; bugs could break property creation order or produce invalid nesting/dataset updates, but changes are scoped and well-tested.
> 
> **Overview**
> The importer now supports **creating nested properties where parent and child are both created in the same run**, by enriching create payloads with all metadata-declared values, **topologically sorting** the create list by `parent_name` (with cycle/duplicate detection), and updating in-memory lookups as each property is created.
> 
> `TeamPropertyLookups` gains a `register()` method (and `refresh()` uses it) so newly created/updated properties are immediately available for subsequent name→UUID resolution, and `_resolve_parent_property_for_create` is extended to resolve **item-level** parents as well.
> 
> Item-level metadata creation now carries `parent_name`/`trigger_condition` with the same nesting validation, and dataset assignment skips nested item properties to avoid backend 422s. Tests are expanded substantially to cover sorting, enrichment, item-level resolution, lookup registration, and an end-to-end in-batch nesting scenario.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d88ad61ee7234ffbbdb1d9ab0acc7d89840132cd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/darwin/importer/importer.py
+++ b/darwin/importer/importer.py
@@ -40,6 +40,7 @@ from darwin.datatypes import (
 from darwin.future.data_objects.properties import (
     FullProperty,
     PropertyGranularity,
+    PropertyKey,
     PropertyType,
     PropertyValue,
     SelectedProperty,
--- a/darwin/importer/importer.py
+++ b/darwin/importer/importer.py
@@ -40,12 +40,10 @@ from darwin.datatypes import (
 from darwin.future.data_objects.properties import (
     FullProperty,
     PropertyGranularity,
-    PropertyKey,
     PropertyType,
     PropertyValue,
     SelectedProperty,
     TriggerCondition,
-    property_key,
 )
 from darwin.item import DatasetItem
 from darwin.path_utils import is_properties_enabled, parse_metadata
@@ -560,16 +558,33 @@ def _topologically_sort_properties_to_create(
     if not properties_to_create:
         return []
 
-    index_by_key: Dict[PropertyKey, int] = {
-        property_key(p): i for i, p in enumerate(properties_to_create)
-    }
+    # Disambiguate item-level from class-level properties so two entries
+    # whose ``annotation_class_id`` is ``None`` (e.g. two item-level props,
+    # or an unvalidated class-level prop colliding with an item-level one
+    # of the same name) cannot silently overwrite each other in the index.
+    SortKey = Tuple[str, Optional[int], PropertyGranularity]
+
+    def _sort_key(prop: FullProperty) -> SortKey:
+        return (prop.name, prop.annotation_class_id, prop.granularity)
+
+    index_by_key: Dict[SortKey, int] = {}
+    for i, prop in enumerate(properties_to_create):
+        key = _sort_key(prop)
+        if key in index_by_key:
+            raise ValueError(
+                "Cannot topologically sort properties: duplicate entry "
+                f"for {key!r} in the properties-to-create batch."
+            )
+        index_by_key[key] = i
     children_by_parent_index: Dict[int, List[int]] = defaultdict(list)
     in_degree: List[int] = [0] * len(properties_to_create)
 
     for idx, prop in enumerate(properties_to_create):
         if prop.parent_name is None:
             continue
-        parent_idx = index_by_key.get((prop.parent_name, prop.annotation_class_id))
+        parent_idx = index_by_key.get(
+            (prop.parent_name, prop.annotation_class_id, prop.granularity)
+        )
         if parent_idx is None:
             # Parent not in this batch -> treat this node as a root.
             continue
--- a/darwin/importer/importer.py
+++ b/darwin/importer/importer.py
@@ -634,7 +634,9 @@ def _enrich_properties_with_metadata_values(
             if m_item is not None:
                 metadata_values = m_item.get("property_values") or []
         else:
-            class_name = class_name_by_id.get(prop.annotation_class_id or -1)
+            class_name = class_name_by_id.get(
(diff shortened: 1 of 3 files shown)


umbertoDifa:
Re Bugbot: thanks, addressed in 97aa251.

The collision was real: `property_key(p)` returning `(name, annotation_class_id)` could silently drop a node when two entries shared a key (e.g. two item-level props both with `annotation_class_id=None`). Fix:
- Use a local sort key `(name, annotation_class_id, granularity)` inside `_topologically_sort_properties_to_create` so item-level vs class-level entries can never collapse onto the same index.
- Raise `ValueError` on duplicate keys instead of silently overwriting, so the failure surfaces immediately.
- Added two regression tests: `test_class_and_item_level_with_null_class_id_do_not_collide` and `test_duplicate_entries_raise`.

The public `property_key` helper is left unchanged because other callsites (e.g. `darwin/datatypes.py`) still rely on its existing `(name, annotation_class_id)` shape.

darwin/importer/importer.py
```diff
@@ -545,6 +546,114 @@ def _property_value_from_metadata(m_value: Dict[str, Any]) -> PropertyValue:
...
+    is preserved. Raises ``ValueError`` on cycles.
+    """
+    if not properties_to_create:
+        return []
+
+    # Disambiguate item-level from class-level properties so two entries
+    # whose ``annotation_class_id`` is ``None`` (e.g. two item-level props,
+    # or an unvalidated class-level prop colliding with an item-level one
+    # of the same name) cannot silently overwrite each other in the index.
+    SortKey = Tuple[str, Optional[int], PropertyGranularity]
```
umbertoDifa, line 566: can use `property_key`?

umbertoDifa: This comment was made on the original `property_key`-based code; that path was changed in the Bugbot fix (commit 7be8cec2 after rebase) so the topo-sort key now disambiguates by `granularity` as well — `(name, annotation_class_id, granularity)`. `property_key()` is intentionally a 2-tuple to stay compatible with `TeamPropertyLookups.annotation_properties`, so the topo-sort uses a wider local `_sort_key` instead.

vvihorev (approved):

darwin/importer/importer.py
```diff
@@ -546,6 +547,114 @@ def _property_value_from_metadata(m_value: Dict[str, Any]) -> PropertyValue:
...
+    }
+
+    for prop in properties_to_create:
+        metadata_values: Optional[List[Dict[str, Any]]] = None
+        if prop.granularity == PropertyGranularity.item:
+            m_item = metadata_item_prop_lookup.get(prop.name)
+            if m_item is not None:
+                metadata_values = m_item.get("property_values") or []
+        else:
+            class_name = class_name_by_id.get(prop.annotation_class_id or -1)
```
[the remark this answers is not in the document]
umbertoDifa: Done in 9043b158: switched to an explicit None check.

PR #1177, opened 2026-07-21
feat(coco): export raster mask annotations as COCO RLE instances (DAR-7915)

### LINKED ISSUE #1173: COCO exporter silently drops raster (mask / raster_layer) annotations
### Summary
`darwin.exporter.formats.coco.export` does not handle raster mask annotations
(`annotation_type` = `mask` / `raster_layer`). For a dataset annotated with the
raster/brush mask tool, the produced COCO JSON has **empty `categories` and
empty `annotations`**, with no error raised — only stdout lines like
`skipping unsupported annotation_type 'mask'`. Downstream training/validation
then fails or silently trains on zero objects.

### Version
darwin-py 3.5.3 (latest on PyPI at time of writing).

### Reproduction
```python
from pathlib import Path
from darwin.exporter import exporter
from darwin.exporter.formats import coco

files = list(Path('annotations_with_raster_masks').glob('*.json'))
dt_gen = exporter.darwin_to_dt_gen(files, split_sequences=False)
coco.export(dt_gen, Path('out'))
# out/output.json -> {"categories": [], "annotations": [], "images": [...]}
```

Observed output:
```
skipping unsupported annotation_type 'mask'
skipping unsupported annotation_type 'raster_layer'
...
categories: []
n annotations: 0
n images: 20
```

Root cause:
- `_calculate_categories` only accepts `annotation_type in ["polygon", "bounding_box"]`.
- `_build_annotation` only handles `polygon`, `tag`, `bounding_box`.

Raster annotations parse to type `raster_layer` (see
`utils._parse_darwin_raster_annotation`) and are filtered out on both sides.

### Expected behavior
Raster mask annotations should be exported as COCO instance annotations:
- `categories` populated from the mask annotation class names;
- one annotation per instance, with `bbox` and RLE `segmentation`.

The `raster_layer` already carries a dense RLE plus a
`mask_annotation_ids_mapping` (pixel value -> mask annotation / instance), and
COCO natively supports RLE `segmentation` via `pycocotools`. So the conversion
is lossless and direct (no polygon approximation needed): decode the dense RLE,
and for each mapped instance emit its binary-mask RLE + bounding box.

### Impact
Any dataset annotated exclusively with the raster/brush mask tool cannot be
exported to a usable COCO file, which breaks common instance-segmentation and
detection training pipelines that rely on the COCO exporter.

### Possible workaround / suggestion
Add a `raster_layer` / `mask` branch to the COCO exporter that decodes the dense
RLE and emits per-instance COCO annotations (bbox + RLE segmentation). Happy to
help / open a PR if this is a welcome addition.

umbertoDifa:
## Summary
- The COCO exporter silently dropped `mask`/`raster_layer` annotations, producing empty `categories`/`annotations` for datasets annotated with the raster/brush tool (DAR-7915, closes #1173).
- Mask annotation classes are now included in COCO `categories`.
- Each `raster_layer`'s dense RLE is decoded once per file; every `mask` annotation is exported as a COCO instance annotation with uncompressed column-major RLE `segmentation`, `bbox`, pixel `area`, and `iscrowd: 0`. The conversion is lossless (no polygon approximation) and adds no new dependencies (`upolygon.rle_encode` already emits COCO-compliant counts).
- Malformed raster layers (missing fields, odd-length/invalid `dense_rle`, pixel-count mismatch, video annotations) are skipped with a printed message instead of aborting the export.

## Behaviour notes
- `mask` classes now appear in `categories` for all exports (previously absent).
- Annotation ids in files mixing raster layers with other types shift slightly: `raster_layer` entries no longer consume an id (ids only need to be unique in COCO; affected datasets previously exported empty output anyway).

## Test plan
- [x] `pytest tests/darwin/exporter/formats/export_coco_test.py` — 12 tests: hand-computed RLE counts/bboxes/areas on a 3×4 label map, raster-layer exclusion, orphan mask, malformed `dense_rle` (odd-length, missing field, negative count) skip paths
- [x] End-to-end cycle test committed: Darwin 2.0 JSON → `darwin_to_dt_gen` → `coco.export` → assertions on `output.json` (red-checked: fails on master's exporter, passes here)
- [x] Full suite: 756 passed, 29 skipped

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/darwin/exporter/formats/coco.py
+++ b/darwin/exporter/formats/coco.py
@@ -1,7 +1,7 @@
 from datetime import date
 from operator import itemgetter
 from pathlib import Path
-from typing import Any, Dict, Iterator, List, Optional
+from typing import Any, Dict, Iterator, List, Optional, Tuple
 from zlib import crc32
 
 import numpy as np
@@ -160,11 +160,25 @@ def _build_annotations(
 ) -> Iterator[Optional[Dict[str, Any]]]:
     annotation_id = 0
     for annotation_file in annotation_files:
+        raster_context = _build_raster_context(annotation_file)
         for annotation in annotation_file.annotations:
+            annotation_type = annotation.annotation_class.annotation_type
+            if annotation_type == "raster_layer":
+                # Raster layers are exported through their mask annotations.
+                continue
             annotation_id += 1
-            annotation_data = _build_annotation(
-                annotation_file, annotation_id, annotation, categories
-            )
+            if annotation_type == "mask":
+                annotation_data = _build_mask_annotation(
+                    annotation_file,
+                    annotation_id,
+                    annotation,
+                    categories,
+                    raster_context,
+                )
+            else:
+                annotation_data = _build_annotation(
+                    annotation_file, annotation_id, annotation, categories
+                )
             if annotation_data:
                 yield annotation_data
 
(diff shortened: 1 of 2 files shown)


fgandellini (approved):
I have zero context on this project.
The code I read kind of makes sense to me, but I can't really judge it.

